PR #18, opened 2018-03-11
Complain to stderr about unused raze settings.

mfarrugi:
Found myself upgrading crates and not (quickly) noticing that overrides were no longer being applied.

Not sure if this should be a harder error, lmk if you'd like it done differently. The output is kind of noisy right now as well, which may make this hard to notice if you don't know to look for it.

### DIFF
--- a/src/planning.rs
+++ b/src/planning.rs
@@ -108,14 +108,24 @@ impl<'a> BuildPlanner<'a> {
     let package_ids = try!(find_all_package_ids(source_id, &resolve));
 
     // Verify that user settings are being used
-    let package_strings = package_ids
-      .iter()
-      .map(|id| (id.name().to_string(), id.version().to_string()))
-      .collect::<HashSet<_>>();
+    let mut name_to_versions = HashMap::new();
+    for id in package_ids.iter() {
+      name_to_versions
+        .entry(id.name().to_owned())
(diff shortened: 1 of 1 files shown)


acmcarther:
This looks good to me. At some point I planned to print something like 
```Found X @ t1, which didn't have a matching dependency, did you mean X @ [t2,t3,t4]```
and I think there is code like that floating around somewhere in some other repo..

If you want to change this to hint at the available other versions, that would be nice, else we can do it some time later. LMK either way.

mfarrugi:
@acmcarther I added the version hint

acmcarther:
LGTM, thanks for quick fix!

mfarrugi:
@acmcarther I want to promote this to a hard error (since I ran into it again :|), any objection? 

The other option would be to replace the `Generated $CRATE successfully` with `Generated $count crates successfully`, but I imagine most people are going to ignore non-error output that takes more than 1 screen.

acmcarther:
Changing the warning to an error SGTM

Code was restructured recently, here is where the warning happens now:
https://github.com/google/cargo-raze/blob/afe42555876da022c42fe9cc8c6d16e747c2e4ce/impl/src/planning.rs#L788-L828

Issue #16, opened 2018-03-11
Support git dependencies


palfrey:
I have a git-based dependency in one of my projects that doesn't seem to get picked up by cargo-raze.

`hybrid-clocks = { git = "https://github.com/palfrey/hybrid-clocks.git", branch = "serde-0.9", features = ["serde"] }`

Admittedly, in this case, I think I can just upgrade to the latest version from crates.io now, but it would be good to support the general case.

mfarrugi:
I also ran into this with an unpublished crate, rust-gssapi, we ended up writing the `BUILD` by hand.

acmcarther:
Seems that [cargo-vendor supports this](https://github.com/alexcrichton/cargo-vendor/blob/f5699e45583673976edbe53072837513aa5eb639/src/main.rs#L299), so I don't see a reason why it couldn't be supported on our end.

Remote genmode would need to learn how to vendor things that aren't crates.io crates on S3 though.

I'll look into this. If you don't see updates here, feel free to ping again for an update.

bspeice:
I'm interested in getting the remote working - moving more of the hard work towards bazel. That said, recent PR should add support for Git, I tested it working using the dependency you specified.

acmcarther:
I believe that this is currently "incidentally" supported in vendor mode in https://github.com/google/cargo-raze/pull/28.

I did some ad-hoc manual testing there, but https://github.com/acmcarther/cargo-raze-examples/pull/5 will ensure it.

bspeice:
Confirmed that #28 works for vendored mode. I'll be working on code for remote mode next unless @acmcarther beats me to it.

acmcarther:
@bspeice have at it if you'd like!

Some initial direction:

1) The information you want will probably be available in  `metadata.packages[*].source`, though I'm not sure the best way to parse out the git repo details. Cargo's own source might be a good inspiration. The type of the field before serialization into a string was `SourceId`, and it can probably be turned back into one of those if that's useful.
2) You'll want to add an additional section to `CrateContext` with the information needed to know how to fetch the dependency. This will probably be a processed form of the data from (1)
3) That information will be used in `src/templates/remote_crates.bzl.template` to know whether to create a `new_git_repository` or a `new_http_archive` (as is the current default).

acmcarther:
This was fixed in https://github.com/google/cargo-raze/pull/44

PR #21, opened 2018-03-14
Issue #16 - Initial support for Git repositories

bspeice:
Intends to address google/cargo-raze#16, all tests are passing.


bspeice:
Committed with the wrong email. Going to reset that and try again.

Issue #7, opened 2018-02-28
Investigate using `cargo metadata` internally


acmcarther:
I wrote most of the integration before `cargo metadata` existed. I'm content with just continuting to use Cargo's internals in the event that `cargo metadata` is missing something needed, but if its possible to use it today, we probably should. 

Crosslinks:
https://github.com/rust-lang/cargo/issues/3815

acmcarther:
I've rewritten the planner to use output from `cargo metadata`, but I'm blocked by a small feature request to Cargo. https://github.com/rust-lang/cargo/pull/5122 adds the extra data I need.

Once this change is in, the whole stack will get much more testable as it won't depend on weird internal state of Cargo objects.

acmcarther:
cc https://github.com/rust-lang/cargo/issues/5583

I wasn't aware that these were outstanding issues. I'm currently working around lack of platform-specific details by performing dependency pruning:

https://github.com/google/cargo-raze/blob/2e4128f49ffaf7726002e30930db2aedbf205238/impl/src/planning.rs#L248-L257

(using dependency.target and comparing with the configured target).

https://github.com/google/cargo-raze/issues/53 tracks feature pruning (which is in an unclear state cargo-side as well).

UebelAndre:
Is this investigation still open?

PR #22, opened 2018-03-14
Issue #16 - Initial support for Git repositories

bspeice:
Intends to address google/cargo-raze#16, all tests are passing.

### DIFF
--- a/src/planning.rs
+++ b/src/planning.rs
@@ -100,12 +100,10 @@ impl<'a> BuildPlanner<'a> {
 
     let mut crate_contexts = Vec::new();
 
-    let source_id = match self.registry.clone() {
-      Some(v) => v,
-      None => try!(SourceId::crates_io(&self.cargo_config)),
-    };
-
-    let package_ids = try!(find_all_package_ids(source_id, &resolve));
+    let mut package_ids: Vec<PackageId> = Vec::new();
+    for dep in root_direct_deps.iter() {
+      package_ids.extend(find_all_package_ids(dep.source_id().clone(), &resolve)?);
+    }
 
     // Verify that user settings are being used
     let mut name_to_versions = HashMap::new();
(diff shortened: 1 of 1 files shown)


src/planning.rs
```diff
@@ -100,12 +100,10 @@ impl<'a> BuildPlanner<'a> {
...
 
-    let source_id = match self.registry.clone() {
-      Some(v) => v,
-      None => try!(SourceId::crates_io(&self.cargo_config)),
-    };
-
-    let package_ids = try!(find_all_package_ids(source_id, &resolve));
+    let mut package_ids: Vec<PackageId> = Vec::new();
+    for dep in root_direct_deps.iter() {
+      package_ids.extend(find_all_package_ids(dep.source_id().clone(), &resolve)?);
```
acmcarther: This might be better implemented by changing `find_all_package_ids` to take the root_package_id and filter out only that package_id. The current implementation takes the source_id only to avoid including the root (kinda cludgey, given that we have the root's package_id available, I know).

In any case, this implementation appears to loop over the whole dep set way more times than it probably needs to, given its objective of supporting git deps.

_Addendum_: This extra looping also causes a minor breakage: cargo.bzl and the alias BUILD file end up with one duplicate per dependency (so 5 deps caused there to be five copies of each dependency). Test suite needs help :cry: (that https://github.com/google/cargo-raze/issues/7 will begin to deliver)

src/planning.rs
```diff
@@ -100,12 +100,10 @@ impl<'a> BuildPlanner<'a> {
 
     let mut crate_contexts = Vec::new();
 
-    let source_id = match self.registry.clone() {
-      Some(v) => v,
-      None => try!(SourceId::crates_io(&self.cargo_config)),
-    };
-
-    let package_ids = try!(find_all_package_ids(source_id, &resolve));
+    let mut package_ids: Vec<PackageId> = Vec::new();
```
acmcarther: If `Remote` genmode is currently unsupported, perhaps we should fail harder. Something like

```rust
// N.B. Renamed for clarity
let upstream_registry_source_id = try!(SourceId::crates_io(&self.cargo_config));

// Verify that Remote genmode only has crates_io deps (others currently unsupported)
if self.settings.genmode == GenMode::Remote { 
  for dep in root_direct_deps.iter() {
    if dep != root_package_id && dep.source_id() != upstream_registry_source_id {
      panic!("Remote GenMode doesn't currently support crate sources other than crates_io (e.g. `git` is unsupported)");
    }
  }
}
```

src/planning.rs
bspeice: I think what you're ultimately after is some way of deduplicating `package_ids` after the iteration over root deps?
The loop was put in place because it originally assumed a `SourceId` of crates.io, and this now fixes is so that each root dependency uses its own Source (adding support for non-crates.io crates in addition to Git).
Happy to write up some test cases, wanted to get a commit started to solicit feedback like this!

src/planning.rs
bspeice: Agreed, and that's straight-forward to change. I want to test that I didn't accidentally introduce support for it in remote GenMode (and will help me figure out what is needed there), but following confirmation that it's broken, will add a failback.

bspeice:
As discovered in acmcarther/cargo-raze-examples#5, this doesn't actually "fix" Git. The dependencies are resolved, but the BUILD files aren't created correctly.
Given that this doesn't truly fix anything at the moment, and the refactoring work in #28, I think this is better off closed for now. I need to take some time and make sure that everything is fixed on the refactored version, and also take remote genmode into account.

Issue #23, opened 2018-03-14
Automate smoke testing


acmcarther:
Due to the deficient test suite, when I'm preparing to submit code to google/cargo-raze I run a smoke test against cargo-raze-examples. The steps look roughly like this:

```
# From working directory
$ cargo install -f
$ git clone git@github.com:acmcarther/cargo-raze-examples
$ cd cargo-raze-examples/internal && ./evaluate-local-raze.sh
```

This runs `cargo-raze` across a swath of configurations and attempts to build them. If we run Travis in `sudo` mode, it should be possible to do this in CI -- though this will then require:
1) The rust tests to be run "manually" (instead of automagicially, as they are now)
2) `bazel` to be installed before build

PR #27, opened 2018-03-14
Add cargo-raze-examples smoketest

acmcarther:


acmcarther:
Verified that CI breaks for changes that produce broken bazel builds:

Branch:
https://github.com/google/cargo-raze/compare/acm-better-tests...acm-broken

Failed Build:
https://travis-ci.org/google/cargo-raze/builds/353514507

---

This is seriously imperfect, in part because 
- It is slow
- It sacrificed nightly and beta builds to gain back some speed
- It might get flaky if acmcarther/cargo-raze-examples gets some broken stuff checked in

... but it does fix https://github.com/google/cargo-raze/issues/23

PR #28, opened 2018-03-15
Refactor planning in the manner of `cargo metadata` (but keep using internals)

acmcarther:
This is going to be a very disruptive PR to https://github.com/google/cargo-raze/pull/22, so I'm putting it up while i'm still wrapping it up.

@bspeice I can roll your changes into this PR if you'd like. Actually, I suspect I already did -- I'm not filtering on source_id at all anymore.


src/metadata.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+  use super::*;
+  use serde_json;
+
+  #[test]
+  fn test_metadata_deserializes_correctly() {
+    let metadata_file_contents = include_str!("../test_fixtures/metadata.txt");
```
bspeice: Can the test fixtures be moved inside the project? It's not intuitive what content this actually represents.

bspeice:
Eh, for as much as was actually changed in #22, I can't imagine it would take me too long to re-do on top of this. Go ahead and focus on the refactor, I'm still working to bring myself up to speed.
I'm getting an exception when running though, I'll put the stack trace here and try and get the git remote bazel build packaged up to submit for the examples repo. I imagine that helps both with the smoke testing you rolled out, and in general proving "hey, this is useful in practical circumstances."

acmcarther:
Thanks for the trace @bspeice, I'll get that working in this PR.

Note to self: Review recent commits (don't let tests or new features get lost!)

acmcarther:
Submitting this for now to unblock git deps.

Plenty more tests to be added in a subsequent PR.

PR #29, opened 2018-03-15
Check package.source under Remote genmode for non crates-io srcs

acmcarther:
Also prints full metadata on generation failure. This might be a bit overwhelming to the user, but it appears 
above the actual error, so it should be OK. It will provide important diagnostic context for user reported errors here.

My machines can't seem to agree on the rust format. Frustrating. Will fix before submit


UebelAndre:
@acmcarther Is this PR still relevant in terms of the functionality it's trying to provide? Maybe this could be converted into a feature request issue?

Issue #32, opened 2018-03-19
Set RUSTC environmental for build scripts


SirVer:
Some crates, like `cargo_version` (https://github.com/Kimundi/rustc-version-rs/blob/master/src/lib.rs#L126) use this environment variable to figure out which rustc is running. This crate is for example a dependency for rusoto.

Changing the auto generated rusoto_core `genrule` to settings this variable using `export RUSTC=\"$$PWD/$(location @rust_linux_x86_64//:rustc)\";"` and adding it in the tools section worked for me, but maybe this variable should always be set?

acmcarther:
Hmm.... I think there is an additional subtlety here:

If a user has some version of rustc that's not following the exact import pattern that is suggested by `rules_rust`, they might see some really confusing stuff when they set `gen_buildrs = true` for the first time.

Perhaps this `rustc` path should point at an alias rule defined in `//.../cargo` as `rustc`. I don't think there is a good way to avoid clobbering this on regen, but it would at least be a place where this could be overridden/configured. Or, maybe this could be a raze setting. That would avoid having multiple sources of truth.

Thoughts?

SirVer:
I think it is perfectly fine to hardcode the naming schema that rules_rust are using. raze could be advertised and bound to rules_rust as a package - which is very attractive indeed. And if somebody feels like they need to support a different naming scheme, they can always use the `bind` escape hatch - if I understand the bazel docs correctly.

acmcarther:
Ah, actually, we're already hardcoding the exact path in existing rules, as part of the load statement:

```
    "@io_bazel_rules_rust//rust:rust.bzl",
```

Whenever we add a `rules_rust_path` rule, we can solve everything at once. For now, I don't think we need to worry about such flexibility.

acmcarther:
I'm reverting that CL for two reasons

1) It hardcodes the _linux_ rustc.
2) It didn't include `rustc` in the _tools_ section.

I only noticed when trying to use this for a local build. Need to figure out why CI didn't catch it.

acmcarther:
I actually only just got around to reverting this, but I have a replacement for it scoped for https://github.com/google/cargo-raze/pull/54

Since that PR introduces a notion of platform (derived from Cargo target_triple) we should be able to perform a `select` against the *host platform* to locate the appropriate tool for the build script.

dayfine:
Just came here to note that this still does not work as I am attempting to build external protobuf package with bazel and raze.

[...]

PR #33, opened 2018-03-19
Add RUSTC as env_var to buildscripts

acmcarther:
Fixes https://github.com/google/cargo-raze/issues/32

Probably will be broken in CI. Will fix.

### DIFF
--- a/src/templates/partials/build_script.template
+++ b/src/templates/partials/build_script.template
@@ -32,6 +32,7 @@ genrule(
     local = 1,
     cmd = "mkdir {{ crate_name_sanitized}}_out_dir_outputs/;"
         + " (export CARGO_MANIFEST_DIR=\"$$PWD/{{ workspace.workspace_path | replace(from="//", to="") }}/vendor/{{ crate.pkg_name }}-{{ crate.pkg_version }}\";"
+        + " export RUSTC=\"$$PWD/$(location @rust_linux_x86_64//:rustc)\";"
         + " export TARGET='{{ crate.platform_triple }}';"
         + " export RUST_BACKTRACE=1;"
         {%- for feature in crate.features %}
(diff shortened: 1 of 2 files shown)


acmcarther:
Thanks @SirVer for the exact fix suggestion, looks correct (and is probably correct +- oddities in examples)!

acmcarther:
Oh, it just works.

I guess this expresses the "chicken-or-egg" problem of new features in the current project structure. We know that this didn't break the existing examples, but what we really want to know is if this works with a crate where it's actually used.

... but I can't add that example without this feature.

If I can think of a reliable way to defer the actual execution of the example generation until after the code is submitted (so it doesn' t clutter up diffs) the example repo can be merged back in. That would let us atomically add examples *and* features.

acmcarther:
This is broken in a couple of ways (not sure how CI missed it):

Exhibit A:
 bazel build //third_party/cargo:all                                                                     [acm-rm-shame]
ERROR: /usr/local/google/home/acmcarther/.cache/bazel/_bazel_acmcarther/790a941678244b6c942ed3c699de3d11/external/raze__bindgen__0_35_0/BUILD.bazel:45:11: in cmd attribute of genrule rule @raze__bindgen__0_35_0//:bindgen_build_script_executor: label '@rust_linux_x86_64//:rustc' in $(location) expression is not a declared prerequisite of this rule
ERROR: Analysis of target '//third_party/cargo:cargo_bin_bindgen' failed; build aborted: Analysis of target '@raze__bindgen__0_35_0//:bindgen_build_script_executor' failed; build aborted
INFO: Elapsed time: 16.346s
```
(rustc needs to be added to `tools`)

In addition, we can't just use `@rust_linux_x86_64//:rustc`, as that compiler is target specific.

Going to roll back for now.

PR #45, opened 2018-04-14
Remove remote templates and use additional data in context

acmcarther:
PR currently has some garbage diff from a bad locally installed formatter. Going to fix on home machine.


acmcarther:
Huzzah!

PR #47, opened 2018-06-03
Add SHA-256 to the local_repositories

damienmg:
The SHA-256 is available from Cargo so it's easy to expose and it helps Bazel with caching and reproducibility.

This PR also contains 2 minor fixes:
  - Update Cargo.lock, the current one fail on my machine (Ubuntu 18.04)
  - Remove some unused variables and mark two modules as test config to remove unused warnings.

### DIFF
--- a/impl/src/metadata.rs
+++ b/impl/src/metadata.rs
@@ -73,6 +73,7 @@ pub struct Package {
   pub targets: Vec<Target>,
   pub features: HashMap<String, Vec<FeatureOrDependency>>,
   pub manifest_path: String,
+  pub sha256: Option<String>,
 }
 
 #[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
@@ -272,6 +273,9 @@ impl<'config> MetadataFetcher for CargoInternalsMetadataFetcher<'config> {
         features.insert(feature.clone(), features_or_dependencies.clone());
       }
 
+      // Cargo use SHA256 for checksum so we can use them directly
+      let sha256 = package.manifest().summary().checksum().and_then(|x| Some(x.to_string()));
+
       packages.push(Package {
         name: package.name().to_string(),
         version: package.version().to_string(),
@@ -284,6 +288,7 @@ impl<'config> MetadataFetcher for CargoInternalsMetadataFetcher<'config> {
         targets: targets,
         features: features,
         manifest_path: package.manifest_path().display().to_string(),
+        sha256: sha256,
       });
     }
 
@@ -353,6 +358,7 @@ pub mod testing {
       targets: Vec::new(),
       features: HashMap::new(),
       manifest_path: String::new(),
+      sha256: None,
     }
   }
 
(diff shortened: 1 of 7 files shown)


damienmg:
The update to cargo.lock is actually addressing the problem we see in Travis at HEAD.

impl/src/templates/remote_crates.bzl.template
```diff
@@ -10,7 +10,10 @@ def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
         name = "{{workspace.gen_workspace_prefix}}__{{crate.pkg_name | slugify | replace(from="-", to="_")}}__{{crate.pkg_version | slugify | replace(from="-", to="_")}}",
         url = "https://crates-io.s3-us-west-1.amazonaws.com/crates/{{crate.pkg_name}}/{{crate.pkg_name}}-{{crate.pkg_version}}.crate",
         type = "tar.gz",
-        strip_prefix = "{{crate.pkg_name}}-{{crate.pkg_version}}",
+        {% if crate.sha256
```
acmcarther: Nit: use "{%-" to accomplish the same thing as you're accomplishing by splitting the tag.

Example: https://github.com/google/cargo-raze/blob/master/impl/src/templates/crate.BUILD.template#L23

impl/src/metadata.rs
```diff
@@ -272,6 +273,9 @@ impl<'config> MetadataFetcher for CargoInternalsMetadataFetcher<'config> {
         features.insert(feature.clone(), features_or_dependencies.clone());
       }
 
+      // Cargo use SHA256 for checksum so we can use them directly
+      let sha256 = package.manifest().summary().checksum().and_then(|x| Some(x.to_string()));
```
acmcarther: Two Nits:

s/`and_then(|x| Some(x.to_string()))`/`map(ToString::to_string)`

The `and_then` is only necessary when the mapping function yields an option. `map` is similar but without the option. `ToString::to_string` is the direct reference to the method (though `|x| x.to_string()` is totally fine as well).

impl/src/templates/remote_crates.bzl.template
damienmg: Much nicer. Done!

impl/src/metadata.rs
damienmg: Done.

damienmg:
PTAL. Thanks for the review!

acmcarther:
This PR needs a followup. Not usre how I missed this in review, but you added a field into a struct that I technically don't control the definition of -- Metadata.{...} is controlled by cargo as it's what they emit from `cargo metadata`.

SerializedPackage: 
https://github.com/rust-lang/cargo/blob/9c78c3a17ac4bc0c8b3b837095f60aa84d09c426/src/cargo/core/package.rs#L32-L49

When I throw the switch to use `cargo metadata` instead of the internals, the new sha256 field will always be empty. I think Cargo team might be open to accepting a PR that adds this to their output though. I'll followup in https://github.com/google/cargo-raze/issues/57

PR #49, opened 2018-06-04
Add version number to imported crates

damienmg:
Now that rust_library/rust_binary supports it.

### DIFF
--- a/impl/src/templates/partials/build_script.template
+++ b/impl/src/templates/partials/build_script.template
@@ -21,6 +21,7 @@ rust_binary(
       {%- endfor %}
     ],
     data = glob(["*"]),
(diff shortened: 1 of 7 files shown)


acmcarther:
I suspect this either needs to be smarter and detect if rust_library supports the version attr, or be driven by a flag in the `[raze]` section.

May not be a problem in practice, thoughts?

damienmg:
Rebased.

I don't think this will be a problem, people that use latest version of cargo-raze are going to use the latest version of rules_rust (whereas an incorrect version is resulting in weird stuff, like "generated by rust protobuf 0.0.0"). I don't see a case where we don't want to propagate this information.

acmcarther:
That SGTM.

PR #50, opened 2018-06-04
Remote: do not create http repository if it already exists.

damienmg:
Using native.existing_rules allow to use several crates.bzl file that
would import the same crates without failing with duplicate rules.

### DIFF
--- a/impl/src/templates/remote_crates.bzl.template
+++ b/impl/src/templates/remote_crates.bzl.template
@@ -4,9 +4,13 @@ cargo-raze crate workspace functions
 DO NOT EDIT! Replaced on runs of cargo-raze
 """
 
+def _new_http_archive(name, **kwargs):
+    if not native.existing_rule(name):
+        native.new_http_archive(name=name, **kwargs)
+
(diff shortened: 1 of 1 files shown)


acmcarther:
Sync w/ master should fix your build (your other CL was merged).

Can you share a little bit more detail about the use case that motivated this change? I see that, mechanically speaking, it lets you combine several `crates.bzl` files. This sounds fine to me, but I'm also interested to hear what kind of situation led to needing several `crates.bzl` files....

damienmg:
Rebased.

It's not something I encountered with raze yet but it something that happens all the time in other similar tools. The classic use case is I include libA and libB, both use cargo raze and both wrap the repositories in a bzl file exposed to the people. So you call liba_repositories() and llibb_repositories() and you get a collisions.

acmcarther:
I'm really nervous thinking about users having multiple cargo-raze resolved crate sets in the same workspace, but in reality this is just fixing a bug that someone will inevitably run into.

I'm thinking that ultimately "library-like" workspaces will need a way to express their third party crate dependencies, and a root workspace will need a mechanism to supply them. The prior art here isn't very encouraging.

Issue #43, opened 2018-04-03
Support for targeting multiple architectures


negator:
(Moving from: https://github.com/acmcarther/cargo-raze/issues/33)
It would be great if there was a straightforward way to add support for multiple architectures at once. A bit like how the Bazel --ios_multi_cpus works, but perhaps just allow an array of targets in Cargo.toml, which generates arch specific rules.

Answer from @acmcarther:

"I'm certainly open to this, but each target generates features, build scripts, and dependencies that are different, and I haven't had a chance to give enough though to the best way to organize the generated files. Do you have any ideas in that respect?"

negator:
My first thought was to organize the generate rust rules into arch specific subdirectories under `deps`, like rust does with `target`.  And generate a top level `BUILD` file that alias the arch specific ones switching on target cpu specified by `config_setting`. 

Something to the effect of:
-deps
  |-BUILD (aliases)
  |-arch_64_apple_ios
     |-BUILD
     |-remote
        |-...BUILD (referencing aliases)

As for the target triple mismatch, maybe its sufficient to statically map the most commonly used overlap between rustc and bazel. 

The default bazel toolchain (as of 0.8.1) looks like:
   darwin,
   darwin_x86_64,
   k8,
   ios_x86_64,
   ios_i386,
   ios_armv7,
   ios_arm64,
   watchos_i386,
   watchos_armv7k,
   tvos_x86_64,
   tvos_arm64,
   armeabi-v7a

akhilles:
Would it make more sense to do the select switching at the alias level or the generated BUILD file level? Also, rust_rules already has constraints based on target triples (https://github.com/bazelbuild/rules_rust/blob/master/rust/platform/platform.bzl). These can probably be used in the dependency selection.

UebelAndre:
Hi, I'm interested in seeing this get done. Would it be possible to teach `cargo-raze` about the platform specific dependency tags in [Cargo.toml](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies)?

Particularly:
```
[dependencies]
ferris-says = "0.2.0"

[target.x86_64-pc-windows-gnu.dependencies]
winhttp = "0.4.0"

[target.i686-unknown-linux-gnu.dependencies]
openssl = "1.0.1"
```

Here, 3 platforms would be targeted with the following dependencies
```
[host]
ferris-says = "0.2.0"

[.x86_64-pc-windows-gnu]
[...]

PR #54, opened 2018-06-22
WIP: Support multiple targets in output generation

acmcarther:
This will resolve https://github.com/google/cargo-raze/issues/43

This is an active PR, but i want to be able to see the diff. I'll tag stakeholders when PR is ready.

The strategy pivots around generating `config_setting` entries in the cargo dir BUILD file that can be referenced by all generated build rules. Those config_settings basically translate target triples into `@bazel_tools//platform` entries, so they're a near enough match.

Doing this requires differentiating between deps common to all envs, and deps specific to one or more targets, so planning needs some updates as well. 

Hypothetical output BUILD file:
```python
 package(default_visibility = ["//visibility:public"])

INTERNAL_WORKSPACE = "@raze_internal"

licenses([
  "notice", # "BSD-3-Clause"
])

load(
    "@io_bazel_rules_rust//rust:rust.bzl",
    "rust_library",
    "rust_binary",
    "rust_test",
    "rust_bench_test",
)

load(
    INTERNAL_WORKSPACE + "//:util.bzl",
    "mandatory_flags",
    "crate_target",
    "crate_config_setting",
)

rust_library(
    name = "adler32",
    crate_root = "src/lib.rs",
    crate_type = "lib",
    srcs = glob(["**/*.rs"]),
    deps = [
      crate_target("semver", "0_8_0")
    ] + select({
        INTERNAL_WORKSPACE + ":x86_64-unknown-linux-gnu": [
          crate_target("error_chain", "0_11_0")
        ]
    }),
    rustc_flags = [
    ] + mandatory_flags() + select({
        INTERNAL_WORKSPACE + ":x86_64-unknown-linux-gnu": [
            "--target=x86_64-unknown-linux-gnu",
        ]
    }),
    crate_features = [
    ] + select({
        INTERNAL_WORKSPACE + ":x86_64-unknown-linux-gnu": [
            "some_feature"
        ]
    }),
)
```


damienmg:
A few follow-up after our VC:

1. Why not put all the constraints in the rules_rust repository? Also probably better if they are linked to the actual toolchain since they are definitely matching the platform concept in Bazel.
2. The target flag is definitely something that should be set by the toolchain, you don't want to set it by rule whereas it should always be there.

acmcarther:
> 1. Why not put all the constraints in the rules_rust repository? Also probably better if they are linked to the actual toolchain since they are definitely matching the platform concept in Bazel.

The constraint settings _can_ live in rules rust -- there isn't anything workspace-specific about it. I do want to have a good story for extending the default supported triples... but we can hammer this out in a rules_rust PR.

> 2. The target flag is definitely something that should be set by the toolchain, you don't want to set it by rule whereas it should always be there.

When you say "target flag", are you referring to the target triple? If so then yes, I did envision somehow using the target triple from the toolchain to select the appropriate config settings.... or perhaps something in reverse. Maybe the toolchain should be internally configured based on the config settings values to match this. I'm not sure if the toolchain can be configured using config_settings though... I would need to do some research to find out.

damienmg:
> > 2. The target flag is definitely something that should be set by the toolchain, you don't want to set it by rule whereas it should always be there.
>
> When you say "target flag", are you referring to the target triple? If so then yes, I did envision somehow using the target triple from the toolchain to select the appropriate config settings.... or perhaps something in reverse. Maybe the toolchain should be internally configured based on the config settings values to match this. I'm not sure if the toolchain can be configured using config_settings though... I would need to do some research to find out

Yes I am referring to the target triple. The toolchain rule transport provider and can provide various attribute to the rule that can inject directly those flag or target specific dependencies.

acmcarther:
@damienmg:

I need to review the changes again as I haven't looked in a while. IIRC this predates my reworking of toolchain management in rules_rust (which was a blocker of this PR). The toolchain-side issues were not completely resolved to my satisfaction, as there needs to be a way to properly select a crosslinker (to run on the exec_triple and target the target_triple).

I can commit to updating and polishing this up and providing a more specific set of blockers over the weekend, if that works for you.

damienmg:
That would be awesome, thanks. Also tell me if I can help making this
change happen :)

acmcarther:
@damienmg: I didn't get a chance to devote the time to this that it deserved this weekend. Regardless, I do have a good suggestion on how you can help push it forward:

I've mentioned several times that selecting the right linker per `toolchain {exec_triple, target_triple}` is a blocker. It remains a blocker. Its quite difficult to supply dedicated linker binaries for all exec/target combinations. I haven't explored the solution space very much, but I did have a possible avenue of exploration if you've got time to look at it:

Rust is currently moving to use [LLD](https://lld.llvm.org/) (the llvm linker) instead of gcc. This linker has the benefit of being guaranteed to be able to cross link to whatever target is specified. I think it might be reasonable to partially enable cross compilation via this linker [(using the appropriate rustc invocation flag)](https://github.com/rust-lang/rust/issues/52501). Unfortunately, it is nightly only [(as well as tier 1 plat only -- so the impl needs to fail gracefully)](https://github.com/rust-lang/rust/issues/52362#issuecomment-409940038), but we could probably constraint our cross compilation support to nightly in the short term.

I checked out rustc-nightly-x86_64-unknown-linux-gnu and the lld binary is indeed present @ rustc/lib/rustlib/x86_64-unknown-linux-gnu/bin/rust-lld. Rustc can probably find it automatically with just the feature enabled, though, provided that the rust-lld bin is present in the filegroup specified by the toolchain BUILD contents.

damienmg:
@acmcarther at the moment I am not so much interested in cross compilation but rather to make my protobuf PR works on OSX too. I can manually amend the BUILD file to add the correct select but that does not scale (especially since I intend to add documentation about how to do it, since most of the case you want to regenerate it for your own project for now).

acmcarther:
Oh I see. I misunderstood your objectives.

I lost the thread of the original design of this PR, but I believe it requires predefined platforms to be added to rules_rust. I've added them here: https://github.com/bazelbuild/rules_rust/pull/144, and we can start using them in cargo-raze once that is submitted.

sayrer:
Is there any update on this?

It looks like this a volunteer effort, so I don't really have many expectations, but the docs on the README are wrong now (the code no longer produces a rustc target argument). Or, would it make more sense to wait for a "cargo_crate" rule in rules_rust?

sayrer:
or, maybe this mostly works now? I just saw https://github.com/bazelbuild/rules_rust/issues/205

UebelAndre:
This PR can likely be closed since #212 got merged.

UebelAndre:
@acmcarther this PR could probably be closed

PR #56, opened 2018-06-25
Clean up the planner code

acmcarther:
Breaks out the BuildPlannerImpl classes' responsibilities into a bunch of smaller planners which have reasonable scope. Removes weird hash maps. Adds CargoCrateCatalog and uses it to look up things.

This gives me a reasonable place from which to continue https://github.com/google/cargo-raze/pull/54 without having to keep a billion hashmaps and random blobs in my head.

EDIT:
This also renames and gently restructures some code I nitted but allowed through during review. Gotta have those incredibly long variable names or I let my Java heritage down. I hope nobody is offended.

### DIFF
--- a/impl/src/bazel.rs
+++ b/impl/src/bazel.rs
@@ -46,8 +46,14 @@ impl BazelRenderer {
           "templates/partials/rust_library.template",
           include_str!("templates/partials/rust_library.template"),
         ),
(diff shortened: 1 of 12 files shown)


acmcarther:
The refactoring saga is far from over. The current relationship between the Renderer and the Planner is still not well spec'd: the renderer just takes random fields out of the context and uses them in not well specified ways.

Changes here are fairly well contained though and unit + smoke tests pass.

PR #60, opened 2018-08-25
Remove rust_bench_test from generated imports

damienmg:
This rule is never used and was renamed in recent rust rules into
rust_bechmark.

### DIFF
--- a/impl/src/templates/crate.BUILD.template
+++ b/impl/src/templates/crate.BUILD.template
@@ -22,7 +22,6 @@ load(
     "rust_library",
     "rust_binary",
     "rust_test",
-    "rust_bench_test",
 )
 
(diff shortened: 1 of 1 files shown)


acmcarther:
LGTM, thanks for the quick fix.

In case any code explorers land here and wonder why rust_bench_test isn't generated: The reason is that dev dependencies aren't included in lockfile (and therefore, dependency tree) generation, and aren't necessarily available for use in the test targets.

pcj:
Please release a new version on crates.io that includes this change, thanks!

acmcarther:
@pcj: Pushed 0.1.1 @ dc18b4bfa2c9757f8998d6271a2440b08f51722c to crates.io

Issue #66, opened 2018-10-20
Fails to build on systems with OpenSSL 1.1.1


acmcarther:
Moved from https://github.com/acmcarther/cargo-raze/issues/34

Per @alaroldai

---

Related: sfackler/rust-openssl#987

Because cargo-raze depends on an old version of cargo (v0.20.0, current version is v0.30.0), which in turn depends on an old version of openssl, cargo-raze fails to build on systems with OpenSSL v1.1.1 installed because the openssl crate fails to detect the installed OpenSSL version.

I've tried fixing this locally by updating the dependency on Cargo, but it looks like there are some breaking API changes that make that a more complicated process than I expected. Someone more familiar with the Cargo API might have a better shot at this.

In case it's related, here's some system info:
OS: Arch Linux (Linux v4.18.12)
OpenSSL version: 1.1.1 (11 Sep 2018)
Version of the Cargo binary I used to build cargo-raze: 1.29.0

Here's a sample of the build output after updating the version of cargo:

---

I've observed that the main issues are
1) Changes in CLI handling
2) Replacement of rustc-serialize with serde
3) Changes in error handling to the `failure` crate

There are other miscellaneous small changes in param types or fields as well.

I'm looking at this now, but might not have a PR out immediately as I learn about these new APIs.

PR #67, opened 2018-10-20
Update dependencies and fix for cargo 0.30.0

acmcarther:
The largest change of consequence from this upgrade is that the error handling had to change a bunch because strings don't directly coerce to CargoError. I added a RazeError type for this purpose, but I didn't put much thought into it. 

Fixes https://github.com/google/cargo-raze/issues/66

CI is probably going to fail for some reason or other.

EDIT: Reminder to self -- this is going to need a release.

### DIFF
--- a/impl/Cargo.lock
+++ b/impl/Cargo.lock
@@ -1,11 +1,7 @@
 [[package]]
-name = "advapi32-sys"
-version = "0.2.0"
+name = "adler32"
+version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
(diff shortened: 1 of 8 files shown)


acmcarther:
Something appears to be amiss with the github-travis webhooks that trigger the CI builds. It's probably related to the recent outages. The build for the latest commit is here (it is green): https://travis-ci.org/google/cargo-raze/builds/444548184

I fixed the failing smoketest by pinning the version of `futures` that the non_crateio_library examples uses. It was originally set to "master" -- not a good idea for reproducibility.

Issue #69, opened 2018-10-21
Bazel builds of packages error out on recent bazelbuild/rules_rust with "Option 'target' given more than once"


matt-duch:
When trying to build the example rust workspace with Bazel and using recent versions of the bazelbuild/rules_rust repo, 

```bash
bazel  build --verbose_failures //vendored/hello_cargo_library
```
Results in an error: `error: Option 'target' given more than once`

Works on the commit previous to this one:
https://github.com/bazelbuild/rules_rust/commit/b3c8badf2c10ebad8b71a106d3398e28804728d0

```
ERROR: /home/matt/dev/cargo-raze/examples/vendored/hello_cargo_library/cargo/vendor/cfg-if-0.1.2/BUILD:27:1: error executing shell command: '/bin/bash -c set -e; if [ ! -z "${TMPDIR+x}" ]; then mkdir -p $TMPDIR; fi; rm -rf bazel-out/k8-fastbuild/bin/vendored/hello_cargo_library/cargo/vendor/cfg-if-0.1.2/cfg_if.deps; mkdir bazel-out/k8-f...' failed (Exit 101): bash failed: error executing command 
  (cd /home/matt/.cache/bazel/_bazel_matt/66f27ff99aab57cee57e8b3b1584c611/execroot/io_bazel_rules_rust && \
  exec env - \
    PATH=/home/matt/.cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/matt/bin \
  /bin/bash -c 'set -e; if [ ! -z "${TMPDIR+x}" ]; then mkdir -p $TMPDIR; fi; rm -rf bazel-out/k8-fastbuild/bin/vendored/hello_cargo_library/cargo/vendor/cfg-if-0.1.2/cfg_if.deps; mkdir bazel-out/k8-fastbuild/bin/vendored/hello_cargo_library/cargo/vendor/cfg-if-0.1.2/cfg_if.deps
[...]

PR #70, opened 2018-10-22
Remove rust_library.rustc_flag.target setting per rules_rust changes

acmcarther:
Fixes https://github.com/google/cargo-raze/issues/69

The target setting is now being set (since https://github.com/bazelbuild/rules_rust/pull/136) by rules_rust internals and it is no longer appropriate for us to be setting it here. Per the current path we're taking for cross compilation, the toolchain itself is the only entity that knows the proper target setting. It has not yet been determined how the complation target will be set. Work and supplemental discussion is ongoing in https://github.com/google/cargo-raze/pull/54

The builds for this PR will be broken until https://github.com/google/cargo-raze/pull/67 is submitted (or a build containing the fix to the non_cratesio_library example.

### DIFF
--- a/impl/src/templates/partials/build_script.template
+++ b/impl/src/templates/partials/build_script.template
@@ -13,7 +13,6 @@ rust_binary(
     ],
     rustc_flags = [
         "--cap-lints allow",
-        "--target={{workspace.platform_triple}}",
     ],
     crate_features = [
       {%- for feature in crate.features %}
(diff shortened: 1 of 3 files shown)


acmcarther:
Github webhooks are still hosed, but the build for c26e735 completed successfully:

https://travis-ci.org/google/cargo-raze/builds/444663847?utm_source=github_status&utm_medium=notification

impl/src/templates/partials/build_script.template
```diff
@@ -35,6 +34,8 @@ genrule(
     local = 1,
     cmd = "mkdir -p {{ crate_name_sanitized}}_out_dir_outputs/;"
         + " (export CARGO_MANIFEST_DIR=\"$$PWD/$$(dirname $(location :Cargo.toml))\";"
+        # TODO(acmcarther): This needs to be revisited as part of the cross compilation story.
```
acmcarther: Perhaps not the best idea to insert a todo into generated code. Oh well...

PR #72, opened 2018-10-27
Remove --target flag

damienmg:
It is now automatically set by rules_rust. It should always be correct
even for the build script with Bazel handling the configuration.


acmcarther:
This overlaps somewhat with https://github.com/google/cargo-raze/pull/70 and I'm not sure why github is failing to identify it. Its a more "complete" version of that PR though, so SGTM.

PR #74, opened 2018-11-08
README: correct name for generated cargo_fetch_remote_crates()

th0br0:
cargo-raze generates `cargo_fetch_remote_crates()` and not `raze_fetch_remote_crates()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,8 @@ raze_fetch_remote_crates()
 This tells Bazel where to get the dependencies from, and how to build them:
 using the files generated into //cargo.
 
+*Note that this method's name depends on your `gen_workspace_prefix` setting*.
+
 You can depend on any *explicit* dependencies in any Rust rule by depending on
 `//cargo:your_dependency_name`.
 


acmcarther:
The function name is actually derived from your workspace prefix:
https://github.com/google/cargo-raze/blob/d3ea1174d63977d8854691e599619f4580271e8b/impl/src/templates/remote_crates.bzl.template#L14-L17

https://github.com/google/cargo-raze/blob/afe42555876da022c42fe9cc8c6d16e747c2e4ce/impl/src/settings.rs#L58-L60

... And the default _is_ raze:

https://github.com/google/cargo-raze/blob/afe42555876da022c42fe9cc8c6d16e747c2e4ce/impl/src/settings.rs#L169-L173

This might be a bit of a misfeature though. I did it this way so that multiple independent import functions could be loaded into the WORKSPACE file's scope without having a name conflict.

acmcarther:
Sorry, that last comment wasn't really prescriptive.

I'm totally fine with updating documentation, and even reconsidering how we generate that function -- within reason for existing users, of course. I just wanted to make sure the etymology was clear. Let me know what you think.

th0br0:
Thanks for the explanation.
I think that this actually makes sense, esp. as the external dependencies are wrapped in rule existance checks. I've adjusted the PR to just add a warning about how the method name comes to be.

acmcarther:
Change LGTM, thanks for following up!

PR #77, opened 2018-11-10
Remote mode: use Label() and non native git/http.

damienmg:
* native.new_git_repository is replaced by the skylark
new_git_repository
* native.new_http_archive is replaced by the skylark http_archive
* Surround labels with the Label function so that crate.bzl can be
  used from other repositories.


acmcarther:
LGTM

PR #80, opened 2019-01-09
Support for Rust editions

### LINKED ISSUE #79: edition support
https://github.com/bazelbuild/rules_rust/issues/157

without rules_rust changes, raze can emit `--edition 2018` as needed

kornholi:
Closes #79.

@acmcarther could you take a look?

### DIFF
--- a/impl/src/metadata.rs
+++ b/impl/src/metadata.rs
@@ -88,6 +88,7 @@ pub struct Package {
   pub targets: Vec<Target>,
   pub features: HashMap<String, Vec<String>>,
   pub manifest_path: String,
+  pub edition: String,
   pub sha256: Option<String>,
 }
 
@@ -124,6 +125,7 @@ pub struct Target {
   pub kind: Vec<String>,
   pub crate_types: Vec<String>,
   pub src_path: String,
+  pub edition: String,
 }
 
 /**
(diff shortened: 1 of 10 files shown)


impl/src/metadata.rs
```diff
@@ -305,7 +307,8 @@ impl<'config> MetadataFetcher for CargoInternalsMetadataFetcher<'config> {
           name: target.name().to_owned(),
           kind: util::kind_to_kinds(target.kind()),
           crate_types: crate_types,
-          src_path: target.src_path().display().to_string(),
+          src_path: target.src_path().path().display().to_string(),
```
kornholi: This will panic if the target is a [metabuild](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#metabuild),  but I don't think these are yet exposed in the metadata?

acmcarther:
Thanks for sending out this PR! I really appreciate it, and I'd bet mfarrugi@ does as well.
This looks good to me. I'll merge and cut a new release ASAP.

Quick note to self:
This CL adds an edition field at both the target level and the crate level. Most build units use their associated target-level edition settings, but build scripts have no associated target, and therefore inherit the crate's setting.

impl/src/metadata.rs
acmcarther: I actually wasn't aware that there was forward movement on the "metadeps" thing. I'll read into this, but I'm pretty comfortable with the change given that the tests are all passing.

acmcarther:
Published to crates.io as [0.2.0](https://crates.io/crates/cargo-raze/0.2.0) in https://github.com/google/cargo-raze/commit/8ec495b8c6201e40533bd498077ee1ddd3f023e3

kornholi:
So the bazel builds are failing on Travis, but I'm guessing this has to do with the bazel version (0.11). I'll submit a PR to bump that up to 0.17 (minimum for rules_rust)

acmcarther:
I wasn't aware that the CI was not working correctly.

I'll fix this today.

PR #81, opened 2019-01-09
Bump Bazel on Travis to 0.17.1

kornholi:
0.17 is the minimum version supported by rules_rust at this point.

Still unclear why Bazel build failures didn't cause a Travis failure (https://travis-ci.org/google/cargo-raze/builds/477524733). Maybe 0.11 doesn't return a bad exit code during loading phase failures?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ addons:
 before_install:
   - OS=linux
   - ARCH=x86_64
-  - V=0.11.1
+  - V=0.17.1
   - GH_BASE="https://github.com/bazelbuild/bazel/releases/download/$V"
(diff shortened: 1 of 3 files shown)


acmcarther:
[Stack overflow users claim](https://stackoverflow.com/questions/2870992/automatic-exit-from-bash-shell-script-on-error) that failures on the left hand side of a boolean operator don't properly trigger the `set -e` behavior.

Since master currently fails, I can test [the fix](https://github.com/google/cargo-raze/compare/acm-fix-ci) and merge this PR in once this PR is in.

acmcarther:
Thanks for making this followup PR to update rules_rust and add that missing dependency!

PR #82, opened 2019-01-09
Fix Broken smoke-test.sh

acmcarther:
Related: 
https://github.com/google/cargo-raze/pull/80
https://github.com/google/cargo-raze/pull/81

Waiting on #81 to finish and get submitted.

---

It seems smoke tests were broken in a variety of ways, some of which are really inexplicable:
1) **non_cratesio_library's futures dependency was plainly not compatible with the code.** It was trying to use a version of futures that didn't have functions it needed. CI for the PR that made the most recent changes passed, so perhaps we have some kind of caching issue?
2) **The workspace name for the example workspace was the same as rules_rust.** This apparently confuses Bazel greatly on newer versions of Bazel (but not older ones?)
3) Per the original PR name in smoke-test.sh **`bazel build X && bazel build Y` causes failures in the build of X to not trigger the `set -e` behavior.**

### DIFF
--- a/smoke-test.sh
+++ b/smoke-test.sh
@@ -74,7 +74,8 @@ for ex in $(find $EXAMPLES_DIR -mindepth 2 -maxdepth 2 -type d); do
     bazel_cargo_path="//$ex_type/$ex_name/cargo:all"
 
     echo "Running Bazel build for $bazel_path, $bazel_cargo_path"
-    bazel build "$bazel_path" && bazel build "$bazel_cargo_path"
+    bazel build "$bazel_path"
+    bazel build "$bazel_cargo_path"
 done
 
 cd "$PWD"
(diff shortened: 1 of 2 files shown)


acmcarther:
```
[23 / 24] Compiling Rust bin non_cratesio_library (1 files); 0s linux-sandbox
error[E0432]: unresolved import `futures::executor::block_on`
 --> vendored/non_cratesio_library/src/main.rs:7:5
  |
7 | use futures::executor::block_on;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `block_on` in `executor`
```

No idea what is going on here. It doesn't make sense for this to fail separately from current `master`, given the change in this CL, and it  doesn't really make sense for it to be failing at all since `futures` in this example is from git, at a specific tagged release: "0.1.19".

I'll figure out what's going on with this example and update this PR some time today.

Issue #89, opened 2019-04-05
proc-macro aliases aren't generated


mfarrugi:
https://github.com/google/cargo-raze/commit/1e389359c769a35b00497627e3eb2f3e1e18a547#r33043678

It looks like that line also prevents proc_macro aliases from being defined (specifically serde_derive and structopt_derive)

acmcarther:
Fixed in #90

mfarrugi:
Looks like https://github.com/google/cargo-raze/blob/03b0005e26b2983a3bedf759b7f2822a681097f1/impl/src/templates/partials/rust_binary.template#L12 is missing a sanitize too, blows up on `protobuf-codegen` and some others.

PR #90, opened 2019-04-05
Remove requirement for lib in crate for adding aliases

acmcarther:
I'm not sure why I originally added this constraint, but I'll remove it and see if CI accepts it.

cc @mfarrugi
https://github.com/google/cargo-raze/issues/89


acmcarther:
Published to crates.io as 0.2.2

PR #91, opened 2019-04-06
sanitize binary's lib dep

mfarrugi:
Fix other problem in #90 

I'm only 40% confident this works / is the correct thing to do. (ie. i did not build this locally)


acmcarther:
I wish I was more deliberate about the nomenclature ("rust target" vs "bazel target"). Looks to me like "lib_target_name" refers to a "rust target", so performing the sanitization here is appropriate given the status quo.

Issue #92, opened 2019-04-14
Question: Why not generate BUILD.bazel like other configuration generators?


mstg:
Hello!
This may be a "corner case" situation, but given other tools like Gazelle is generating BUILD.bazel, why isn't cargo-raze doing the same?

This would especially be helpful when using a case insensitive file system (Windows + WSL) considering some crates (openssl-sys) can have a directory called `build` and that makes cargo-raze fail with the following error message:
```
error: Is a directory (os error 21)
```

One can circumvent this problem by manually changing and recompiling, but maybe it'll be better to change it all together to BUILD.bazel. Any thoughts?

acmcarther:
Outputting `BUILD` instead of `BUILD.bazel` was not a deliberate choice on my part. The tool was developed primarily with the context of Google's internal monorepo, which still typically follows the `BUILD` convention. I'm quite open to converging on the more common filename.

I do have a concern that making this change quietly would cause subsequent runs of `cargo-raze` over existing filetrees to end up with two BUILD files, so a PR that makes this change should either make the output configurable, or be inteligent enough to remove any existing `BUILD` files. I don't have a strong opinion about those options.

My bandwidth is seriously limited recently, so I can't promise this change will happen quickly.

mstg:
Thanks for your response.
Maybe a flag to output `BUILD.bazel` instead of `BUILD` could be introduced? Then it can just be added to the README for case insensitive file systems.

I'll take a look and see what I can do!

mstg:
Closing this issue in favor of #93 PR

PR #93, opened 2019-04-15
[settings] Add output_buildfile_suffix option

mstg:
Related issue: #92 

Default is BUILD
Other values can be for example: BUILD.bazel
Any value is accepted.

This is to follow the default behaviour of other bazel related tools
that generate BUILD.bazel instead of BUILD firstly to separate
what is generated and what is written and to avoid name collisions on
case insensitive file systems (Windows + WSL) since some crates can
have a directory named build which leads to cargo-raze unable to create
a file called BUILD. A project having a directory named BUILD.bazel is
much less likely or less to none chance.

How to use (example):
Add following to the `Cargo.toml` file
```
[raze]
output_buildfile_suffix = "BUILD.bazel"
```

I am still very new to Rust, and may have written something not idiomatic at all, and would appreciate any advice to make it better.

EDIT:
This PR initially added a `GenMode` called VendoredBuildBazel but upon feedback changed to another option called `output_buildfile_suffix` to ensure greater control and compatibility with all `GenMode`s

### DIFF
--- a/impl/src/bazel.rs
+++ b/impl/src/bazel.rs
@@ -14,6 +14,10 @@
 
 use cargo::CargoError;
 use cargo::util::CargoResult;
+use tera;
+use tera::Context;
+use tera::Tera;
+
 use context::CrateContext;
 use context::WorkspaceContext;
 use planning::PlannedBuild;
@@ -21,9 +25,6 @@ use rendering::BuildRenderer;
 use rendering::FileOutputs;
 use rendering::RenderDetails;
 use util::RazeError;
-use tera::Context;
-use tera::Tera;
-use tera;
 
 pub struct BazelRenderer {
   internal_renderer: Tera,
@@ -140,7 +141,9 @@ impl BuildRenderer for BazelRenderer {
     planned_build: &PlannedBuild,
   ) -> CargoResult<Vec<FileOutputs>> {
     let &RenderDetails {
-      ref path_prefix, ..
+      ref path_prefix,
+      ref buildfile_suffix,
+      ..
     } = render_details;
     let &PlannedBuild {
       ref workspace_context,
@@ -164,7 +167,7 @@ impl BuildRenderer for BazelRenderer {
       })
     }
 
-    let build_file_path = format!("{}/BUILD", &path_prefix);
+    let build_file_path = format!("{}/{}", &path_prefix, buildfile_suffix);
     let rendered_alias_build_file = try!(
       self
         .render_aliases(&workspace_context, &crate_contexts)
@@ -186,7 +189,9 @@ impl BuildRenderer for BazelRenderer {
     planned_build: &PlannedBuild,
   ) -> CargoResult<Vec<FileOutputs>> {
     let &RenderDetails {
-      ref path_prefix, ..
+      ref path_prefix,
+      ref buildfile_suffix,
+      ..
     } = render_details;
     let &PlannedBuild {
       ref workspace_context,
@@ -197,7 +202,7 @@ impl BuildRenderer for BazelRenderer {
 
     // N.B. File needs to exist so that contained xyz-1.2.3.BUILD can be referenced
     file_outputs.push(FileOutputs {
-      path: "remote/BUILD".to_owned(),
+      path: format!("remote/{}", buildfile_suffix).to_owned(),
       contents: String::new(),
     });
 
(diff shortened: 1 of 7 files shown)


acmcarther:
Thanks for sending out a PR to fix this!

My general preference is to add this as a new `RazeSetting`, eg `RazeSetting.output_buildfile_suffix` rather than as a `RazeSetting.GenMode`, perhaps with a default value of "BUILD". I think this makes more sense, since the file name decision is relevant to all GenModes. Then, both branches of `CrateCatalogEntry.local_build_path` can consult the new setting when yielding a file path.

I'm definitely open to learning if there are additional benefits to having this as a full GenMode though.

mstg:
Hello!

First of all, thanks for your feedback! I greatly appreciate it.

I thought of that first, but since remote mode uses `package-name.BUILD` format I kind of assumed it would not be necessary for that mode specifically, but thinking back if any new `GenMode` (if needed for whatever reason) is added this option may be useful for that too. It also makes sense since this actually does not change the "storage" strategy.

I'll see if I can set something up quick.

mstg:
I have since added a new option called `output_buildfile_suffix` and removed the `VendorBuildBazel` GenMode.
The change was introduced in a force push.

I have also added some smoke tests to ensure the new build files work.

If anything is missing or should be done another way, I would greatly appreciate your feedback.

Thanks.

mstg:
There is some tests failing when `GenMode` is set to `Remote` especially with the `build_file` directive in `crates.bzl`, even though it should be working. I'll take a closer look at this tomorrow, but feel like this is the right path for now.

impl/src/planning.rs
```diff
@@ -133,6 +135,7 @@ struct CrateSubplanner<'planner> {
 pub struct PlannedBuild {
   pub workspace_context: WorkspaceContext,
   pub crate_contexts: Vec<CrateContext>,
+  pub buildfile_suffix: String,
```
acmcarther: Nit: I think RenderDetails is a better place for this field. It is currently initialized [here](https://github.com/google/cargo-raze/blob/30fb8b43192858316272b3a0e79c9b2ef86ed0bb/impl/src/bin/cargo-raze.rs#L134-L136), and the suffix can be extracted from the settings at that location.

impl/src/bazel.rs
```diff
@@ -250,39 +253,42 @@ impl BuildRenderer for BazelRenderer {
...
 
-  fn dummy_binary_crate() -> CrateContext {
+  fn dummy_binary_crate_with_name(buildfile_suffix: &str) -> CrateContext {
     CrateContext {
       pkg_name: "test-binary".to_owned(),
       pkg_version: "1.1.1".to_owned(),
       edition: "2015".to_owned(),
       features: vec!["feature1".to_owned(), "feature2".to_owned()].to_owned(),
-      expected_build_path: "vendor/test-binary-1.1.1/BUILD".to_owned(),
+      expected_build_path: format!("vendor/test-binary-1.1.1/{}", buildfile_suffix).to_owned(),
```
acmcarther: You can omit this `to_owned()` now that you've switched to `format!`.

mstg:
Hello!

Following additional changes has been done:
* Additional smoke tests removed
* `vendored/hello_cargo_library` now generates `BUILD.bazel` files
* `buildfile_suffix` is now moved to `RenderDetails`
* Redundant `to_owned()` is now removed

I have also rebased the commits to one commit only.

If you have any other suggestions, I would greatly appreciate it.
Thanks!

smoke-test.sh
```diff
@@ -38,6 +38,7 @@ git_repository(
     remote = "https://github.com/bazelbuild/rules_rust.git",
 )
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
acmcarther: Thanks for fixing this!

mstg:
Yes, it is the work of rustfmt :)

Awesome, thanks for bearing with me.

acmcarther:
Just tried to publish but I've got an ancient rustc on my workstation. Got a meeting now, but will push 0.2.3 within the hour.

mstg:
Great! I also forgot to add any documentation for this, so if you could do that or just ping me that would be awesome.

acmcarther:
Pushed 0.2.3 with this change.

I don't think there is a dedicated place for documentation of individual configuration fields beyond the link in the README to impl/settings.rs, so I guess what we have is sufficient.

PR #94, opened 2019-04-26
[BUILD] Fix inconsistent casing in targets and dep targets

mstg:
Not sure if it's WSL only, but mixed case crates has mixed case target
names, but they're only referred to with lowercase in deps.
This fixes the casing issue.

Example crate is `MacTypes_sys` which is a dependency of `security-framework-sys`.
Building that target would fail with `unknown target mactypes_sys` because the target was called `MacTypes_sys`

Not sure if this has something to do with WSL (case insensitive file system), but this seems to fix it.
If this approach is flawed I would greatly appreciate any feedback.

This also removes the unnecessary comment of unsupported kind.

Thanks.

EDIT:
This PR has drastically changed the approach to fixing this issue.

### DIFF
--- a/impl/src/templates/crate.BUILD.template
+++ b/impl/src/templates/crate.BUILD.template
@@ -29,6 +29,7 @@ load(
 {%- include "templates/partials/build_script.template" %}
 {%- endif %}
 {%- for target in crate.targets %}
+{%- set target_name_sanitized = target.name | replace(from="-", to="_") %}
 {%- if target.kind == "lib" %}
 {%- include "templates/partials/rust_library.template" %}
 {%- elif target.kind == "bin" %}
--- a/impl/src/planning.rs
+++ b/impl/src/planning.rs
@@ -140,7 +140,7 @@ pub struct PlannedBuild {
 impl CrateCatalogEntry {
 
   pub fn new(package: &Package, is_root: bool, is_root_dep: bool) -> CrateCatalogEntry {
-    let sanitized_name = util::sanitize_ident(&package.name);
+    let sanitized_name = package.name.replace("-", "_");
     let sanitized_version = util::sanitize_ident(&package.version);
 
     CrateCatalogEntry {
--- a/impl/src/templates/partials/rust_binary.template
+++ b/impl/src/templates/partials/rust_binary.template
@@ -1,4 +1,4 @@
-{%- set target_name_sanitized = target.name | replace(from="-", to="_") %}
+
 rust_binary(
     # Prefix bin name to disambiguate from (probable) collision with lib name
     # N.B.: The exact form of this is subject to change.
--- a/impl/src/templates/partials/rust_library.template
+++ b/impl/src/templates/partials/rust_library.template
@@ -1,4 +1,3 @@
-{%- set target_name_sanitized = target.name | replace(from="-", to="_") %}
 {%- if target_name_sanitized != crate_name_sanitized %}
 alias(
   name = "{{ crate_name_sanitized }}",
--- a/impl/src/templates/remote_crates.bzl.template
+++ b/impl/src/templates/remote_crates.bzl.template
@@ -18,7 +18,7 @@ def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
 {% for crate in crates %}
     {%- if crate.source_details.git_data %}
     _new_git_repository(
-        name = "{{workspace.gen_workspace_prefix}}__{{crate.pkg_name | slugify | replace(from="-", to="_")}}__{{crate.pkg_version | slugify | replace(from="-", to="_")}}",
+        name = "{{workspace.gen_workspace_prefix}}__{{crate.pkg_name | replace(from="-", to="_")}}__{{crate.pkg_version | slugify | replace(from="-", to="_")}}",
         remote = "{{crate.source_details.git_data.remote}}",
         commit = "{{crate.source_details.git_data.commit}}",
         build_file = Label("{{workspace.workspace_path}}/remote:{{crate.pkg_name}}-{{crate.pkg_version}}.BUILD"),
@@ -26,7 +26,7 @@ def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
     )
     {%- else %}
     _new_http_archive(
-        name = "{{workspace.gen_workspace_prefix}}__{{crate.pkg_name | slugify | replace(from="-", to="_")}}__{{crate.pkg_version | slugify | replace(from="-", to="_")}}",
+        name = "{{workspace.gen_workspace_prefix}}__{{crate.pkg_name | replace(from="-", to="_")}}__{{crate.pkg_version | slugify | replace(from="-", to="_")}}",
         url = "https://crates-io.s3-us-west-1.amazonaws.com/crates/{{crate.pkg_name}}/{{crate.pkg_name}}-{{crate.pkg_version}}.crate",
         type = "tar.gz",
         {% if crate.sha256 -%}
--- a/impl/src/util.rs
+++ b/impl/src/util.rs
@@ -12,20 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::error::Error as StdError;
+use std::fmt;
+use std::iter::Iterator;
+use std::process::Command;
+use std::str;
+use std::str::FromStr;
+
 use cargo::CargoError;
 use cargo::core::TargetKind;
 use cargo::util::CargoResult;
 use cargo::util::Cfg;
-use slug;
-use std::fmt;
-use std::iter::Iterator;
-use std::process::Command;
-use std::str::FromStr;
-use std::str;
-use std::error::Error as StdError;
+use failure::Context;
 use failure::Error;
 use failure::Fail;
-use failure::Context;
+use slug;
 
 pub const PLEASE_FILE_A_BUG: &'static str =
   "Please file an issue at github.com/google/cargo-raze with details.";
--- a/smoke_test/remote/complicated_cargo_library/cargo/Cargo.lock
+++ b/smoke_test/remote/complicated_cargo_library/cargo/Cargo.lock
@@ -1,17 +1,17 @@
-[root]
-name = "compile_with_bazel"
-version = "0.1.0"
+[[package]]
+name = "MacTypes-sys"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "regex 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "specs 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.53 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "aho-corasick"
-version = "0.6.4"
+version = "0.6.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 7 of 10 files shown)


impl/src/templates/crate.BUILD.template
```diff
@@ -39,7 +40,5 @@ load(
 {%- include "templates/partials/rust_library.template" %}
 {%- elif target.kind == "rlib" %}
 {%- include "templates/partials/rust_library.template" %}
-{%- else %}
-# Unsupported target "{{ target.name }}" with type "{{ target.kind }}" omitted
```
acmcarther: What's the rationale for removing this line? I added it here because I thought it might be unclear to people why some targets are missing. This note gives them something to search for.

acmcarther:
Generally LGTM, one comment. I wasn't aware that some crate names are capitalized (and transitively, what effect this could have).

impl/src/templates/crate.BUILD.template
mstg: Ah sorry for not including that. Not sure if I did anything wrong but this line seemed to comment out the start of a rule (`rust_binary`) in the `cc` crate (iirc). The templates were with LF line endings and adding a new line after the unsupported target line did not work.

mstg:
Just replied to your review. Also CI seems to be failing of unrelated reasons. Any way to re-test?

acmcarther:
Ugh, seems rand crate owners moved the crate. I'm fixing this here in https://github.com/google/cargo-raze/pull/95. Can you hold off until I submit that, then sync and push?

mstg:
Of course. Just ping me when #95 is merged :)

impl/src/templates/crate.BUILD.template
acmcarther: Can you provide me an example of a generated BUILD file that does this?

the "{%-" has some semantic meaning around keeping or removing newlines, and might be related.  Unclear to me why only this target would cause you issues though since the other lines use the same thing. Your BUILD file should help.

mstg: The BUILD file causing the issue is here: https://gist.github.com/mstg/6991d07372dabf4aeb64a5c664b879a4

acmcarther: If this only occurred after adding L32 (the set), remove "-" from the initial "{%". You might be removing a required newline inside of this for loop. Seems likely that the spacing among these partials is inconsistent enough that you might be fine removing that extra newline in most cases, but not the "[.., unsupported, rust_binary, ...]" case.

mstg: I fixed this by adding a new empty line on top of `rust_binary.template`. I'll add those changes together with merged #93.

acmcarther:
Merged #95

mstg:
Think I'll just block this for a short moment, when converting all target names to lowercase rust can't find the required crates (when using `extern crate`)

I hunted the problem down to `slugify` in `util.rs` line 154.
Reverting the `lower` in the template and removing slugify keeps casing and actually fixes the problem (since slugify converts everything to lower case)
The problem is I don't know the implications of removing slugify, so perhaps you could chime in and advise on the best solution.

Thanks.

mstg:
I've added the proposed changes so you can take a look and to check if any tests break.

acmcarther:
Apologies, I don't understand the state of play here and have a hard time understanding if changing the slug behavior is right. I suspect not (primarily because slugify does weird stuff to avoid some characters, out of scope of this PR).

For the sake of clarity, can we restate the status quo and expected behavior in a structured way for ease of understanding the problem and solution? Here's my swing at this, please quote me and steal and update the tables.

|                 |Actual Crate Name (crates.io)|Generated Target Name (name = "NAME")|Target Name in Cross references (//dep:NAME)|
|-----------------|-----------------|-------------------------------------|--------------------------------------------|
|Current Behavior |MacTypes_sys     |"mactypes_sys" (?)                     |//...:mactypes_sys (?)                            |
|Expected Behavior|N/A              |(?)                                  |(?)                                         |

|"System"   |Path Case Sensitive|
|-----------|-------------------|
|Bazel Label|No (?)             |
|crates.io  |Yes (?)            |
|UNIX FS    |Yes                |
|WSL FS     |No (?)             |

mstg:
I'll try to explain the best I can.
I'll use the `system-framework-sys` crate as an example as that is the only case I have for this problem.

### Current revision behaves like this:
* Mixed case crates create a target with the same name. So if the crate name is `MacTypes-sys` the target name is `MacTypes_sys`.
* When running raze the `system-framework-sys` depends on `//...:mactypes_sys` and not `//...:MacTypes_sys` which caused this error:
> no such target '//src/bycrates.io/cargo/vendor/MacTypes-sys-2.1.0:mactypes_sys': target 'mactypes_sys' not declared in package 'src/bycrates.io/cargo/vendor/MacTypes-sys-2.1.0' (did you mean 'MacTypes_sys'?)

### The first proposed change was behaving like this:
* Mixed case crates create a target with the same name but in lower case. So if the crate name is `MacTypes-sys` the target name is `mactypes_sys`.
* Bazel would not cause any errors, but rust failed with errors like `extern crate MacTypes_sys is not found` because of the inconsistent casing

### The following happens when slugify is removed:
* Mixed case crates create a target with the same name and casing. So if the crate name is `MacTypes-sys` the target name is `MacTypes_sys`
* Target in dependencies is also added as `//...:MacTypes_sys`
* Bazel does not cause any errors and rust can find the crates with `extern crate`

mstg:
Here is the table updated for a better overview:

||Actual Crate Name (crates.io)|Generated Target Name (name = \"NAME\")|Target Name in Cross references (//dep:NAME)|Crate name (rustc)|
|-----------------|-----------------------------|-------------------------------------|--------------------------------------------|------------------|
|Current Behavior |MacTypes-sys                 |\"MacTypes_sys\" (?)                   |//...:mactypes_sys (?)                      |MacTypes_sys(?)  |
|Expected Behavior|MacTypes-sys                          |\"MacTypes_sys\"                                  |//...:MacTypes_sys                                         |MacTypes_sys               |

|\"System\"   | Case Sensitive|
|-----------|-------------------|
|Bazel Label|Yes (?)             |
|crates.io  |Yes (?)            |
|UNIX FS    |Yes                |
|WSL FS     |No (?)             |
|rustc  | Yes [crate name] (?) |

acmcarther:
Wow, that is enlightening. It's news to me that labels are case sensitive in Bazel. Thank you for reporting this and your patience in explaining it.

I've done some digging of my own to build confidence that this is the right thing to do.

1. "Data Lineage" of rust_library.name
target_name_sanitized 
-> BuildableTarget.name | replace("-", "_") 
-> CrateCatalogEntry.package.name (Package.name)
-> "opaque cargo internals"

TL;DR: the only postprocessing on target names is "-" -> "_"

2. "Data Lineage" of rust_library.deps
dependency.buildable_target
-> BuildableDependency.name
-> CrateCatalogEntry.workspace_path_and_default_target
-> workspace_path + package_ident + sanitized_name (sanitized_name is what's interesting here)
-> util::sanitize_ident(&package.name) (=slugify(package.name))

What you have here replaces the implementation of sanitize_ident, but I think that's actually impacting more than what we want. To converge these implementations, we just want "sanitized_name" to be equivalent to Package.name with "-" replaced with "_". 

I suggest replacing the assignment of sanitized_name in CrateCatalogEntry.new with a simpler replace op. I'm running [this change](https://github.com/google/cargo-raze/compare/acm-speculative-label-fix) [against CI](https://travis-ci.org/google/cargo-raze/builds/525226352?utm_source=github_status&utm_medium=notification) to see if it breaks anything. I think it might fix the problem without any of the other changes. I will follow up here either way.

Gotta say this handling of identifiers is really sloppy. Mea culpa.

acmcarther:
All right, so I "experimentally verified" this.

**/tmp/example/cargo/Cargo.toml**
```
[package]
name = "example"
version = "0.1.0"

[dependencies]
# This is notable. MacTypes-sys dep was removed in later versions
security-framework-sys = "0.2.2"

[lib]
path = "fake_lib.rs"

[raze]
workspace_path = "//cargo"
```

**/tmp/example/WORKSPACE**
```
workspace(name = "example")

load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")

git_repository(
    name = "io_bazel_rules_rust",
    commit = "5894d35bb7b5f982478dfbf71bc411426fae3451",
    remote = "https://github.com/bazelbuild/rules_rust.git",
)

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "bazel_skylib",
    sha256 = "eb5c57e4c12e68c0c20bc774bfbc60a568e800d025557bc4ea022c6479acc867",
    strip_prefix = "bazel-skylib-0.6.0",
    url = "https://github.com/bazelbuild/bazel-skylib/archive/0.6.0.tar.gz",
)

load("@io_bazel_rules_rust//:workspace.bzl", "bazel_version")

bazel_version(name = "bazel_version")

load("@io_bazel_rules_rust//rust:repositories.bzl", "rust_repositories")

rust_repositories()
```

**Vendor + Raze**
```
$ cd /tmp/example/cargo
$ cargo vendor -x
$ cargo raze
```

Using [master] raze, this happens:

```
$ bazel build //cargo/...
DEBUG: Rule 'io_bazel_rules_rust' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1547098113 -0500"                                                                                                                       
ERROR: /tmp/example/cargo/vendor/security-framework-sys-0.2.3/BUILD:28:1: no such target '//cargo/vendor/MacTypes-sys-2.1.0:mactypes_sys': target 'mactypes_sys' not declared in package 'cargo/vendor/MacTypes-sys-2.1.0' (did you mean 'MacTypes_sys'?) defined by /tmp/example/cargo/vendor/MacTypes-sys-2.1.0/BUILD and referenced by '//cargo/vendor/security-framework-sys-0.2.3:security_framework_sys'
ERROR: Analysis of target '//cargo/vendor/security-framework-sys-0.2.3:security_framework_sys' failed; build aborted: Analysis failed
INFO: Elapsed time: 0.086s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
```

Using that branch, we get
```
$ bazel build //cargo/...
DEBUG: Rule 'io_bazel_rules_rust' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1547098113 -0500"                                                                                                                       
INFO: Analysed 5 targets (6 packages loaded, 180 targets configured).
INFO: Found 5 targets...
INFO: Elapsed time: 1.183s, Critical Path: 0.99s
INFO: 3 processes: 3 processwrapper-sandbox.
INFO: Build completed successfully, 4 total actions
``` 

OK. So, I have a suggestion on how to move forward here. Not a good one -- that identifier code needs to be revisited for sure -- but we can get this fix submitted and a "test" added to ensure this keeps working.

1. Update this PR change to match [this](https://github.com/google/cargo-raze/compare/acm-speculative-label-fix)
2. Add "security-framework-sys = "0.2.2"" to deps of smoke_test/vendored/complicated_cargo_library/cargo/Cargo.toml and optionally smoke_test/remote/complicated_cargo_library/cargo/Cargo.toml. Running `cargo generate-lockfile` in both of those directories would be best, but it should work either way.

Ideally this could be captured in a simple unit test, but the test suite here isn't great. I need to build internal test libs for generate mock directories and mock crates for testing against for scenarios like this instead of just chucking more stuff into the smoke test.

mstg:
Just merged the previous changes from master and also added advised changes. If the tests pass, I'll rebase it into one commit.

mstg:
I've since rebased and this PR can be summed up like this:
* Target names are not slugified anymore
* Remote target names are not slugified anymore
* Casing is no longer inconsistent in target and deps
* Bazel does not error out because of inconsistent casing
* Rust does not error out because of inconsistent casing

All systems are behaving like expected and tests are passing. If there is anything else please advice.

acmcarther:
LGTM, thanks again for fixes and followup. Will cut 0.2.5 including this fix.

PR #106, opened 2019-07-05
Fix broken link in README.md

ian-h-chamberlain:
It looks like the code was refactored into the `impl` directory at some point, but the README link was never updated.


acmcarther:
Thanks, LGTM

PR #115, opened 2019-10-26
Fix a invalid function name in README.md

himanoa:
FYI. https://github.com/google/cargo-raze/blob/master/examples/WORKSPACE#L27-L28

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,9 +144,8 @@ $ cargo raze
 
 Finally, invoke the remote library fetching function within your WORKSPACE:
 ```python
-load("//cargo:crates.bzl", "raze_fetch_remote_crates")
-
-raze_fetch_remote_crates()
+load("//cargo:crates.bzl", "complicated_cargo_library_fetch_remote_crates")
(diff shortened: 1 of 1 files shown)


himanoa:
@googlebot I signed it!

acmcarther:
The name of this function is actually indirectly configurable via `raze.gen_workspace_prefix`. The value you see here is is from the `gen_workspace_prefix` setting of the "complicated_cargo_library" example:
https://github.com/google/cargo-raze/blob/3b9cb84d1c254140b93182e07a4ddb61dac51470/smoke_test/remote/complicated_cargo_library/cargo/Cargo.toml#L16
 
When "gen_workspace_prefix" is omitted, the default value "raze" is used.

Here's the actual link in the template:
https://github.com/google/cargo-raze/blob/3b9cb84d1c254140b93182e07a4ddb61dac51470/impl/src/templates/remote_crates.bzl.template#L17

---

I suspect this functionality is not useful and actually just confusing. Maybe I should remove it and just have the function have a consistent name?

PR #118, opened 2019-11-29
Fix docopt's usage

ok32:
Otherwise it ignores cmd line options


acmcarther:
Apologies for the delay getting back on this. The fix is much appreciated!

(Merging now)

Issue #119, opened 2019-11-29
proc-macro aliases aren't generated


PtrTeixeira:
Hey! On upgrading from 0.2.8 to 0.2.10, it looks like aliases for proc-macro crates stopped being generated. I think this was originally reported as #89 and was fixed by #90 , but #109 appears to have added the same constraint back. If you want I can open a PR to remove the check, but should [this check](https://github.com/google/cargo-raze/blob/105c64783745e97b883b6ed2d7bf85571a115e3c/impl/src/planning.rs#L445) include proc_macro crates as well?

Thank you!

PtrTeixeira:
Closed in #120

PR #120, opened 2019-12-10
[Planning] Allow proc-macros to export (#119)

GregBowyer:
This is a minor fix for #119, rather than edit the template (which seems to get fixed and unfixed) we allow the lib to be a proc-macro as well as an rlib

### DIFF
--- a/impl/src/planning.rs
+++ b/impl/src/planning.rs
@@ -442,7 +442,7 @@ impl<'planner> CrateSubplanner<'planner> {
     let mut lib_target_name = None;
     {
       for target in &targets {
-        if target.kind == "lib" {
+        if target.kind == "lib" || target.kind == "proc-macro" {
           lib_target_name = Some(target.name.clone());
           break;
         }


acmcarther:
Thanks for sending out this quick fix, and my apologies on the breakage in  https://github.com/google/cargo-raze/issues/119. Change LGTM, but in doing due diligence, dumb question follows.

The original assumption of this line is that (the list of targets in a crate) could only contain one lib, and thus we could safely assume that that target was the "authoritative single target" for the lib. This change tweaks that assumption with the following two additional assumptions:

1. A proc-macro (target contained in a) crate cannot contain a normal lib target.
2. A proc-macro (...) crate will only contain one proc-macro target.

Are these both safe assumptions to your knowledge? If you're not sure, I can check but I don't know either.

GregBowyer:
Right now it seems that you can only have a single lib target in cargo (and proc-macro is a type of lib as I understand it)

This is the best reference I found, might be wrong

https://internals.rust-lang.org/t/how-about-changing-lib-to-lib-to-allow-multiple-library-in-a-crate/2022

acmcarther:
Sorry for the delay on this. 

[Your link](https://internals.rust-lang.org/t/how-about-changing-lib-to-lib-to-allow-multiple-library-in-a-crate/2022) is good enough for me.

LGTM.

Issue #116, opened 2019-11-08
Typenum 1.11.2 build failure


klochek:
As of version 1.11.2, the [Typenum](https://github.com/paholg/typenum) library no longer builds, due to the following lines in [src/lib.rs](https://github.com/paholg/typenum/commit/e23e8af8c85c5c60e3829940d2cdcb450072952e#diff-b4aea3e418ccdb71239b96952d9cddb6R65):
```

include!(env!("TYPENUM_BUILD_OP"));
include!(env!("TYPENUM_BUILD_CONSTS"));
```

Those environment variables are set as part of the build.rs pre-compile; here is one such example (from [build/op.rs](https://github.com/paholg/typenum/commit/e23e8af8c85c5c60e3829940d2cdcb450072952e#diff-8cebe1bb9087438769a4e6f2849f2fdaR21)):

```
let dest = ::std::path::Path::new(&out_dir).join("op.rs");
println!("cargo:rustc-env=TYPENUM_BUILD_CONSTS={}", dest.display());
```

So, the authors are putting the paths to the generated `OUT_DIR` source files in those vars.

I could imagine a fix whereby the raze entry in the Cargo.toml can accept hard-coded environment vars, but I don't see any way to pass those on to the `rust_library` rule (my understanding of Bazel is still naive, but I'm getting the feeling that allowing users to specify arbitrary env vars might lead to Bad Things.)

Alternatively, maybe the right thing is just to fix this library?

acmcarther:
Thanks for filing this issue.

---

_On build script environment variables_

Yeah, cargo-raze doesn't currently support configuration options for environment variables. I have the same understanding you do -- that doing this is somewhat awkward in Bazel. Build scripts by necessity need to set some environment variables though, and some hacks already exist to set some of them.

If we're willing to extend the hackiness, we could support this partially by accepting a new Cargo.toml raze hashmap parameter which gets directly copied into the generated build script. This would be pretty janky and could only support configuration known ahead of time though. Allowing the crate to compile at all is an improvement over the status quo though.

Here's where those environment variables are set:
https://github.com/google/cargo-raze/blob/3b9cb84d1c254140b93182e07a4ddb61dac51470/impl/src/templates/partials/build_script.template#L36-L51

These don't propagate to the rust_library though, as you noted.

---

[...]

PR #121, opened 2019-12-11
Remote patches

GregBowyer:
So this is a wip and might not be what we want.

This adds in the ability to control the various `patch` flags for remote args as per upstream bazel.
This is a weapon that can be used for minor upstream fixing such as doing the following

```python
[raze.crates.typenum.'1.11.2']
patches = ["@bedrock//third_party/rust/patch:typenum-1.11.2-env-vars.patch"]
patch_args = ["-p1"]
```

which works around #116 
I wouldnt see this as the ultimate solution, but a workable version.


acmcarther:
Sorry for the ridiculous delay in looking at this. Working through the backlog of issues and PRs now.

---

This looks good to me and is superior to my current approach of cloning and manually patching the dependency. It enables much easier sharing of fixes as well.

You call this PR "WIP". Are there additional changes you had in mind?

GregBowyer:
I guess somehow doing a test for our illustrious CI might be good

GregBowyer:
I am going to hold off on tests right now and take a look at something like insta, but I think we should get this in

acmcarther:
Yeah, lack of tests is a bummer, but I understand.

PR #122, opened 2019-12-17
Update raze

GregBowyer:
In working through needing to add rename support I bumped into raze needing some spring cleaning.

This is a round of spring cleaning (currently ontop of #119) but I can rebase that does the following:

* Migrate to rust 2018
* Update cargo as a dep
* Update tera as a dep
* Do some `rustfmt` and `cargo clippy` runs


impl/.rustfmt.toml
```diff
@@ -0,0 +1 @@
+tab_spaces = 2
```
acmcarther: Sorry for this. Honestly this two space thing is a hold over I've had from long long ago back before four spaces was a strong standard. I've intended to make a pass through and fix everything to four spaces, but I've been lazy.

acmcarther:
Looks like the cargo install step needs to be updated to the new syntax.

Problematic line is here: https://github.com/google/cargo-raze/blob/8a75aa024a70a60fbf3c4a847078df495061a5f0/.travis.yml#L40

impl/.rustfmt.toml
GregBowyer: I figured it was more a google culture thing, I am agnostic and can convert it to 4 spaces (maybe in a followup PR as it will be a reformat)

I only added in rustfmt because I try to avoid all dogma on formatting, even if I _personally_ dont like some of its rules the fact that its the acceptable standard makes the silly arguments go away.

GregBowyer:
Do we want to merge this, or is there outstanding here

acmcarther:
No outstanding changes, looks great to me. I'll merge this and review the followup now.

PR #123, opened 2019-12-19
Allow crate aliases

GregBowyer:
Allow for crates to be "renamed" in Cargo.toml, and expose these details to rules_rust (PR pending)

This is needed for futures to be built with `compat` as a feature, but could also be useful to alias out multiple versions of a package in a tree.

PR #127, opened 2020-01-30
Semver matching

GregBowyer:
This is a little WIP, as we / I need to add unit and integration tests for this.

In reworking #123 a little, to allow for aliases to be presented in the "root" raze crate and to avoid a problem with crates presenting multiple versions of aliases I added in _basic_ support for using the semver library to do matches for alias determination.

It occured to me, that rather than _just_ use semver in that one specific place we should expose the richness of the semver objects into raze more generally. As such this commit exposes semver to raze and allows it to use semver for both version details and in matching configuration settings.

# Semver Matching
Raze settings come in the form of a [TOML] section like so

```toml
[raze.crates.some-crate.'1.2.3']
some_new_setting = true
```

Where the version number (in the above example `1.2.3`) is a literal, hardcoded
value.

This works but is a little inflexible for dependencies, _especially_
dependencies that are transiant and update a lot (for instance `syn`).

Since Cargo follows [Semver] we can use this fact, along with the richness of
the exported and serialisable `semver::Version` types to perform section
matching with [Semver] in the same fashion as Cargo itself.

This means that the above example can be written with semver expressions, for
example it can be written as:

```toml
[raze.crates.some-crate.'1.2.*']
some_new_setting = true

[raze.crates.some-crate.'~1.2.3']
some_new_setting = true

[raze.crates.some-crate.'^1.2.3']
some_new_setting = true

[raze.crates.some-crate.'=1.6.6']
some_new_setting = true
```

_Note_: Bare versions follow the semantics _as if they had_ specified a `^`,
which is in keeping with [Cargo semver semantics] but is distinct from previous
behaviour of raze (in which these would be interpreted as exacting or `=`).

This is deliberate as we should aim to mirror the semantics of cargo as much as
possible to avoid confusion.

We presently do not allow for multiple matches in raze settings, this is
intentional and is presented to the end user as an error.

[TOML]: https://github.com/toml-lang/toml
[Semver]: https://semver.org/spec/v2.0.0.html
[Cargo semver semantics]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

### DIFF
--- a/impl/src/planning.rs
+++ b/impl/src/planning.rs
@@ -137,7 +137,7 @@ mod tests {
   use super::*;
   use cargo_metadata::PackageId;
   use httpmock::MockServer;
-  use semver::Version;
+  use semver::{Version, VersionReq};
 
   // A wrapper around a MetadataFetcher which drops the
   // resolved dependency graph from the acquired metadata.
@@ -613,11 +613,14 @@ mod tests {
       )
       .unwrap();
 
+    let version = Version::parse("3.3.3").unwrap();
+
     // We expect to have a crate context for the binary dependency
     let context = planned_build
       .crate_contexts
       .iter()
-      .find(|ctx| ctx.pkg_name == "some-remote-crate" && ctx.pkg_version == "3.3.3")
+      .inspect(|x| println!("{}{}", x.pkg_name, x.pkg_version))
+      .find(|ctx| ctx.pkg_name == "some-remote-crate" && ctx.pkg_version == version)
       .unwrap();
 
     // It's also expected to have a checksum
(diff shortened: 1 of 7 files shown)


acmcarther:
This feature looks good to me, and is more user friendly than the original implementation. I originally justified my laziness here with the assumption that overrides/settings would need to be tweaked between versions, but in practice I haven't seen this be true and it's a nuisance to manually bump these versions.

Can you rebase this off master now that https://github.com/google/cargo-raze/pull/121 is merged?

acmcarther:
I'm not sure if this is waiting on me, or is stuck on something else. 

Is this ready to submit? Looks like it needs a rebase.

GregBowyer:
I was going to write some unit tests but had not got around to it

impl/src/planning.rs
```diff
@@ -461,14 +454,41 @@ impl<'planner> WorkspaceSubplanner<'planner> {
...
+      .crates
+      .get(&package.name)
+      .map_or(Ok(None), |settings| {
+        let mut versions = settings
+          .iter()
+          .filter(|(ver_req, _)| ver_req.matches(&package.version))
+          .peekable();
+
+        match versions.next() {
+          None => Ok(None),
```
dfreese, line 478: In what case would this actually occur?

Looking at the unwrap_or_default on line 535, it seems like this would be an error, but maybe I'm not understanding a particular case.

GregBowyer: I cant think of how, although I guess its habit for me to handle Option without `unwrap` (because you know code maintenance)

I will poke at it a bit and see if I can figure out if needed

UebelAndre:
@GregBowyer Hey, is this still relatively close to being done? 

Additionally, would this allow for me to also specify the following?
```toml
[raze.crates.some-crate.'*']
some_new_setting = true
```

GregBowyer:
It would / it does.

Personally I always do that with a range (sometimes crates remove / alter things a glob can mess up), but nothing is stopping that usage.

Its waiting on me not being lazy :S I will try to find some time this week.

UebelAndre:
@GregBowyer Thanks! Really looking forward to this change getting in 😄

UebelAndre:
@GregBowyer Hey, just a friendly poke? Do you think you'll have some time this week for this pull request? 😅

GregBowyer:
@UebelAndre its my top thing today (sorry took a while). Since you have built out an awesome test framework I can make good headway with ... some actual tests :P

impl/src/planning.rs
GregBowyer: This is specific to times where we dont match (no match there is a use the default, but that `unwrap_or_default` highlighted does not provide a default config here)

GregBowyer:
Ok unit test pushed so we can take a review

impl/src/planning.rs
```diff
@@ -645,13 +648,193 @@ mod tests {
...
+    let toml_file = indoc! { r#"
+    [package]
+    name = "semver_toml"
+    version = "0.1.0"
+
+    [lib]
+    path = "not_a_file.rs"
+
+    [dependencies]
+    # This has not settings
```
dfreese, line 672:
```suggestion
    # This has no settings
```

impl/src/planning.rs
```diff
@@ -645,13 +648,193 @@ mod tests {
...
     assert!(planned_build
       .crate_contexts
       .iter()
-      .find(|ctx| ctx.pkg_name == "wasm-bindgen-cli" && ctx.pkg_version == "0.2.68")
+      .find(|ctx| ctx.pkg_name == "wasm-bindgen-cli" && ctx.pkg_version == wasm_version)
       .is_none());
   }
+
+  #[test]
+  fn test_semver_matching() {
```
dfreese, line 662: can this be broken up into a couple, more specific, tests?

dfreese (approved):
LGTM after a rebase and some very picky nits.

impl/src/planning.rs
GregBowyer: I would like to just commit as I am getting exhausted with this PR

impl/src/planning.rs
GregBowyer: Fixed thanks

impl/src/planning.rs
dfreese: fair enough

dfreese (approved):

dfreese:
Thanks for putting this together!  Assuming CI passes, I'll go ahead and release this under 0.7.0

dfreese:
> Thanks for putting this together! Assuming CI passes, I'll go ahead and release this under 0.7.0
```
The command "if [ -n "$(git status --porcelain)" ]; then 
  echo '/examples is out of date. Please run `./smoke-test.sh` locally and commit the changes' >&2
  exit 1
fi" exited with 0.
````
Looks like the examples need to be updated.

GregBowyer:
Just git add examples/vendored?

UebelAndre:
The vendored source should be ignored 

`git stage examples` should get you the relevant new things I think

GregBowyer:
Humm after running smoke-test then I get nothing for examples

UebelAndre:
Oh, that's not the error.

`cargo-raze` appears to have failed to build tests

GregBowyer:
Weird ... pushed the fix

dfreese (approved):

dfreese (approved):

dfreese:
Released in 0.7.0

UebelAndre:
@dfreese how did you deploy? Isn't there normally a commit associated with a deploy?

dfreese:
> @dfreese how did you deploy? Isn't there normally a commit associated with a deploy?

1b7e178d1b0fca18f46ad5d689200ff10d9c7b48.  Forgot to push the commit.

Issue #132, opened 2020-02-10
Build time


sayrer:
In working on

https://github.com/dtolnay/cxx/pull/32

we noticed `cargo install cargo-raze` is very slow. While I think Cargo should be able to install pre-built binaries, is there some way to speed this up in the short term?

luser:
The killer is the dependency on `cargo` itself:
https://github.com/google/cargo-raze/blob/818ca547e6febf6ea973669ae341c00986425a9a/impl/Cargo.toml#L27

It looks like `cargo-raze` is actually using cargo internals so I don't see a feasible way to replace that. If it were just the metadata there's a [`cargo-meta`](https://crates.io/crates/cargo-meta) crate that provides a nice API over the `cargo metadata` command without pulling in all of cargo itself.

When I was at Mozilla a colleague implemented [cargo's `--build-plan` support](https://doc.rust-lang.org/cargo/reference/unstable.html#build-plan) which is probably *almost* what you want here except [apparently it has a bunch of problems that need to be addressed](https://github.com/rust-lang/cargo/issues/7614) (and it might need to be redesigned for a v2). Given that you have a working tool here it'd probably be useful for you to note what the existing build plan functionality is lacking compared to the method you currently use.

acmcarther:
Sorry for the delay https://github.com/google/cargo-raze/issues/132#issuecomment-585001351 is exactly correct. The slowness is compiling cargo.

IIRC, the dependence on cargo's internals is already replacable with a call out to `cargo metadata`. The ["MetadataFetcher" trait](https://github.com/google/cargo-raze/blob/818ca547e6febf6ea973669ae341c00986425a9a/impl/src/metadata.rs#L45-L47) was actually created with this purpose and a "CargoSubcommandMetadataFetcher" was written to be used in place of cargo internals. I think both codepaths existed at the time of writing because the cargo metadata subcommand was new... or maybe there was something missing from the generated metadata.

There's been some development since I originally started this work, but the ultimate path forward would be to drop the "CargoInternalsMetadataFetcher" and push forward with the subcommand fetcher. Any other dependencies on cargo are incidental and could be replaced by local types as has already been done with the metadata types. 

I'm a bit busy right now, but I'll take a look at this this weekend. If this issue is super stale and I've dropped the ball, ping back here again when you're feeling impatient. Alternatively, send a pull request.

sayrer:
I worked on this a bit in #142, but it's not ready to merge.

dfreese:
[...]

Issue #61, opened 2018-09-22
Build with bazel?


pcj:
Given that this project is a bazel build file generator, seems like it should build with bazel.  Specifically, I'd like to pull down and reference a `rust_binary` without having to use crates.io / cargo itself!.

acmcarther:
Similar issue from a while ago:
https://github.com/acmcarther/cargo-raze/issues/14

Historical blocker has been that this crate transitively depends on SSL crates (via Cargo itself), which require manual tweaks. Some thought also would need to go into what the bootstrapping behavior should look like. I've punted on this because I'd like a more general/durable approach for patching generated BUILD files.

I'm interested and historically have been interested in a feature like this because it enables using cargo-raze within a genrule or repository rule, which might be interesting.

mfarrugi:
@pcj is vendoring (a release of) the binary sufficient? 

Is your goal for a user to be able to build a repository with just an installation of bazel, or something else?

pcj:
Well, I think what I'd like to do is use cargo-raze in a similar manner to bazel-gazelle.  In that scenario, a user declares a dependency on `io_bazel_rules_go`, a dependency on `bazel_gazelle`, and then has a `gazelle()` rule in a `BUILD` file that, when run, modifies the `BUILD` files in the workspace.  It has to do a realpath trick to break out of the sandbox and figure out the actual workspace directory.

Analogously, if I had an existing dependency on `io_bazel_rules_rust` and `cargo_raze`, I could probably build a `cargo_raze()` rule that I could invoke as `bazel run //:cargo-raze` and have it (a) build the cargo-raze tool from source (if needed), and then (b) use that tool to update the dependencies in a `Cargo.toml` file.

Granted, I suppose the ergonomics aren't that different from the existing workflow, but anyway that's what I'm thinking.  And it spares me from having to have rust and cargo already on my system.

qzmfranklin:
Agree with @pcj on making cargo-raze be similar to bazel-gazelle in terms of bootstrapping. That is more general.

Looks like this issue has been stale for a while. Is there any initiative from within Google that does this?

acmcarther:
There isn't any existing effort to make this happen. This project is a 20% effort (largely less than that recently), and community input and contributions are very welcome.

[...]

PR #134, opened 2020-02-16
Use `cargo-metadata` by default (instead of cargo internals)

acmcarther:
This PR updates the implementation to use cargo metadata instead of compiled-in cargo internals.

It adds some backup flags in case this breaks some folks. 
-- `--cargo-bin-path` Lets you set the path to the cargo binary used for running cargo-metadata
-- `--deprecated-use-cargo-internals` Forces usage of the cargo internals implementation

I'm going to up the minor version for this change.

cc: https://github.com/google/cargo-raze/issues/132

This doesn't fix that bug since it doesn't remove the cargo dep, but it opens the door to fixing it after this has been released for a bit. If folks don't have issues, we can drop the "internals" impl, and remove lingering usages of the cargo dependency


GregBowyer:
@acmcarther fwiw this might be a route towards #61

Issue #139, opened 2020-03-02
b3sum binary raze configuration is ignored


vitalyd:
I've vendored the [b3sum](https://crates.io/crates/b3sum) crate, and trying to expose its `b3sum` binary via `cargo raze`.  However, raze appears to ignore it:

Cargo.toml:
```toml
[dependencies]
b3sum = "0.2.2"

[raze.crates.b3sum.'0.2.2']
extra_aliased_targets = [
    "cargo_bin_b3sum",
]
```
`cargo raze` execution does spit this out to `stdout`:
```
"b3sum": {
            "0.2.2": CrateSettings {
                additional_deps: [],
                skipped_deps: [],
                extra_aliased_targets: [
                    "cargo_bin_b3sum",
                ],
                additional_flags: [],
                gen_buildrs: false,
                data_attr: None,
                buildrs_additional_environment_variables: [],
                patch_args: [],
                patch_cmds: [],
                patch_cmds_win: [],
                patch_tool: None,
                patches: [],
            },
        },
```
However, it also generates this warning:
```
Found unused raze settings for all of b3sum-["0.2.2"]
```
We have a similar configuration for `ripgrep 0.10.0`, to generate the `rg` binary.  This used to work (at least the `ripgrep` variant) with an older `cargo` (back when `cargo-vendor` was its own plugin) and `cargo raze 0.2.8`.  I've been away from this part of our configuration for a while, and have updated to the latest `cargo`* and `cargo raze 0.2.15`.

\* - actually, I have to use the beta `cargo`, which is `1.42.0-beta`  because the `--versioned-dirs` option isn't available on current stable, `1.41.0 (626f0f40e 2019-12-03)`.

I'm not sure what exactly changed here as I'm not intimately familiar with raze's internals.  @acmcarther , any ideas?

By the way, I'm able to reproduce this issue by playing with `Cargo.toml` in the `examples/vendored/complicated_cargo_library/cargo`.

vitalyd:
Played around with this issue a bit more today.  Decided to try the `--deprecated-use-cargo-internals` flag to see if the difference is due to using `cargo`-as-a-lib vs parsing `cargo metadata ...`.

I immediately hit a panic:
```rust
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorMessage { msg: "unsupported source protocol: \"registry" }', src/libcore/result.rs:1188:5
```
[...]

PR #140, opened 2020-03-06
Fix incorrect package source id String in CargoInternalsMetadataFetcher

vitalyd:
This addresses the panic described in https://github.com/google/cargo-raze/issues/139#issuecomment-594678552.


acmcarther:
Hmm, alright then.

Thanks for this fix!

PR #142, opened 2020-03-16
Disentangle cargo internals.

sayrer:
This is a first pass at disentangling cargo internals from this crate. I wasn't sure how the maintainers would approach certain problems, but it does build and pass tests in both configurations.

### DIFF
--- a/impl/Cargo.toml
+++ b/impl/Cargo.toml
@@ -22,9 +22,10 @@ name = "cargo-raze"
 path = "src/bin/cargo-raze.rs"
 
 [dependencies]
+anyhow = "1.0"
 failure = "0.1.5"
 docopt = "1.0.2"
-cargo = "0.42.0"
+#cargo = "0.42.0"
 cargo-platform = "0.1.0"
 itertools = "0.8.0"
 slug = "0.1.4"
@@ -34,7 +35,13 @@ serde = "1.0.84"
 toml = "0.4.10"
 serde_json = "1.0.34"
 tempdir = "0.3.7"
+url = "2.0"
+
+cargo = { version = "0.42.0", optional = true }
 
 [dev-dependencies]
 lazy_static = "1.2.0"
 hamcrest2 = "0.3.0"
+
+[features]
+cargo_internals = ["cargo"]
--- a/impl/src/bin/cargo-raze.rs
+++ b/impl/src/bin/cargo-raze.rs
@@ -18,18 +18,22 @@ use std::{
   path::{Path, PathBuf},
 };
 
-use cargo::{util::Config, CargoResult, CliResult};
+#[cfg(cargo_internals)]
+use cargo::util::Config;
 
 use docopt::Docopt;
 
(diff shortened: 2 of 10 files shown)


sayrer:
@acmcarther wondering what you think here. This seemed to more than halve compilation time for the crate, at the expense of copying a little bit of code from cargo.

impl/Cargo.toml
```diff
@@ -34,7 +35,13 @@ serde = "1.0.84"
...
+url = "2.0"
+
+cargo = { version = "0.42.0", optional = true }
 
 [dev-dependencies]
 lazy_static = "1.2.0"
 hamcrest2 = "0.3.0"
+
+[features]
+cargo_internals = ["cargo"]
```
acmcarther, line 47: Does it make sense to make this be default enabled so that it doesn't break existing users?

impl/src/bin/cargo-raze.rs
```diff
@@ -65,44 +69,60 @@ Options:
...
+
+#[cfg(not(cargo_internals))]
+fn create_fetcher(options: &Options) -> Box<dyn MetadataFetcher> {
+  create_subcommand_fetcher(options)
+}
 
-  if let Err(e) = result {
-    cargo::exit_with_error(e, &mut *config.shell());
+#[cfg(cargo_internals)]
+fn create_cargo_internals_fetcher(options: &Options) -> Box<dyn MetadataFetcher> {
```
acmcarther, line 95: Can this function body be inlined into the create_fetcher for cargo_internals?

As is, the name is misleading (we might get a subcommand fetcher).

impl/src/bin/cargo-raze.rs
```diff
@@ -126,7 +146,10 @@ fn real_main(options: &Options, cargo_config: &mut Config) -> CliResult {
     GenMode::Remote => {
       // Create "remote/" if it doesn't exist
       if fs::metadata("remote/").is_err() {
-        fs::create_dir("remote/").map_err(failure::Error::from)?;
+        match fs::create_dir("remote/") {
```
acmcarther, line 149: Any reason to prefer the match over `map_err(|e| Err(anyhow::format_err!(...)))`?

impl/src/context.rs
```diff
@@ -53,6 +56,137 @@ pub struct SourceDetails {
...
 
+#[derive(Debug, Clone)]
+pub struct SourceId {
+  pub kind: SourceKind,
+  pub url: Url,
+  pub precise: Option<String>,
+}
+
+#[derive(Debug, Clone)]
+pub enum SourceKind {
```
acmcarther, line 67: Can you run rustfmt over this file? Looks like some of the file is using the proper 4 space scheme instead of my terrible 2 space scheme.

I really should fix that.

sayrer:
@acmcarther your comments are reasonable. I will try to address them shortly.

kornholi:
We merged very similar changes in #156

PR #145, opened 2020-04-21
Clean-up no-op conditional

smklein:


acmcarther:
Thanks for the cleanup, this LGTM.

Issue #147, opened 2020-04-24
Feature request: Utilize cargo_metadata for parsing Cargo metadata


smklein:
See title: https://docs.rs/cargo_metadata/0.9.1/cargo_metadata/index.html

I think this crate could be used for parsing metadata from cargo, instead of relying on a hand-rolled implementation in impl/src/metadata.rs.

Hopefully this would make tracking upstream Cargo changes slightly easier to follow?

What do ya'll think? Would this be a welcome change?

acmcarther:
Yeah, this sounds good to me. Duplicating cargo's structs as we're doing now isn't great.

I'll look into this this upcoming weekend, or feel free to take this. I think we use an interface for metadata fetching, so it should be reasonably low effort to drop in a cargo-metadata wrapped implementation (or just drop the abstraction entirely).

acmcarther:
Oh, I lied.

We actually already do this, and I think we recently flipped the default to use cargo-metadata. Here is the type:

https://github.com/google/cargo-raze/blob/7daf5ba22612b5807492aa568c2a7b71c8630d4e/impl/src/metadata.rs#L162-L165

I'm guessing your ask here is that we remove the old implementation. I'm not averse to this.

PR #150, opened 2020-04-27
Rely on cargo_metadata crate

smklein:
This patch is ready for review; feedback is welcome.

Attempting to use as much of https://docs.rs/cargo_metadata/0.9.1 as possible, rather than copies of those structs.

Overview of the changes:

- Most of "metadata.rs" has been deleted in favor of using the "cargo_metadata" crate explicitly. The different usage of types propagates through the rest of the codebase (e.g., "String" vs "semver::Version").
- As a result of using "cargo_metadata", many of the types may only be created via deserialization of cloning, rather than explicit construction. This required some test re-working to be compatible. The most significant impact of these changes is visible in "planning.rs", where the reliance on "dummy data" has generally been replaced by relying on real data (and cloning/modifying it as necessary to meet the demands of specific tests).


smklein:
Related to: https://github.com/google/cargo-raze/issues/147

impl/src/bin/cargo-raze.rs
```diff
@@ -54,8 +50,8 @@ Generate BUILD files for your pre-vendored Cargo dependencies.
 
 Usage:
     cargo raze (-h | --help)
-    cargo raze [--verbose] [--quiet] [--color=<WHEN>] [--dryrun] [--cargo-bin-path=<PATH>] [--deprecated-use-cargo-internals]
-    cargo raze <buildprefix> [--verbose] [--quiet] [--color=<WHEN>] [--dryrun] [--cargo-bin-path=<PATH>] [--deprecated-use-cargo-internals]
+    cargo raze [--verbose] [--quiet] [--color=<WHEN>] [--dryrun] [--cargo-bin-path=<PATH>]
```
acmcarther, line 53: Ideally the deletion of the old internals mode can be decoupled from the update of the cargo-metadata mode from being shell command based to using the dedicated crate.

(I'm going to assume this is implied in the rest of the code -- I won't comment on things that are part of the old mode and are being deleted here)

Is it possible to do that in this change? If not, I'll have to go make sure the feature gaps described in https://github.com/google/cargo-raze/pull/149#issuecomment-621493706 before I'm comfortable with this change being submitted.

impl/src/context.rs
```diff
@@ -69,7 +69,6 @@ pub struct CrateContext {
   pub targets: Vec<BuildableTarget>,
   pub build_script_target: Option<BuildableTarget>,
   pub source_details: SourceDetails,
-  pub sha256: Option<String>,
```
acmcarther, line 72: kornholi@ commented on the utility of this field here https://github.com/google/cargo-raze/pull/149#issuecomment-621493706 .

I actually forgot what this was for and I want to revisit this. I would be sad to see it deleted without remembering what it was for. No action item from you yet, just mostly a reminder for me to figure out what's going on here.

impl/src/metadata.rs
```diff
@@ -12,34 +12,21 @@
...
 
-pub type PackageId = String;
-pub type Kind = String;
-pub type TargetSpec = String;
-
 /**
  * An entity that can retrive deserialized metadata for a Cargo Workspace.
  *
- * The `CargoInternalsMetadataFetcher` is probably the one you want.
+ * The `CargoSubcommandMetadataFetcher` is probably the one you want.
```
acmcarther, line 29: It's probably safe to just drop this comment. It was more cautionary before when there was one "good" implementation and one "experimental" implementation.

(no reason to annotate every trait with the preferred implementation of the trait, in general)

impl/src/metadata.rs
```diff
@@ -211,254 +81,11 @@ impl MetadataFetcher for CargoSubcommandMetadataFetcher {
       dir
     };
 
-    // Shell out to cargo
-    let exec_output = Command::new(&self.cargo_bin_path)
+    MetadataCommand::new()
+      .cargo_path(&self.cargo_bin_path)
```
acmcarther, line 85: Wow I'm impressed that this just drops in here.

(no action item)

impl/src/planning.rs
```diff
@@ -136,7 +140,7 @@ impl CrateCatalogEntry {
     is_workspace_crate: bool,
   ) -> Self {
     let sanitized_name = package.name.replace("-", "_");
-    let sanitized_version = util::sanitize_ident(&package.version);
+    let sanitized_version = package.version.clone();
```
acmcarther, line 143: I think this is wrong, but your tests are passing so I'm really confused.

util::sanitize_ident was not just concerned with type conversions. It also [removed funny punctuation and converted dashes to underscores](https://github.com/google/cargo-raze/blob/8337c7b5113aad3de2038bc89ef23d47b82c35e6/impl/src/util.rs#L164-L166). I guess it's possible that we lack tests that cover this behavior.

Do you know if this is being done for us by cargo_metadata?

impl/src/planning.rs
```diff
@@ -248,19 +252,14 @@ impl CrateCatalogEntry {
 impl CrateCatalog {
   /** Produces a CrateCatalog using the package entries from a metadata blob.*/
   pub fn new(metadata: &Metadata) -> Self {
+    let resolve = metadata.resolve.as_ref().unwrap();
```
acmcarther, line 255: As a practice, I've tried to make sure my `unwraps()` have a justification as a comment above the invocation site. Is this known to be safe, or is there some reason why it's always safe in this case? Same for relow.

impl/src/planning.rs
```diff
@@ -568,26 +568,26 @@ impl<'planner> CrateSubplanner<'planner> {
...
         }
       }
 
-      match dep.kind.as_deref() {
-        None | Some("normal") => normal_dep_names.push(dep.name.clone()),
-        Some("dev") => dev_dep_names.push(dep.name.clone()),
-        Some("build") => build_dep_names.push(dep.name.clone()),
-        something_else => {
+      match dep.kind {
+        DependencyKind::Normal => normal_dep_names.push(dep.name.clone()),
```
acmcarther, line 581: Just checking, is this equivalent to the old code? Where "no" dependency kind results in inferring normal?

I think this is really a question of what cargo_metadata returns in that case.

(again, I think the tests are not covering this case....)

impl/src/metadata.rs
```diff
@@ -54,128 +41,11 @@ pub struct CargoWorkspaceFiles {
...
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-pub struct ResolveNode {
-  pub id: PackageId,
-  pub dependencies: Vec<PackageId>,
-  // Optional due to recent feature addition in Cargo.
-  pub features: Option<Vec<String>>,
-}
-
 /** A workspace metadata fetcher that uses the Cargo Metadata subcommand. */
 pub struct CargoSubcommandMetadataFetcher {
```
acmcarther, line 45: Nit, maybe this deserves a name change? 

Not really sure what to call it though. Maybe just CargoMetadataFetcher? That doesn't really distinguish the type very well, but I don't like the next best alternative (CargoCrateMetadataFetcher) either because "Crate" is ambiguous there.

No strong request here unless you have a suggestion (or you're OK with either just CargoMeta(..) or CargoCrateMeta(...).

impl/src/bin/cargo-raze.rs
smklein: I'm going to try to land #149 first - IMO, it simplifies the assumptions of this code - but please feel free to push back on me if you think that's the wrong direction!

impl/src/context.rs
smklein: Ack - this is preserved in the most recent update to #149 , where the value is retrieved from the Lockfile.

impl/src/metadata.rs
smklein: Done!

impl/src/planning.rs
smklein: cargo_metadata uses a semvar type for Version which I believe uses type-safety to prevent invalid punctuation: 
https://docs.rs/cargo_metadata/0.9.1/cargo_metadata/struct.Package.html
https://docs.rs/semver/0.9.0/semver/struct.Version.html

impl/src/planning.rs
smklein: Rather than justifying them, I've just replaced them with error handlers.

IMO it would be better to propagate an explicit error if "cargo metadata" returned malformed output, rather than panicking.

impl/src/planning.rs
smklein: Following-up here: I think you're 100% right! The smoke tests appear to be failing with this patch, and the difference is "dot" vs "underscore". I'll add a commit that fixes this.

smklein: This has been fixed, and test coverage has been added checking that the workspace path does not contain any "." or "-" characters.

impl/src/metadata.rs
smklein: \o/

acmcarther:
I'll cut a release that includes this CL immediately.

acmcarther:
This is in [release 0.3.0](https://github.com/google/cargo-raze/commit/548fa54732f19d40abda4b518f5f847ff70ff0d4)

PR #151, opened 2020-04-29
[smoke-test] Update rules_rust version

smklein:
This is required for bazel 3.1.0; otherwise
https://github.com/bazelbuild/bazel/issues/7793
causes build failures.


acmcarther:
Looks good, thanks for the fix!

acmcarther:
Looks good, thanks for the fix!

PR #152, opened 2020-04-30
Allow customizing generated BUILD files

### LINKED ISSUE #58: Add build_file_path [raze.crate.version] override that supports providing a custom build file
This is a pretty finicky proposition, all told. Dependencies are generally versioned and those versions can change due to non-local causes which would require manually updating the build file. One example of such a cause would be newly including a dependency that forces another dependency to downgrade.

The cleanest implementation of this would somehow allow the BUILD file author to refer to dependencies "at whatever version they would otherwise be" via an exported BZL file or something else.. An initial implementation could forego these niceties though.

cc: @mfarrugi

kornholi:
This adds a new crate option `additional_build_file` for appending custom content to the generated
BUILD files. For example, to include custom Bazel targets for building non-Rust parts of a crate.

I started out with an option to completely override the BUILD file, but I think that's more trouble than it's worth maintenance-wise. So far I haven't needed to change anything Raze generates besides adding to it (e.g. `cc_library` targets).

@mfarrugi - have you run into any cases where this wouldn't be sufficient?

Suggestions welcome on a better name. I was thinking of adding `additional_build_file_contents` to mirror Bazel's `build_file`/`build_file_contents`, but that seems too much.

Fixes #58.

### DIFF
--- a/impl/src/bazel.rs
+++ b/impl/src/bazel.rs
@@ -133,6 +133,28 @@ impl BazelRenderer {
   }
 }
 
+fn include_additional_build_file(
+  package: &CrateContext,
+  existing_contents: String,
+) -> CargoResult<String> {
+  match &package.raze_settings.additional_build_file {
+    Some(file_path) => {
+      let additional_content =
+        std::fs::read_to_string(file_path).map_err(|e| RazeError::Rendering {
+          crate_name_opt: Some(package.pkg_name.to_owned()),
+          message: format!("failed to read additional_build_file: {}", e),
+        })?;
+
+      Ok(format!(
+        "{}\n# Additional content from {}\n{}",
+        existing_contents, file_path, additional_content
+      ))
+    }
+
+    None => Ok(existing_contents),
+  }
+}
+
 impl BuildRenderer for BazelRenderer {
   fn render_planned_build(
     &mut self,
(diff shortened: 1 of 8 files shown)


mfarrugi:
This would have worked for the sys crates I mentioned, but haven't seen many more cases since I have not been a user for a while.

dfreese:
For what it's worth, I basically ended up on this strategy when building ring.

I just mentally transpiled the build.rs file's cc commands into a cc_library target.  This let me skip the build.rs file, and then add an additional dependency on ring for my custom cc_library.  Since cargo raze doesn't export files, or have this feature, I had to re-import the crate as an additional http_repository, and then add a custom build file.

acmcarther:
The current name SGTM. If we have another setting later with inlined context, it could be called "additional_build_contents". 

I'll cut a release that includes this change after reviewing the remaining pending PRs.

nikclayton-dfinity:
Re 

> So far I haven't needed to change anything Raze generates besides adding to it (e.g. cc_library targets).

https://github.com/google/cargo-raze/issues/269 is an example of an issue where this functionality would be useful. It seems to be a bug in `cargo-raze`, and until that bug is fixed it would be great if there was an easy way to say "I've got a /some/other/BUILD file that I know works, use this one in preference".

PR #153, opened 2020-04-30
Update cargo vendor usage in smoke test

kornholi:


acmcarther:
Looks good, thanks for the quick fix!

PR #156, opened 2020-05-14
Remove remaining dependencies on Cargo

kornholi:
This replaces `CargoResult` with anyhow's `Result`, and uses `SourceId` from cargo_lock instead.

As far as I can tell, we lose 101 dependencies total :)


kornholi:
> It feels like there are a lot of different "result" crates in the rust ecosystem...

Since all the `Error` trait changes landed, dtolnay's [anyhow](https://github.com/dtolnay/anyhow) and [thiserror](https://github.com/dtolnay/thiserror) have been getting lots of traction. In fact, the latest `CargoResult` is an alias to `anyhow::Result`.

dfreese:
This cut a notable amount of time off of building one of our containers with cargo raze in it.  Thanks @kornholi!

FYI: @smklein

PR #162, opened 2020-06-01
Fix: Alias renamed dependencies in generated build rules.

### LINKED ISSUE #143: Actix-web failed with: failed to resolve: use of undeclared type or module `fail_ure`
Hi, I'm trying to compile the following actix-web example in my Bazel project: https://github.com/actix/examples/tree/master/multipart
and I have ended up with the following error:

```
error[E0433]: failed to resolve: use of undeclared type or module `fail_ure`
   --> third_party/cargo/vendor/actix-http-1.0.1/src/error.rs:968:24
    |
968 | impl ResponseError for fail_ure::Error {}
    |                        ^^^^^^^^ use of undeclared type or module `fail_ure`

error: aborting due to previous error
```
The same code perfectly compiles using cargo...

Any ideas how to fix it?

Arm1stice:
This PR fixes #143 by checking for renamed packages while exploring dependencies. If a package has been renamed, it will be added to the `aliases` parameter in the generated build rule.

I am still fairly new to Rust programming and it's best practices, so please feel free to nit my implementation.

Once you approve of the implementation, I can write unit tests before merging.

### DIFF
--- a/impl/src/planning.rs
+++ b/impl/src/planning.rs
@@ -586,7 +586,7 @@ impl<'planner> CrateSubplanner<'planner> {
         if let Some(alias) = aliased_dep_names.get(&dep_package.name) {
           aliased_deps.push(DependencyAlias{
             target: buildable_target.clone(),
-            alias: util::sanitize_name(alias),
+            alias: util::sanitize_ident(alias),
           })
         }     
       }
--- a/impl/src/planning.rs
+++ b/impl/src/planning.rs
@@ -590,13 +590,12 @@ impl<'planner> CrateSubplanner<'planner> {
         normal_deps.push(buildable_dependency);
         // Only add renamed normal deps to the HashMap
         let package_alias = renamed_dep_names.get(&dep_package.name);
-        if package_alias.is_some() {
-          // UNWRAP: Safe from check above.
+        package_alias.as_ref().map(|alias| {
           renamed_deps.insert(
             buildable_target.clone(),
-            package_alias.unwrap().replace("-", "_"),
+            alias.replace("-", "_"),
           );
-        }        
+        });      
       }
     }
 
@@ -655,13 +654,12 @@ impl<'planner> CrateSubplanner<'planner> {
       }
 
       // Check if the dependency has been renamed
-      if dep.rename.is_some() {
-        // UNWRAP: Safe from above check
+      dep.rename.as_ref().map(|rename| {
         renamed_dep_names.insert(
           dep.name.clone(),
-          format!("{}", dep.rename.as_ref().unwrap()),
+          format!("{}", rename),
         );
-      }
+      });
     }
 
     Ok(DependencyNames {
--- a/impl/src/planning.rs
+++ b/impl/src/planning.rs
@@ -972,26 +972,33 @@ dependencies = [
     "
   }
 
-  fn make_basic_workspace() -> (TempDir, CargoWorkspaceFiles) {
+  fn make_workspace(toml_file: &'static str, lock_file: Option<&'static str>) -> (TempDir, CargoWorkspaceFiles) {
     let dir = TempDir::new("test_cargo_raze_metadata_dir").unwrap();
     let toml_path = {
       let path = dir.path().join("Cargo.toml");
       let mut toml = File::create(&path).unwrap();
-      toml.write_all(basic_toml().as_bytes()).unwrap();
+      toml.write_all(toml_file.as_bytes()).unwrap();
       path
     };
-    let lock_path = {
-      let path = dir.path().join("Cargo.lock");
-      let mut lock = File::create(&path).unwrap();
-      lock.write_all(basic_lock().as_bytes()).unwrap();
-      path
+    let lock_path = match lock_file {
+      Some(lock_file) => {
+        let path = dir.path().join("Cargo.lock");
+        let mut lock = File::create(&path).unwrap();
+        lock.write_all(lock_file.as_bytes()).unwrap();
+        Some(path)
+      },
+      None => None,
     };
     let files = CargoWorkspaceFiles {
-      lock_path_opt: Some(lock_path),
+      lock_path_opt: lock_path,
       toml_path,
     };
 
     (dir, files)
+  }
+
+  fn make_basic_workspace() -> (TempDir, CargoWorkspaceFiles) {
+    make_workspace(basic_toml(), Some(basic_lock()))
   }
 
   #[test]
@@ -1260,6 +1267,50 @@ dependencies = [
       PlatformDetails::new("some_target_triple".to_owned(), Vec::new() /* attrs */),
     );
     assert!(planned_build_res.unwrap().crate_contexts.is_empty());
+  }
+
+  #[test]
+  fn test_plan_build_produces_aliased_dependencies() {
+    let toml_file =     "
+    [package]
+    name = \"advanced_toml\"
+    version = \"0.1.0\"
+    
+    [lib]
+    path = \"not_a_file.rs\"
+
+    [dependencies]
+    actix-web = \"2.0.0\"
+    actix-rt = \"1.0.0\"
+        ";
+    let (_temp_dir, files) = make_workspace(toml_file, None);
+    let mut fetcher = WorkspaceCrateMetadataFetcher::default();
+    let mut settings = settings_testing::dummy_raze_settings();
+    settings.genmode = GenMode::Remote;
+
+    let mut planner = BuildPlannerImpl::new(&mut fetcher);
+    // N.B. This will fail if we don't correctly ignore workspace crates.
+    let planned_build_res = planner.plan_build(
+      &settings,
+      files,
+      PlatformDetails::new("some_target_triple".to_owned(), Vec::new() /* attrs */),
+    );
+
+    let crates_with_aliased_deps: Vec<CrateContext> = planned_build_res.unwrap().crate_contexts.into_iter().filter(|krate| krate.aliased_dependencies.len() != 0).collect();
+    
+    // Vec length shouldn't be 0
+    assert!(crates_with_aliased_deps.len() != 0, "Crates with aliased dependencies is 0");
+
+    // Find the actix-web crate
+    let actix_web_position = crates_with_aliased_deps.iter().position(|krate| krate.pkg_name == "actix-http");
+    assert!(actix_web_position.is_some());
+
+    // Get crate context using computed position
+    let actix_http_context = crates_with_aliased_deps[actix_web_position.unwrap()].clone();
+
+    assert!(actix_http_context.aliased_dependencies.len() == 1);
+    assert!(actix_http_context.aliased_dependencies[0].target ==  "@raze_test__failure__0_1_8//:failure");
+    assert!(actix_http_context.aliased_dependencies[0].alias == "fail_ure");
   }
 
   // TODO(acmcarther): Add tests:
--- a/impl/src/planning.rs
+++ b/impl/src/planning.rs
@@ -63,8 +63,8 @@ struct DependencyNames {
   build_dep_names: Vec<String>,
   // Dependencies that are required for tests
   dev_dep_names: Vec<String>,
-  // Dependencies that have been renamed
-  renamed_dep_names: HashMap<String, String>,
+  // Dependencies that have been renamed and need to be aliased in the build rule
+  aliased_dep_names: HashMap<String, String>,
 }
 
 // TODO(acmcarther): Remove this struct -- move it into CrateContext.
@@ -77,7 +77,7 @@ struct DependencySet {
   // Dependencies that are required for tests
   dev_deps: Vec<BuildableDependency>,
   // Dependencies that have been renamed and need to be aliased in the build rule
-  renamed_deps: HashMap<String, String>,
+  aliased_deps: Vec<String>,
 }
 
 /** An entry in the Crate catalog for a single crate. */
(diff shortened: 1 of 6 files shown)


dfreese:
Sounds fairly similar to #123.  I think that approach was okay, but he wanted to add further tests.  My reading of that thread was that approach was alright, if you want to compare.

impl/src/planning.rs
```diff
@@ -624,12 +653,22 @@ impl<'planner> CrateSubplanner<'planner> {
           )
         }
       }
+
+      // Check if the dependency has been renamed
+      if dep.rename.is_some() {
```
smklein, line 658: Might be more canonical to use:

```
dep.rename.map(|rename| {
  renamed_dep_names.insert(
    dep.name.clone(),
    format!("{}", rename),
  );
})
```

As this would avoid the unwrap

impl/src/planning.rs
```diff
@@ -572,6 +588,15 @@ impl<'planner> CrateSubplanner<'planner> {
 
       if normal_dep_names.contains(&dep_package.name) {
         normal_deps.push(buildable_dependency);
+        // Only add renamed normal deps to the HashMap
+        let package_alias = renamed_dep_names.get(&dep_package.name);
+        if package_alias.is_some() {
```
smklein, line 593: Same as comment below; a `map` might help avoid the unwrap.

impl/src/planning.rs
Arm1stice: Thank you for this! Looks like I had to add `.as_ref()` before `.map` to appease rustc.

impl/src/planning.rs
```diff
@@ -624,12 +652,21 @@ impl<'planner> CrateSubplanner<'planner> {
           )
         }
       }
+
+      // Check if the dependency has been renamed
+      dep.rename.as_ref().map(|rename| {
```
dfreese, line 657: Talked with @smklein about this a little later, so sorry for asking to change this again, but rust also has a [`if let`](https://doc.rust-lang.org/stable/rust-by-example/flow_control/if_let.html) pattern matching syntax.  That would probably be the clearest here.

```
if let Some(alias) = dep.rename.as_ref() {
  // everything in the closure
}
```

impl/src/planning.rs
```diff
@@ -624,12 +652,21 @@ impl<'planner> CrateSubplanner<'planner> {
...
+        renamed_dep_names.insert(
+          dep.name.clone(),
+          format!("{}", rename),
```
dfreese, line 660: can this format be shortened to rename.clone()?

impl/src/planning.rs
```diff
@@ -486,6 +491,13 @@ impl<'planner> CrateSubplanner<'planner> {
         }
       }
     }
+
+    // Take the hashmap and create a vector of strings that will be inserted into the template.
+    let mut aliased_dependencies = Vec::new();
+    for (name, alias) in &renamed_deps {
+      aliased_dependencies.push(format!("\"{}\": \"{}\",", name.clone(), alias.clone()));
```
dfreese, line 498: Are the clone() calls needed here, or could the variables be borrowed?  &name, &alias

impl/src/planning.rs
```diff
@@ -63,6 +63,8 @@ struct DependencyNames {
   build_dep_names: Vec<String>,
   // Dependencies that are required for tests
   dev_dep_names: Vec<String>,
+  // Dependencies that have been renamed
+  renamed_dep_names: HashMap<String, String>,
```
dfreese, line 67: could this be called `aliased_dep_names`, and then `aliased_deps` line 80?

I think that will make it a little more clear that aliases are being used the entire time.

impl/src/planning.rs
```diff
@@ -74,6 +76,8 @@ struct DependencySet {
   build_deps: Vec<BuildableDependency>,
   // Dependencies that are required for tests
   dev_deps: Vec<BuildableDependency>,
+  // Dependencies that have been renamed and need to be aliased in the build rule
+  renamed_deps: HashMap<String, String>,
```
dfreese, line 80: It doesn't look like you use the lookup into this hashmap.  Could this, instead, be a Vec of a struct, similar to [DependencyAlias](https://github.com/google/cargo-raze/pull/123/commits/fde76fdffe5fc645a41c4615486191b3330da27a#diff-b839b62c6331bbd8524ee98a856d496eR57-R61) here?

impl/src/planning.rs
```diff
@@ -572,6 +588,14 @@ impl<'planner> CrateSubplanner<'planner> {
 
       if normal_dep_names.contains(&dep_package.name) {
         normal_deps.push(buildable_dependency);
+        // Only add renamed normal deps to the HashMap
+        let package_alias = renamed_dep_names.get(&dep_package.name);
+        package_alias.as_ref().map(|alias| {
+          renamed_deps.insert(
+            buildable_target.clone(),
+            alias.replace("-", "_"),
```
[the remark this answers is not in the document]
dfreese: (Doing this replace could also probably be saved until the last `format!(..., util::sanitize_name(alias))` you make to build up the `aliased_dependencies` vector.

impl/src/planning.rs
Arm1stice: It was giving me compiler errors before for some reason, but now that we switched to the `if let` syntax I'll change it.

impl/src/planning.rs
Arm1stice: Sounds good. Changed to `if let`

impl/src/planning.rs
Arm1stice: Changed the names.

impl/src/planning.rs
Arm1stice: I went ahead and changed it to a Vec<String>. Because of this, I moved the logic that creates the alias strings for the template into `produce_deps`

impl/src/planning.rs
Arm1stice: This logic no longer lives in this function and now doesn't involve looping through a HashMap. Resolving.

impl/src/planning.rs
```diff
@@ -486,6 +491,13 @@ impl<'planner> CrateSubplanner<'planner> {
         }
       }
     }
+
+    // Take the hashmap and create a vector of strings that will be inserted into the template.
+    let mut aliased_dependencies = Vec::new();
```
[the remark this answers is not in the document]
Arm1stice: This logic has been reworked to avoid looping through a HashMap. Resolving.

impl/src/planning.rs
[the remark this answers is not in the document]
Arm1stice: Went ahead and created `sanitize_name`. For your last comment, I moved the logic to call the function when producing the Vec<String> in `produce_deps`

impl/src/planning.rs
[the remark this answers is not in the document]
Arm1stice: This has been renamed to `aliased` in the latest commit (I pushed about 30 seconds after your comments popped up) per https://github.com/google/cargo-raze/pull/162#discussion_r433554446. In the HashMap the key is the label of Bazel target, and the value is the name it is being aliased to in the `rust_library`/`rust_binary` rule.

impl/src/planning.rs
```diff
@@ -593,6 +618,9 @@ impl<'planner> CrateSubplanner<'planner> {
     let mut dev_dep_names = Vec::new();
     let mut normal_dep_names = Vec::new();
 
+    // Store renamed dependencies in a HashMap
+    let mut renamed_dep_names = HashMap::new();
```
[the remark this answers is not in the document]
Arm1stice: The variable is defined here because the usage is at the bottom of the `for` loop.

Arm1stice:
I made changes based on @dfreese's comments, and responded to @acmcarther's comments

impl/src/planning.rs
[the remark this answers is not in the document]
acmcarther: (My suggestion here is to move this declaration to L656)

impl/src/planning.rs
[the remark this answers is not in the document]
acmcarther: This needs to be pushed I think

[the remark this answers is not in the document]
Arm1stice: https://github.com/google/cargo-raze/pull/162/commits/993622e766ccad5f6693ef1e2ad2e414b80a7a11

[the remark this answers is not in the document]
acmcarther: Sorry, I think I was confused by the github UI. Thanks for clarifying, marking resulved.

impl/src/planning.rs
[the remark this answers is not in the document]
Arm1stice: [L656 in the commit this review is based on](https://github.com/google/cargo-raze/blob/5708a78840fd58389278de6c2695a9002a2bb9b8/impl/src/planning.rs#L656) is inside of the `for` loop. The logic looks weird, but that `map` (which has now been changed to `if let` syntax) was only to avoid unwrapping `dep.rename`. The HashMap needs to be declared above the `for` loop so that any renamed dependencies can be added to it. [The latest commit](https://github.com/google/cargo-raze/blob/993622e766ccad5f6693ef1e2ad2e414b80a7a11/impl/src/planning.rs#L612-L650) makes this logic much more readable.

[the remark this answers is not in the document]
acmcarther: My bad, you're definitely right. Another useless comment from me, hah.

Marking resolved.

impl/src/planning.rs
```diff
@@ -74,6 +76,8 @@ struct DependencySet {
   build_deps: Vec<BuildableDependency>,
   // Dependencies that are required for tests
   dev_deps: Vec<BuildableDependency>,
+  // Dependencies that have been renamed and need to be aliased in the build rule
```
acmcarther, line 79: Optional: This appears to contain strings of keys and values. This isn't obvious from the type Consider adding a note, like

```
  // Dependencies that have been renamed and need to be aliased in the build rule
  // formatted as `"$TARGET_NAME": "$BUILD_ALIAS"`
```

A slightly more type safe way to do this would be to use a simple struct containing these values, and do this formatting "later", but I don't feel strongly enough about this to really suggest it here.

impl/src/util.rs
```diff
@@ -163,6 +163,10 @@ pub fn sanitize_ident(ident: &str) -> String {
   slug::slugify(&ident).replace("-", "_")
 }
 
+pub fn sanitize_name(name: &str) -> String {
```
acmcarther, line 166: This is identical to sanitize_ident. Can that be used instead?

Arm1stice: This was made due to the suggestion at https://github.com/google/cargo-raze/pull/162#discussion_r433539678

impl/src/planning.rs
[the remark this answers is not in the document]
Arm1stice: No worries!

impl/src/planning.rs
Arm1stice: Just pushed a new commit pulling this out to a simple struct. Let me know what you think.

impl/src/util.rs
acmcarther: I guess lets hear if @dfreese intended to have a function with a different name but the same implementation. I think he expected that sanitize_name would just contain the "-" -> "_", but it ended up containing the "slugify" as well.

impl/src/templates/partials/rust_binary.template
```diff
@@ -36,4 +36,11 @@ rust_binary(
         "{{feature}}",
         {%- endfor %}
     ],
+    {%- if crate.aliased_dependencies | length != 0 %}
+    aliases = {
+        {%- for alias in crate.aliased_dependencies %}
+        {{alias}}
```
acmcarther, line 42: Does this need to be fixed to match rust_library.template?

impl/src/planning.rs
acmcarther: This looks good to me, thanks for making this change. Marking resolved.

impl/src/util.rs
dfreese: The intention was for it to just be `name.replace("-", "_")`

dfreese: Though, it was in part because I wasn't completely familiar with slugify.  I guess it could work to just use the `sanitize_ident` function in all cases.  @acmcarther do you have a preference on that?

impl/src/templates/partials/rust_binary.template
Arm1stice: Good catch, thank you

impl/src/util.rs
Arm1stice: Ah I see. For now, I have removed `sanitize_name` and changed it to call `sanitize_ident`. I can change it further if needed.

acmcarther:
I see the build in https://travis-ci.org/github/google/cargo-raze/builds/694036367 but it's not showing up here. Investigating this now.

I would second the request for a unit test as well.

Arm1stice:
I agree with writing unit tests. I'll try to get them done by EOW

Arm1stice:
Added a unit test based on the issue that this PR fixes. Let me know if it looks alright.

dfreese:
Sweet!  Looks great!  Minor nit: looks like `cargo fmt` needs be run in `impl`, otherwise good to me.

Arm1stice:
Okay, just pushed the new commit with `cargo fmt`

dfreese:
Thanks!  LGTM

Arm1stice:
Awesome! Thank you again for all of the comments.

PR #167, opened 2020-06-10
Add workspace-wide default for gen_buildrs

illicitonion:
I want to be able to optimistically default to generating build scripts by default.


acmcarther:
LGTM!

PR #170, opened 2020-06-12
Add build script dependencies on sys crates build scripts.

a1ph:
System crates build scripts may produce environment variables
that may be used by their direct dependencies build script.

For example openssl-sys generates DEP_OPENSSL_VERSION variable
that is then used by openssl crate build scripts.


acmcarther:
Thanks for your patience. This change LGTM, I'll include it in the next release after reviewing the other CLs.

PR #174, opened 2020-06-16
Fix: Make crate license identification more robust

Arm1stice:
This PR changes how cargo-raze identifies crate licenses. Currently, proper SPDX syntax is not properly recognized, so a crate with a license "MIT OR Apache-2.0" will be marked as restricted. While one solution would be to split the license string at "OR", I believe this implementation is much more robust, as it will support "AND" if that ever shows up in a crate license.

Using the SPDX crate, a crate's license string is parsed as an SPDX expression. The parser will verify that the licenses and syntax of the expression are valid. If a parse error is returned, the license is marked as restricted and a message is displayed next to the license in the build file. If multiple licenses are allowed ("OR"), the more permissive license is used. If multiple licenses are required ("AND"), the less permissive license is used.

Internal documentation states that the license field should only contain one parameter in the array, which should represent the least permissive license that the package is bound by. Therefore, the functionality has been changed to only show one license, rather than the previous functionality, which shows all types of licenses.

### DIFF
--- a/impl/src/license.rs
+++ b/impl/src/license.rs
@@ -63,19 +63,17 @@ pub struct BazelSpdxLicense {
 impl BazelSpdxLicense {
   fn combine_license_expr(license1: &Self, license2: &Self, operator: &str) -> String {
     // Surround license expressions with parenthesis if it isn't just the node name
-    let expr_str1: String;
-    if license1.name != license1.expression {
-      expr_str1 = format!("({})", license1.expression);
+    let expr_str1 = if license1.name != license1.expression {
+      format!("({})", license1.expression)
     } else {
-      expr_str1 = license1.expression.clone();
-    }
+      license1.expression.clone()
+    };
 
-    let expr_str2: String;
-    if license2.name != license2.expression {
-      expr_str2 = format!("({})", license2.expression);
+    let expr_str2 = if license2.name != license2.expression {
+      format!("({})", license2.expression)
     } else {
-      expr_str2 = license2.expression.clone();
-    }
+      license2.expression.clone()
+    };
 
     // Combine them using the operator
     format!("{} {} {}", expr_str1, operator, expr_str2)
--- a/impl/src/license.rs
+++ b/impl/src/license.rs
@@ -61,20 +61,20 @@ pub struct BazelSpdxLicense {
 }
 
 impl BazelSpdxLicense {
-  fn combine_license_expr(expr1: &Self, expr2: &Self, operator: &str) -> String {
+  fn combine_license_expr(license1: &Self, license2: &Self, operator: &str) -> String {
     // Surround license expressions with parenthesis if it isn't just the node name
     let expr_str1: String;
-    if expr1.name != expr1.expression {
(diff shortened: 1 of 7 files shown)


impl/src/license.rs
```diff
@@ -40,30 +45,143 @@ impl BazelLicenseType {
...
-pub fn get_available_licenses(cargo_license_str: &str) -> Vec<(String, BazelLicenseType)> {
-  let mut available_licenses = Vec::new();
-  for license_name in cargo_license_str.split('/') {
-    if license_name.is_empty() {
-      continue;
+/**
+ * A data structure for calculating a crate's license restrictions
+ */
+#[derive(Debug)]
+pub struct BazelSpdxLicense {
```
acmcarther, line 52: Probably doesn't need to be pub.

impl/src/license.rs
```diff
@@ -40,30 +45,143 @@ impl BazelLicenseType {
...
+  // The license of the name field. This represents the least restrictive license for the
+  // expression field.
+  pub license: BazelLicenseType,
+}
+
+impl BazelSpdxLicense {
+  fn combine_license_expr(license1: &Self, license2: &Self, operator: &str) -> String {
+    // Surround license expressions with parenthesis if it isn't just the node name
+    let expr_str1: String;
+    if license1.name != license1.expression {
```
acmcarther, line 67: consider assign-to-expression instead

```rust
let expr_str1 = if license1.name != license1.expression {
  format!("({})", license1.expression)
} else {
  license1.expression.clone()
}
```

impl/src/license.rs
```diff
@@ -40,30 +45,143 @@ impl BazelLicenseType {
...
+      expr_str1 = format!("({})", license1.expression);
+    } else {
+      expr_str1 = license1.expression.clone();
     }
 
-    // Trimming motivated by reem/rust-unreachable
-    let trimmed_license_name = license_name.trim();
+    let expr_str2: String;
```
acmcarther, line 73: Same here.

Arm1stice: Fixed both instances

impl/src/license.rs
Arm1stice: `get_available_licenses` currently returns this new type. Should I just make it return `LicenseData`?

acmcarther: Oh, nevermind. I think I was under the impression that this was redundant with some other "processing finished" type, but I think you're actually returning instances of this type from the license functino.

acmcarther: Actually yes, if that's possible then that is probably better.

I have some preference for having a "declarative", "simple" type for the values returned from the module, but I don't feel strongly about it, and I think "convenience" is a good enough argument for whatever you choose to do.

Arm1stice: Okay, newest commit has this change

acmcarther:
Oh I guess I need to merge this.

acmcarther:
This is in 0.3.5 now available on crates.io

PR #175, opened 2020-06-19
Add --output param specifying output folder for generated files.

boxdot:
Also fix:
* remote folder created on dry run
* remote/BUILD folder does not use prefix path parameter


boxdot:
@googlebot I signed it!

acmcarther:
LGTM

PR #176, opened 2020-06-22
Tag all cargo-raze rules with "cargo-raze" tag.

a1ph:
The tags can be used to exclude cargo-raze generated rules from, e.g.
"//...:all" targets. It is useful to prevent Bazel from building
unsupported targets in multiplatform configurations.


acmcarther:
Makes sense to me, LGTM.

PR #178, opened 2020-06-27
Root path fix for windows

jelmansouri:


impl/src/planning.rs
```diff
@@ -746,7 +746,9 @@ impl<'planner> CrateSubplanner<'planner> {
         .into_owned()
         // TODO(acmcarther): Is this even guaranteed to work? I don't think the `display` output
         // can be guaranteed....
-        .split_off(package_root_path.display().to_string().len() + 1);
+        .split_off(package_root_path.display().to_string().len() + 1)
+        // Handle Windows case where we end up here with backslashes
+        .replace("\\", "/");
```
dfreese, line 751: This behavior on lines 743-749/751 looks quite a bit like [strip_prefix](https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix).  Could that be used instead?  Would that help get rid of the special casing for windows?

jelmansouri: We would still endup with src\lib.rs for example instead of src/lib.rs.

acmcarther:
I don't have a strong preference on this one way or the other, but would like to wait for dfreese@'s LGTM since he provided an initial review.

impl/src/planning.rs
dfreese: Thanks for the clarification.  For my own reference, from [std::path::Path](https://doc.rust-lang.org/std/path/struct.Path.html):
```
separated by / on Unix and by either / or \ on Windows
```

It would be nice to have a test case for this, though I'm not sure if CI even runs on windows.  @acmcarther can you confirm.

acmcarther: Yeah, CI doesn't currently run on windows. Travis might have some support for running tests on windows though.

jelmansouri: I'll be happy to add CI for windows support.
Commit https://github.com/bazelbuild/rules_rust/commit/d2c4b145595fcb17f0a12521686a56b2c4126fd8 just adds it on the bazel rust rules. So all windows support is quite fresh, if we're building through bazel in the CI we need the latest rust rules.

jelmansouri: Can this be accepted, if no CI work is planned please.

UebelAndre:
@jelmansouri Hey, would you be able to rebase and fix the conflict here?

@dfreese @acmcarther And would you guys be able to do another review here? Perhaps @damienmg would be able to fill in for @acmcarther if he's not reachable.

UebelAndre:
Is this change still necessary?

PR #180, opened 2020-07-01
Support additional enviroment variable for build script

### LINKED ISSUE #179: buildrs_additional_environment_variables is no longer respected after #166
After #166, `buildrs_additional_environment_variables` is no longer respected. An example breakage is that `rust-protobuf` can no longer be built, as its [`build.rs`](https://github.com/stepancheg/rust-protobuf/blob/master/protobuf/build.rs) expected `RUSTC` and `CARGO_PKG_VERSION`. I temporarily unblocked myself by downgrading from 0.3.5 to 0.3.3, but I think the `buildrs_additional_environment_variables` should keep being supported, or it should be removed with its replacement documented.

damienmg:
Since the switch to the cargo_build_script rule, the
buildrs_additional_environment_variables was no longer working, this
change add this support back.

Fixes #179

### DIFF
--- a/impl/src/settings.rs
+++ b/impl/src/settings.rs
@@ -140,8 +140,6 @@ pub struct CrateSettings {
 
   /**
    * Additional environment variables to add when running the build script.
-   *
-   * Values should looks like "KEY=VALUE".
    */
   #[serde(default)]
   pub buildrs_additional_environment_variables: HashMap<String, String>,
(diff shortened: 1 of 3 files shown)


impl/src/settings.rs
```diff
@@ -144,7 +144,7 @@ pub struct CrateSettings {
    * Values should looks like "KEY=VALUE".
```
smklein, line 144: Is this comment out-of-date?

damienmg: Indeed, removed.

PR #184, opened 2020-07-08
Avoid silently dropping rules with conflicting names

smklein:
I ran into the following problem on a repository of mine which used Cargo Raze and Rules Rust:

- I had a dependency -- byteorder -- where I tried adding some compile-time flags in my Cargo.toml file.
- I noticed that, after re-running raze, I'd see a BUILD file for the target which included the flags I wanted...
- ... but when running Bazel, those flags were silently dropped!
- Eventually, this was traced back to an invocation of `rust_proto_repositories()` within my WORKSPACE file. Here's what happened: because my Cargo.toml file depended on an http_archive *also* checked into `rules_rust`, accessed by the `rust_proto_repositories` invocation, "which one got used" was order-dependent. The other was silently dropped.

I consider this a bug. This PR attempts to rectify this problem by changing the BUILD file generation of Raze such that multiple dependencies on the same target are no longer order-dependent. Instead, they explicitly flag an error.

If this change is considered undesirable, I'm happy to chatting about alternatives - but IMO the order-dependent alternative was pretty subtle.

### DIFF
--- a/impl/src/templates/remote_crates.bzl.template
+++ b/impl/src/templates/remote_crates.bzl.template
@@ -6,18 +6,10 @@ DO NOT EDIT! Replaced on runs of cargo-raze
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
 
-def _new_http_archive(name, **kwargs):
-    if not native.existing_rule(name):
-        http_archive(name=name, **kwargs)
-
-def _new_git_repository(name, **kwargs):
-    if not native.existing_rule(name):
-        new_git_repository(name=name, **kwargs)
-
 def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
 {% for crate in crates %}
     {%- if crate.source_details.git_data %}
-    _new_git_repository(
+    new_git_repository(
         name = "{{workspace.gen_workspace_prefix}}__{{crate.pkg_name | replace(from="-", to="_")}}__{{crate.pkg_version | slugify | replace(from="-", to="_")}}",
         remote = "{{crate.source_details.git_data.remote}}",
         commit = "{{crate.source_details.git_data.commit}}",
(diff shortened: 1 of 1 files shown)


impl/src/templates/remote_crates.bzl.template
```diff
@@ -6,18 +6,10 @@ DO NOT EDIT! Replaced on runs of cargo-raze
...
-
-def _new_git_repository(name, **kwargs):
-    if not native.existing_rule(name):
-        new_git_repository(name=name, **kwargs)
-
 def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
 {% for crate in crates %}
     {%- if crate.source_details.git_data %}
-    _new_git_repository(
+    new_git_repository(
```
dfreese, line 12:
```suggested
    git_repository(
```

impl/src/templates/remote_crates.bzl.template
```diff
@@ -6,18 +6,10 @@ DO NOT EDIT! Replaced on runs of cargo-raze
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
 
-def _new_http_archive(name, **kwargs):
-    if not native.existing_rule(name):
-        http_archive(name=name, **kwargs)
-
-def _new_git_repository(name, **kwargs):
-    if not native.existing_rule(name):
-        new_git_repository(name=name, **kwargs)
```
smklein, line 15: @dfreese  In the old version, this was the generated rule, so I was trying to stay consistent. If they're functionally equivalent, I'm happy to also add a migration to `git_repository`, but just a heads up.

dfreese: gotcha, thanks for the clarification.  Briefly glancing at the docs, I'm not sure I can tell the difference between a rule and it's `new_` equivalent.

damienmg:
Historically we were tending to not try to overwrite a repository that was declared earlier as this is the recursive behavior was planed to be the main workspace overload what ever sub-workspace do.

Do I understand correctly that you have several cargo.toml file (i.e. the one from `rust_proto_repositories` and your own) that conflict with each other? Due to the lack of good bazel support this was never intended to be used this way but rather to redeclare the proto dependencies in your own cargo.toml file so you get the correct cargo resolution.

Anyway, I don't think the original intent for those silent add is really significant, so this PR LGTM.

UebelAndre:
Is this still an issue?

dfreese:
closing for now.

PR #185, opened 2020-07-16
workspace_path can now act as a Bazel label within a Bazel workspace

UebelAndre:
This change makes it so that the `Cargo.toml` file can live within the [recommended location](https://doc.rust-lang.org/cargo/guide/project-layout.html) and `cargo-raze` will output content into a [Bazel workspace](https://docs.bazel.build/versions/master/build-ref.html#workspace) in a way that better matches the expectations set by [Bazel's naming conventions](https://docs.bazel.build/versions/master/build-ref.html)
```
├── BUILD
├── Cargo.lock
│
├── Cargo.toml  <- HERE
│
├── README.md
├── cargo
│   │
│   ├-   <- INSTEAD OF HERE
│   │
│   └── libloading_global_static.BUILD
└── src
    └── main.rs
```

This makes it easier to transition Rust projects to start using Bazel since users no longer have to move their `Cargo.toml` files and the use of `workspace_path` (which looks like a [Label](https://docs.bazel.build/versions/master/build-ref.html#labels)) makes more sense. Getting started with Bazel/cargo-raze should now only require users to add the `[raze]` section to their existing `Cargo.toml` file and run `cargo raze` (assuming they're using the remote genmode, but otherwise could vendor the packages and do the same thing). 

The smoke-tests show the up to date examples but I wasn't sure if I should update the [actual examples](https://github.com/google/cargo-raze/tree/master/examples) with the generated results of the smoke-tests. Anyway, this is my first ever PR in Rust and am looking forward to the feedback 😄


UebelAndre:
Hi, is anyone available to take a look at this?

UebelAndre:
@damienmg are you able to take a look at this or ping someone who can?

yesudeep:
I've been using symlinks for this and put the cargo directory in //third_party/cargo. 
This should help!

UebelAndre:
Yeah, I used symlinks to put my Cargo.toml file in the correct place such that rust-analyzer would work. But I feel this is a worth while change. I think cargo-raze shouldn't force users to move their Cargo.toml files. Especially since debugger support for rust through Bazel is kinda spotty. It's nice to be able to lean on the current rust tooling/ecosystem until the Bazel tooling matures enough to completely replace it.

impl/src/bazel.rs
```diff
@@ -23,6 +23,35 @@ use crate::{
...
+};
+
+/** Returns a path to a Bazel workspace root if one could be found, otherwise None */
+pub fn find_workspace_root(current_dir: Option<&Path>) -> Option<PathBuf> {
+  let default = env::current_dir().unwrap();
+  let cwd: &Path = current_dir.unwrap_or(default.as_path());
+
+  let workspace_files = [
+    cwd.clone().join("WORKSPACE.bazel"),
+    cwd.clone().join("WORKSPACE"),
```
dfreese, line 38: Is `clone()` needed here?  It looks [`Path::join`](https://doc.rust-lang.org/std/path/struct.Path.html#method.join) returns an owned PathBuf.

impl/src/bazel.rs
```diff
@@ -23,6 +23,35 @@ use crate::{
   util::RazeError,
 };
 
+use std::{
+  env,
+  path::{Path, PathBuf},
+};
+
+/** Returns a path to a Bazel workspace root if one could be found, otherwise None */
```
dfreese, line 31: You'd also want to document the behavior behind the assumption of the current_dir

impl/src/bazel.rs
```diff
@@ -23,6 +23,35 @@ use crate::{
...
+      return Some(cwd.to_path_buf());
+    }
+  }
+
+  match cwd.parent() {
+    None => {
+      return None;
+    }
+    Some(parent) => return find_workspace_root(Some(parent)),
+  };
```
dfreese, line 52: A match expression can be the returned from a function, so this could be:

```rust
match cwd.parent() {
  None => None,
  Some(parent) => find_workspace_root(Some(parent)),
}
```
which is just:
```rust
cwd.parent().and_then(|parent| find_workspace_root(Some(parent)))
```

That being said.  It doesn't look like you ever use `find_workspace_root(Some(x))` outside of this recursion.  Can you write this recursion in terms of a while loop?

That would allow you to eliminate the input parameter.

impl/src/bazel.rs
```diff
@@ -23,6 +23,35 @@ use crate::{
...
 };
 
+use std::{
+  env,
+  path::{Path, PathBuf},
+};
+
+/** Returns a path to a Bazel workspace root if one could be found, otherwise None */
+pub fn find_workspace_root(current_dir: Option<&Path>) -> Option<PathBuf> {
+  let default = env::current_dir().unwrap();
```
dfreese, line 33: In my comment on lines 47-52, I mentioned trying to eliminate the input parameter.

Perhaps, instead of eliminating the parameter, you could have the user of the function (`main`) pass this in.  That would make the behavior more clear, and also let you handle an `Err` from `current_dir()`, rather than calling unwrap.

impl/src/planning.rs
```diff
@@ -194,8 +195,16 @@ impl CrateCatalogEntry {
...
+  pub fn expected_vendored_path(&self, workspace_path: &String) -> String {
+    let mut dir = find_workspace_root(None).unwrap_or(PathBuf::from("."));
+    dir.push(workspace_path.trim_start_matches('/'));
+
+    format!(
+      "{}/{}{}",
+      dir.display().to_string(),
+      VENDOR_DIR,
+      &self.package_ident
+    )
```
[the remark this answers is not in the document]
dfreese: Update: Looking at the usage of `expected_vendored_path` it looks like it's just passed to `std::fs::metadata`, so it can probably just return std::path::Path, or std::path::PathBuf.

(There may be other invocations that I'm missing though)

impl/src/bazel.rs
dfreese: (Splitting out an `is_workspace_root(dir: &std::path::Path) -> bool` would probably make that a little more clear)

damienmg:
> @damienmg are you able to take a look at this or ping someone who can?

Sorry for the delay I tend to not look at Github outside of working hours (especially I am not even allowed to work on sundays in my country).

impl/src/bin/cargo-raze.rs
```diff
@@ -92,8 +92,27 @@ fn main() -> Result<()> {
...
+  // Allow the command line option to take precidence
+  if options.flag_output.is_empty() {
+    if let Some(workspace_root) = find_workspace_root(None) {
+      prefix_path.clear();
+      prefix_path.push(workspace_root);
+      prefix_path.push(
+        &planned_build
+          .workspace_context
+          .workspace_path
+          .trim_matches('/'),
```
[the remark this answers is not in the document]
UebelAndre: This is not intended manipulating a path, it's removing the leading `//` from the Bazel label. I agree with the use of `Pathbuf::join`

impl/src/planning.rs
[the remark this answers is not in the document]
UebelAndre: I've updated this to use `Pathbuf::join` but didn't change the return type.

impl/src/planning.rs
```diff
@@ -194,8 +195,16 @@ impl CrateCatalogEntry {
    *
    * Not for use except during planning as path is local to run location.
    */
-  pub fn expected_vendored_path(&self) -> String {
-    format!("./{}{}", VENDOR_DIR, &self.package_ident)
+  pub fn expected_vendored_path(&self, workspace_path: &String) -> String {
```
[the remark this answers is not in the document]
UebelAndre: Perhaps but I don't think it would make sense to be added to `CrateCatalogEntry`. Your thoughts there?

smoke_test/remote/complicated_cargo_library/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "compile_with_bazel"
 version = "0.1.0"
 
 [lib]
-path = "fake_lib.rs"
+path = "src/main.rs"
```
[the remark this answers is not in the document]
UebelAndre: There never was a `fake_lib.rs`. Cargo raze appears to not work if there is no `lib` or `bin` defined ([see README](https://github.com/google/cargo-raze/blob/master/README.md#generate-a-cargotoml)). I've updated these files to correctly define a binary.

impl/src/bazel.rs
UebelAndre: It's not, thanks, I removed the use of it here.

impl/src/bazel.rs
UebelAndre: Updated.

impl/src/bazel.rs
UebelAndre: I split out `is_workspace_root`, let me know you're thoughts there.

impl/src/bazel.rs
UebelAndre: I've updated `find_workspace_root` to handle errors and not panic anymore.

UebelAndre:
> @UebelAndre sorry for the delay, and thanks for the PR. I've gone through and reviewed the code based on the current implementation.
> 
> I know that some of my teammates have also symlinked the Cargo.toml and Cargo.lock of our project into the workspace root to be more compatible with RLS, so I'm definitely open to exploring this sort of change, however, would there be a way to do it in a way that wouldn't be a breaking change? One thought would be a flag to opt-in to this behavior.

I added an `incompatible_relative_workspace_path` (names are hard 😜 ) flag to allow for "opt-in" behavior. I feel strongly that this should at some point become the default behavior which is why I updated the smoke-tests and went with an `incompatible` flag.

yesudeep:
> I know that some of my teammates have also symlinked the Cargo.toml and Cargo.lock of our project into the workspace root to be more compatible with RLS

This breaks on Windows, which is why something that doesn't depend on symlinks may be a better idea.

UebelAndre:
@dfreese Hey, got time this week to do another review? 😅

UebelAndre:
Maybe @smklein or @acmcarther might have some time if @dfreese is not available? 😅

UebelAndre:
Hey @damienmg, don't mean to be a bother but do you know how I can get some review support for this PR? 😅

damienmg:
Hi @UebelAndre,

Unfortunately this is not a good time for me, but I'll try to take a look this week. Let see also if @mfarrugi have some bandwidth

damienmg:
(sorry for the misclick)

mfarrugi:
This looks fine to me, and addresses the earlier comments. I don't think I've reviewed anything else for this codebase, though.

damienmg (approved):
I also did a pass and see no issue with this code, I consider this one as good to go then seeing how many times it has been reviewed.

Thanks a lot for your patience!

PR #187, opened 2020-07-20
Allow for aliases at the top level

GregBowyer:
So in closing out #123 for #162 an important use case was missed, that of allowing top level aliases in the cargo that raze itself is going to load.

This is pretty important for large monorepos where doing a full migration of types upfront is pretty challenging.

Previously aliases could only exist inside targets that cargo-raze
depends on and not in the raze "root" crate itself.

This means that a construction such as the following

```toml
bytes = "2.0"
bytes_old = { version = "1.0", package = "bytes" }
```

Could not allow for the production of top level bazel aliases such as:

```python
"//cargo:bytes",
"//cargo:bytes_old",
```

This naturally turns up in larger codebases where migrartions are in
progress from one crate to another.

### DIFF
--- a/impl/Cargo.lock
+++ b/impl/Cargo.lock
@@ -77,7 +77,7 @@ checksum = "8504b63dd1249fd1745b7b4ef9b6f7b107ddeb3c95370043c7dbcc38653a2679"
 dependencies = [
  "gumdrop",
  "petgraph",
- "semver",
+ "semver 0.9.0",
  "serde",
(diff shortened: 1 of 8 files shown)


GregBowyer:
This depends on #127 to have the version expressed in the model as a version and not a string, its possible to detangle this but I think we should land #127 then this

GregBowyer:
This is dead in favor of #282

PR #193, opened 2020-08-05
[wip] Support Cargo Workspaces

samschlegel:
Very Draft PR (only a smoke-test so far) but adding here in hopes that maybe Travis will work 😄 

Also expect some random workflow changes that I'll pull out before making this ready for review.


UebelAndre:
@samschlegel This would be awesome! (or at least what I think this is) Do you think you could explain the new functionality so I know how excited to be? 😄

samschlegel:
So as is, this branch supports generation for a Cargo workspace in remote mode, the test is failing though for a different one so I might have broken something 😅 I also haven't tried vendor mode but that should be as easy as adding a test and working out the bugs.

There are some things I've had to hack around. This list probably isn't complete since I haven't touched this in two weeks so going off vague memories 
* I've only tested with the Cargo.toml in the root, since I'm trying to get this into a place where it can be used with a package in an existing workspace without breaking normal Cargo operations. This required changing some other things to support that, but it also made me realize it would be very nice to be able to split up where the remote/vendor dirs go, and where the Cargo.tomls live.
* Since BUILD files are generated in the same dirs as the Cargo.toml, and they just overwrite whatever's in there, this has issues being used on top of existing codebases. I think that's a problem that cargo-raze already has, but it's less common since it seems most people have a single dummy Cargo.toml in cargo-raze controlled path
* Having a workspace means we might have multiple version dependencies on the same library, but without using aliases. The alias generation needs to be able to handle these conflicts. Right now I have it just always appending the version which is a breaking change and probably not wanted on normal generations

UebelAndre:
Hey @samschlegel I was attempting to find a solution for this and ended up realizing what I really want is workspace support. I had a question I would love to get your thoughts on: 

What do you think should happen to transitive dependencies of workspace members that are also dependencies? This is related to #203. In the example there, would all the dependencies of `lib_a` explicitly be a part of `lib_b` during generation?

I'm also happy to help out with this pull request however I can 😄

TheButlah:
This would be awesome. Let me know if there is anything I can do to help

samschlegel:
Glad to see interest 😄 I might have some time to look over this this weekend, but we decided not to go with cargo-raze/rules_rust for now at work due to all the pain around build.rs and related things, and instead just wrote some rules to wrap `cargo` in a persistent bazel worker. Might try and OSS that worker soonish.

I think there's still two big questions left to answer that would be good to discuss.

* Where do we put BUILD files for sub-projects
  * For single projects these are put in the `workspace_path` I believe? I'd personally like them to end up closer to the actual project, but there are issues around how the BUILD files are currently written in that they just truncate whatever file exists there currently.
* How do we handle aliases when a workspace has multiple versions, since they will end up with the same name
  * This isn't a problem in single projects, because to have multiple version we have to explicitly alias, but that's not the case for workspaces

UebelAndre:
@samschlegel Awesome!

On a side note, I'd love to see wrapper rules for `cargo` you guys wrote. Do you think those will ever be published? 😅

UebelAndre:
>     * Where do we put BUILD files for sub-projects

Ok, I have a rough idea I'm just gonna throw out here. I think `workspace_path` should probably act as a relative path to the output directory based on each `Cargo.toml` file.
So given the following layout:
```
├── Cargo.toml
├── lib_a
│   └── Cargo.toml
├── lib_b
│   └── Cargo.toml
└── tools
    └── lib_c
        └── Cargo.toml
```

Running `cargo raze` anywhere in that directory structure would generate the following with `workspace_path = "cargo"`
```
├── Cargo.toml
├── cargo
├── lib_a
│   ├── Cargo.toml
│   └── cargo
├── lib_b
│   ├── Cargo.toml
│   └── cargo
└── tools
    └── lib_c
        ├── Cargo.toml
        └── cargo
```

The top level `cargo` directory would contain all the definitions for all repositories. The targets generated for each repository would now instead be numbered
```python
rust_library(
    name = "quote_0",
    crate_type = "lib",
    #...
)

rust_library(
    name = "quote_1",
    crate_type = "lib",
    #...
)

rust_library(
    name = "quote_2",
    crate_type = "lib",
    #...
)
```

This would allow for each target to supply the necessary features each dependency would need if they were to differ (obviously if there is no difference your targets would share the same dependency)

I think this also addresses your other point

>     * How do we handle aliases when a workspace has multiple versions, since they will end up with the same name

The `cargo` directories next to each `Cargo.toml` defining a package would have all the plumbing to alias the proper name of the dependency to the numbered versions.

```python
rust_library(
    name = "lib_b",
    deps = [
        "//lib_b/cargo:quote" # This maps to '@raze__quote__0_3_15//:quote_1` or something
    ],
    #...
)
```

Additionally, `cargo raze` should warn you if it's about to write over a file that doesn't have the appropriate header. I may make a PR for that...

I'll try to think about this more over the week but I would love to have your thoughts on this 😄

UebelAndre:
Still super interested in seeing this feature get added. I made #262 as a feature request issue to try and have a single point of entry for tracking any efforts or issues around supporting [cargo workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)

UebelAndre:
@samschlegel I believe #276 might address the efforts in this PR 😄 

Also, any update on those cargo rules?

UebelAndre:
With https://github.com/google/cargo-raze/issues/304 closed, workspace support should be officially done (at least rev1 of it 😅). This PR could be closed.

UebelAndre:
This has long since been supported, going to close per the comments above.

Issue #198, opened 2020-08-17
Add CI check to ensure `/examples` directory is kept up to date.


samschlegel:
When running the smoke test locally, I find it frequently has output changes to this directory. It also seems that the `WORKSPACE` file has been edited manually several times, so it doesn't match up with what the smoke test is actually testing against which seems risky. For example the current sha for rules_rust seems to be wrong for me, so the example doesn't build.

This should be a pretty easy thing to enforce in CI if people agree it's a good idea to do so.

PR #199, opened 2020-08-17
test: check that /examples has no changes

samschlegel:
Fixes https://github.com/google/cargo-raze/issues/198

This will fail until something like https://github.com/google/cargo-raze/pull/191 gets merged to update the examples


.travis.yml
```diff
@@ -40,3 +40,5 @@ script:
   - cargo install --debug -f --path .
   - cd ../
   - ./smoke-test.sh
+  # Check that /examples has no changes
+  - if [ -z "$(git status --porcelain)" ]; then echo "/examples is out of date. Please run `./smoke-test.sh` locally and commit the changes"; exit 1; fi
```
UebelAndre, line 44: Doesn't this need to be '! -z`?
```suggestion
  - if [ ! -z "$(git status --porcelain)" ]; then echo "/examples is out of date. Please run `./smoke-test.sh` locally and commit the changes"; exit 1; fi
```

UebelAndre:
@samschlegel  I apologize for this but because of #202, master will generate some diffs in the examples. Could I get you to add the following diff to this PR?
```diff
diff --git a/impl/src/templates/remote_crates.bzl.template b/impl/src/templates/remote_crates.bzl.template
index ed0d9e8..df88ed5 100644
--- a/impl/src/templates/remote_crates.bzl.template
+++ b/impl/src/templates/remote_crates.bzl.template
@@ -17,8 +17,8 @@ def _new_git_repository(name, **kwargs):

 def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
 {%- if crates %}
-    {% for crate in crates %}
-        {%- if crate.source_details.git_data %}
+{% for crate in crates %}
+    {%- if crate.source_details.git_data %}
     _new_git_repository(
         name = "{{workspace.gen_workspace_prefix}}__{{crate.pkg_name | replace(from="-", to="_")}}__{{crate.pkg_version | slugify | replace(from="-", to="_")}}",
         remote = "{{crate.source_details.git_data.remote}}",
@@ -27,7 +27,7 @@ def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
         init_submodules = True,
         {%- include "templates/partials/remote_crates_patch.template" %}
     )
-        {%- else %}
+    {%- else %}
     _new_http_archive(
         name = "{{workspace.gen_workspace_prefix}}__{{crate.pkg_name | replace(from="-", to="_")}}__{{crate.pkg_version | slugify | replace(from="-", to="_")}}",
         url = "https://crates-io.s3-us-west-1.amazonaws.com/crates/{{crate.pkg_name}}/{{crate.pkg_name}}-{{crate.pkg_version}}.crate",
@@ -39,8 +39,8 @@ def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
         {%- include "templates/partials/remote_crates_patch.template" %}
         build_file = Label("{{workspace.workspace_path}}/remote:{{crate.pkg_name}}-{{crate.pkg_version}}.{{workspace.output_buildfile_suffix}}"),
     )
-        {%- endif %}
-    {% endfor %}
+    {%- endif %}
+{% endfor %}
 {%- else %}
     pass
 {% endif %}
```

I can create my own PR for this if you'd rather not. I just felt like it was a small enough change that it'd be worth making here instead of committing the diffs in the examples directory (which is only a whitespace difference)

samschlegel:
I'm not really sure what happened with the test output here... it looks like it got mixed in with a bunch of output that I'm also not sure where came from since it happened after the smoke test exited 🤔

UebelAndre:
@samschlegel Maybe you need to rebase?

samschlegel:
I think it's because it was outputting to stdout and travis probably does something weird with stdout output at the end. Rebasing anyways, but don't think it's related

UebelAndre:
@samschlegel Yeah, sounds likely. I have very little experience with Travis-CI though so I have to do some more digging to really help out. Hopefully your change fixes it though! 😄

samschlegel:
Going to add a `rust-toolchain` file in this PR as well, since your changes in #185 use PathBuf capacity methods which weren't stabilized till 1.44

samschlegel:
> ~@samschlegel Maybe you need to rebase?~
> 
> edit: I misread, I see what you're saying now. But I still think your check will fail if you don't and do something about my comment above

Yeah my intention is to make sure this fails correctly, then fix things and make sure it works there too. I'd like to be able to test this...but who CI's the CI‽

samschlegel:
🤦 I'm using backticks in a double-quoted string lmaoooooo

samschlegel:
andddd, that rebase did something WEIRD

UebelAndre:
@samschlegel Looks good but I just wanted to get clarity on [this comment](https://github.com/google/cargo-raze/pull/199#issuecomment-680895380). Would you mind making that change as well? No worries if not, I'm just trying to cover up my shame 😅

UebelAndre:
@damienmg @dfreese Could someone with merge powers take a look at this pull-request?

damienmg (approved):
I just have a minor comment, there seems to be an unwanted edit in this PR.

impl/src/templates/remote_crates.bzl.template
```diff
@@ -17,8 +17,8 @@ def _new_git_repository(name, **kwargs):
 
 def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
 {%- if crates %}
-    {% for crate in crates %}
-        {%- if crate.source_details.git_data %}
+{% for crate in crates %}
```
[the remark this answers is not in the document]
UebelAndre: Why are they incorrect?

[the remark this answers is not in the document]
UebelAndre: If you look at 6ea6543b426a45437e14b052d0e10ff5c060a69f you can see that it correctly removes the whitespace. This is also valid jinja and obviously works in `cargo-raze` given the `crates.bzl` file you see in the commit I mentioned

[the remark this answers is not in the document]
damienmg: Ah I see, int that case let me add a suggestion.

impl/src/templates/remote_crates.bzl.template
```diff
@@ -17,8 +17,8 @@ def _new_git_repository(name, **kwargs):
...
+    {%- if crate.source_details.git_data %}
```
damienmg, line 21:
```suggestion
{%   for crate in crates %}
    {%-   if crate.source_details.git_data %}
```

samschlegel: the actual fix we probably want here is just to make the block for the for loop `{%-` instead of `{%` so that the whitespace is removed https://github.com/Keats/tera/blob/master/docs/content/docs/_index.md#whitespace-control

samschlegel: @UebelAndre for now I'm just going to revert the changes related to this since I don't want to block this PR on that. Could you open up a separate PR with these changes?

UebelAndre: Yeah, I can do that. Thanks for the attempt 😄

UebelAndre: I created #207

damienmg:
Thanks!

PR #200, opened 2020-08-19
Emit crate_root with forward slashes on Windows

rickwebiii:
Windows was emitting bazel paths with back slashes for `crate_entry` attrs because we call `path.to_string_lossy`. Fix that.

### DIFF
--- a/impl/src/planning.rs
+++ b/impl/src/planning.rs
@@ -14,6 +14,7 @@
 
 use std::{
   collections::{HashMap, HashSet},
+  path,
   path::PathBuf,
   str::{self, FromStr},
 };
@@ -768,6 +769,11 @@ impl<'planner> CrateSubplanner<'planner> {
       if package_root_path_str.starts_with("./") {
         package_root_path_str = package_root_path_str.split_off(2);
       }
+
+      // Bazel wants forward slashes, but some operating systems don't use '/' as the path
+      // separator. For example, MacOS Classic uses ':', and Windows uses '\'. While Bazel
+      // doesn't and will never support the former, it does support the latter.
+      let package_root_path_str = package_root_path_str.replace(path::MAIN_SEPARATOR, "/");
 
       for kind in &target.kind {
         targets.push(BuildableTarget {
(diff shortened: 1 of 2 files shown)


impl/src/planning.rs
```diff
@@ -769,6 +770,11 @@ impl<'planner> CrateSubplanner<'planner> {
         package_root_path_str = package_root_path_str.split_off(2);
       }
 
+      // Bazel wants forward slashes, but some operating systems don't use '/' as the path
```
rickwebiii, line 773: Maybe a better fix for this would be to change line 750 to not turn the path into a string, call 
```
let package_root_path = target.src_path
  .strip_prefix(&package_root_path())
  .components()
  .map(|c| { c.as_os_str().to_string().expect(format!("Fatal: {:?} has non-UTF-8 characters in it.", c.as_os_str()) })
  .collect::<Vec<String>>()
  .join("/")
```

This
1) Obviates the TODO
2) Obviates 769-771, as components skips . and path separators
3) Calling to_string_lossy on Paths (which are internally sequences of things built on OsString) because you want to write them in a config file so they can be reliably read by another tool is not generally correct. While unconventional, it's legal in Linux to have non-UTF-8 filenames. If we encounter one, it's better to fail fast in this tool rather than letting than rustc failing with file not found on a path name in the config file that might be visually indistinguishable from the actual path on disk. Not that Bazel can even do well with all legal paths containing only UTF-8 either... Spaces anyone?

rickwebiii:
@googlebot I signed it!

impl/src/planning.rs
dfreese: I generally like the approach.  Are you able to to return an error from the function instead of panicking?

rickwebiii: We'd have to change the Result type of this function, as the Error is no longer io::Error, but it's the standard "crate-level Error with a bunch of Variants for all the errors this crate can encounter with a bunch of impl From trait pattern."

I'll have to do something more clever than collect now that I have Vec<Result<String, Error>> rather than Vec<Strings> coming out of the map statement.

Issue #189, opened 2020-07-23
Create BUILD.bazel files instead of BUILD files for packages (possibly make it optional)


yesudeep:
Running cargo-raze currently generates `BUILD` files in the cargo and remote directories.
However, because the name `BUILD` can conflict with a directory named 'build' on case-insensitive
file systems, it's a much better idea to use the .bazel extension instead.

Related issue:
`cargo-raze` generates `BUILD` files in directories that already have a `BUILD.bazel` file as well.

Could you please make it optional to generate `BUILD.bazel` files instead of `BUILD` files perhaps via
a command line flag to set the extension?

mstg:
Hello!
This is already supported (#93) 

Add `output_buildfile_suffix = "BUILD.bazel"` under `[raze]` in `Cargo.toml`

yesudeep:
Thank you. That works!

This bug still exists however:

"cargo-raze generates BUILD files in directories that already have a BUILD.bazel file."

UebelAndre:
Hey @yesudeep, 

I opened #205 to update the default buildfile suffix.

Additionally, If you build using `incompatible_relative_workspace_path = True` and run `cargo-raze` within a Bazel workspace, raze will treat `workspace_path` as a Bazel label and put the generated files in the specified directory. This should avoid any conflicts you might have seen before.

UebelAndre:
@yesudeep does #205 satisfy this issue?

UebelAndre:
If it does, could you close the ticket?

yesudeep:
Hey @UebelAndre thank you! :-)

PR #205, opened 2020-08-24
Updated `output_buildfile_suffix` default to `BUILD.bazel`

UebelAndre:
This is a followup to #189

I've run into many situations where the `BUILD` file has conflicted other things on case insensitive filesystems. This should help avoid these situations.


dfreese:
Based on this thread (https://github.com/bazelbuild/bazel/issues/4517), It seems like this is preferred.

As it does cause some churn on the output, as you noticed with the examples, I'd make sure there is some consensus on this. @damienmg,  @acmcarther, or @smklein any objections?

damienmg:
Yes BUILD.bazel should be the default for automatically generated file, its works better with IDEs.

Issue #172, opened 2020-06-13
Examples do not compile at tag `v0.3.3`


filmil:
```
┬─[fmil@fmil2:~/tmp/cargo-raze/examples]─[11:40:15 PM]
│ (g/b:(HEAD detached at v0.3.3))
╰─>$ bazel build //...
DEBUG: Rule 'non_cratesio__env_logger__0_5_5' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1520561478 +1000"
DEBUG: Repository non_cratesio__env_logger__0_5_5 instantiated at:
  no stack (--record_rule_instantiation_callstack not enabled)
Repository rule new_git_repository defined at:
  /home/fmil/.cache/bazel/_bazel_fmil/f699d91921e3fa7be0a430d6437d5e96/external/bazel_tools/tools/build_defs/repo/git.bzl:182:37: in <toplevel>
DEBUG: Rule 'non_cratesio__log__0_4_0' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1514152054 -0700"
DEBUG: Repository non_cratesio__log__0_4_0 instantiated at:
  no stack (--record_rule_instantiation_callstack not enabled)
Repository rule new_git_repository defined at:
  /home/fmil/.cache/bazel/_bazel_fmil/f699d91921e3fa7be0a430d6437d5e96/external/bazel_tools/tools/build_defs/repo/git.bzl:182:37: in <toplevel>
DEBUG: Rule 'non_cratesio__rand__0_4_1' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1513596352 +0000"
DEBUG: Repository non_cratesio__rand__0_4_1 instantiated at:
  no stack (--record_rule_instantiation_callstack not enabled)
Repository rule new_git_repository defined at:
  /home/fmil/.cache/bazel/_bazel_fmil/f699d91921e3fa7be0a430d6437d5e96/external/bazel_tools/tools/build_defs/repo/git.bzl:182:37: in <toplevel>
INFO: Analyzed 120 targets (0 packages loaded, 0 targets configured).
INFO: Found 120 targets...
[...]

Issue #171, opened 2020-06-13
Examples do not compile at d0ed12d96c595de11ccb1c4ac6c5b1627a9504d3


filmil:
Not sure if this is intended, but I suppose it's desirable to have the head revision compile...

```
─[fmil@fmil2:~/tmp/cargo-raze/examples]─[11:36:36 PM]
│ (g/b:master)
╰─>$ bazel build //...
INFO: Repository io_bazel_rules_rust instantiated at:
  no stack (--record_rule_instantiation_callstack not enabled)
Repository rule http_archive defined at:
  /home/fmil/.cache/bazel/_bazel_fmil/f699d91921e3fa7be0a430d6437d5e96/external/bazel_tools/tools/build_defs/repo/http.bzl:336:31: in <toplevel>
WARNING: Download from https://github.com/bazelbuild/rules_rust/archive/6e5fa2c570ac2f17ac1df840d060fc8aab521a07.tar.gz failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException Checksum was 2e690b7d0caccc3000b98a9831adf2899b36268efec3ced8d3cfaec6322843d1 but wanted a62597502c1ee14fcc30da92fc31c69bc6b436f464939449dc7be88c76218a31
ERROR: An error occurred during the fetch of repository 'io_bazel_rules_rust':
   java.io.IOException: Error downloading [https://github.com/bazelbuild/rules_rust/archive/6e5fa2c570ac2f17ac1df840d060fc8aab521a07.tar.gz] to /home/fmil/.cache/bazel/_bazel_fmil/f699d91921e3fa7be0a430d6437d5e96/external/io_bazel_rules_rust/6e5fa2c570ac2f17ac1df840d060fc8aab521a07.tar.gz: Checksum was 2e690b7d0caccc3000b98a9831adf2899b36268efec3ced8d3cfaec6322843d1 but wanted a62597502c1ee14fcc30da92fc31c69bc6b436f464939449dc7be88c76218a31
[...]

PR #209, opened 2020-08-28
Standard output buildfiles

UebelAndre:
This is the final change I feel should be made to the general structure of the output Bazel files.

Looking at [new_local_repository](https://docs.bazel.build/versions/master/be/workspace.html#new_local_repository) and other Bazel workspace rules, the examples they have for it's use is as follows.
```python
new_local_repository(
    name = "my-ssl",
    path = "/home/user/ssl",
    build_file = "BUILD.my-ssl",
)
```
Note the line `build_file = "BUILD.my-ssl"`

Because of this I think `cargo-raze` should generate files with a similar structure in `remote` mode. I think this improves the readability/tracibility of the output files. Additionally, I feel the files should be suffixed with `.sky` to improve the IDE experience.


UebelAndre:
@dfreese @damienmg If either one of you have time to take a look. Feedback would be greatly appreciated ❤️

damienmg:
Hi @UebelAndre,

Thanks for the PR, AFAICT this was never standardised as one way or another. As it also change the extension of the build file to .bzl which design the full bazel's starlark dialect, which a build file is not (build dialect).

I am fine with changing the extension if build is too much a trouble but that shouldn't be .sky IMHO.

UebelAndre:
@damienmg Ok, I've updated it to `.bzl` 

My thoughts here in first going with `.sky` was because these files are not intended to be "loaded" 

[Bazel - Load](https://docs.bazel.build/versions/master/build-ref.html#load)

> Bazel extensions are files ending in .bzl. Use the load statement to import a symbol from an extension. 

I only wanted to be able to identify the files as Starlark files (which `.sky` would do). Admittedly though, I didn't see anything that explicitly stated `.sky` was the desired extension for Starlark, however, all examples in [the spec](https://github.com/bazelbuild/starlark/blob/master/spec.md) refer to modules with the `.sky` extension.

damienmg:
I think I got my messaging wrong. I think neither `.bzl` nor `.sky` is correct. `.bzl` should be loaded, `.sky` is the generic extension used by the spec. `.bazel` is the one that is found on standard build file, so we should probably settle on this one if we want to always adds an extension (in that case, maybe instead simply change the default name to be `BUILD.bazel`?)

UebelAndre:
@damienmg Ah! I see now, yeah, that makes way more sense. I don't know why I didn't consider that before 😅 

I made a couple additional changes in this PR.

The first is to add the `manual` tag to all generated targets to make them not buildable by default. This "fixes" #171 and #172 (by not having `//...` include the problematic target that's generated by `raze`) and should generally improve the experience around `bazel build //...`. 

Second, I separated the targets generated `cargo raze` in smoke-test to be prefixed by either `remote` or `vendored`. This will prevent any mixing of dependencies (and I think is a much cleaner implementation for projects that run `cargo raze` for multiple parts of their repo. I think it'll lead to less confusing workspace issues until something like #193 lands.

UebelAndre:
@damienmg I still think #172 is an issue worth looking into though. The `redox_syscall` dependency seems to depend on the `nightly` toolchain. I haven't looked deeper into where `redox_syscall` is coming from though, I didn't see it explicitly set in any of the top level dependencies listed in the `Cargo.toml` files for each example. So likely some transitive dependency hidden behind some feature.

impl/src/planning.rs
```diff
@@ -212,8 +212,8 @@ impl CrateCatalogEntry {
   pub fn local_build_path(&self, settings: &RazeSettings) -> String {
     match settings.genmode {
       GenMode::Remote => format!(
-        "remote/{}.{}",
-        &self.package_ident, settings.output_buildfile_suffix,
+        "remote/BUILD.{}.bazel",
```
damienmg, line 215: Why not `{}.BUILD.bazel`? And then why simply not update the default configuration for output_buildfile_suffix from `BUILD` to `BUILD.bazel`?

damienmg (commented):
Sorry for coming back at the naming, except from it the whole PR SGTM.

impl/src/planning.rs
UebelAndre: I like the idea that all build files start with "BUILD" I think it improves the discoverability of them. Also, wasn't "output_buildfile_suffix" updated in #205?

damienmg: I have missed #205 
I feel othewise, the reading `openssl.BUILD.bazel` feels more natural than `BUILD.opensssl.bazel`, in addition finding all the build file would be `*BUILD.bazel`.

Let try to bring in more opinion from the rules_rust folks: @acmcarther @dfreese @smklein @mfarrugi

UebelAndre: Becasue bazel didn't enforce the `.bazel` suffix I think the thing users end up searching would be `BUILD*` (that's what I do for `find`, `grep`, and generally searching for the correct `BUILD` file in Vscode. I can change it if you feel strongly about it but I'd prefer the current implementation. Let me know and/or I'll wait for a response from one of the people you pinged.

damienmg: Since we have to different point of view on this and it is in the end of question of point of view, let's wait and gather more opinion, and see which way to go.

UebelAndre: @acmcarther @dfreese @smklein @mfarrugi anyone available to chime in?

mfarrugi:
I think of BUILD and BUILD.bazel as the file extension, so would prefix the
name.

UebelAndre:
Alright, I've updated the pull request to stick with the `output_buildfile_suffix` suffix (`{{ crate_name }}`.`{{output_buildfile_suffix }}`/`lib.BUILD.bazel`)

UebelAndre:
@damienmg Hey, this is ready for you whenever you get a chance 😄

damienmg:
LGTM!

PR #213, opened 2020-09-09
Formalized 'load_settings' by moving it into the settings module

UebelAndre:
This change moves the functionality of `load_settings` into the `settings` module so that it can be properly tested.


damienmg:
LGTM, Thanks!

PR #214, opened 2020-09-09
Optimized smoke test script

UebelAndre:
The `./smoke-test.sh` script has been updated to build `cargo-raze` before nuking the examples directory so should a build error occur, you no longer have a huge diff to look at.


damienmg (approved):
LGTM, Thanks!

Issue #210, opened 2020-09-02
dump-metadata is broken


UebelAndre:
When running `dump-metadata.sh` (or just directly compiling the binary) the following error occurs.
```output
error[E0432]: unresolved import `cargo_raze::metadata::CargoInternalsMetadataFetcher`
  --> src/bin/dump_metadata.rs:37:5
   |
37 | use cargo_raze::metadata::CargoInternalsMetadataFetcher;
   |     ^^^^^^^^^^^^^^^^^^^^^^-----------------------------
   |     |                     |
   |     |                     help: a similar name exists in the module: `CargoMetadataFetcher`
   |     no `CargoInternalsMetadataFetcher` in `metadata`

warning: unused import: `cargo_raze::metadata::MetadataFetcher`
  --> src/bin/dump_metadata.rs:39:5
   |
39 | use cargo_raze::metadata::MetadataFetcher;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

error: aborting due to previous error; 1 warning emitted

For more information about this error, try `rustc --explain E0432`.
error: could not compile `tools`.
```

What is the status of this tool and should it be fixed/added to CI?

UebelAndre:
@dfreese would you know the answer to this?

UebelAndre:
@damienmg do you know anything about this?

damienmg:
No idea :/ @acmcarther is the original creator of this create and might knows.

acmcarther:
This was added in https://github.com/google/cargo-raze/pull/84 and  https://github.com/google/cargo-raze/pull/86 to help understand why ripgrep wasn't buildable in https://github.com/google/cargo-raze/issues/83

It's probably not complex to update the tool, but I'm probably the only one that cares enough to fix it. I'll set aside some time to do that.

UebelAndre:
@acmcarther Still planning on updating the tool?

acmcarther:
I guess so.

The proximate issue here is that I don't allocate time to work on this. The root issue is that I'm not working on rust-in-bazel projects right now (instead, i'm doing k8s stuff....), so I don't have a strong motivation to rework my side project schedule.

[...]

PR #217, opened 2020-09-15
Add publish-new-version.sh for pushing versions of the crate

acmcarther:
Adds a script for publishing new versions of this crate to codify my shell history into a shell script usable by other folks.

Has some rudimentary checks for the existence of `cargo` and that the working directory is clean.

r: @UebelAndre as well


UebelAndre:
@acmcarther Are there also some examples of using this script? Whenever I add things like this to projects at work, I typically add a block of comments at the top of the file that show simple usage just to make it **crystal clear** as to what to do.
```bash
# Use the script like so
# Args: ...docs...
#
# ./my_script.sh arg1 arg2
```
```bash
# Use the script like so
# Environment: ...docs...
#
# ARG_VAR=1 ./my_other_script.sh
```
Things like that. 

While It's unlikely I'll be doing a deploy for this so If there's some internal Google doc that has this, that works as well. The thing I want to avoid is all maintainers who have familiarity with the project/script being unavailable and therefore no more releases happen because maybe some constraint changed or things are not as obvious to the available/new maintainers. 

Whether or not you decide to do that, I'm thankful for this PR! 😄

acmcarther:
Thanks, that's valuable feedback.

I've added some documentation about what the script does and how to use it to the top of the file. Incidentally, I've also updated the mechanism it uses to take in the new version from an environment variable to a regular argument.

publish-new-version.sh
```diff
@@ -0,0 +1,87 @@
...
+    command -v "$1" >/dev/null 2>&1 || ( echo "Command \`$1\` isn't available. Please install before continuing."; exit 1 )
+}
+
+update_cargo_toml_version() {
+  crate_version=$1
+  sed -i "s/^version =.*/version = \"$crate_version\"/g" Cargo.toml
+}
+
+cargo_build_crate() {
+  cargo build
```
UebelAndre, line 28: Should this have `--release` with it?

publish-new-version.sh
```diff
@@ -0,0 +1,87 @@
...
+
+if ! [ -z "$(git status --porcelain)" ]; then
+  echo "Working directory is not clean. Commit pending changes before running command."
+  exit 1
+fi
+
+PWD="$(pwd)"
+
+REPO_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$REPO_ROOT/impl"
```
UebelAndre, line 83: How would you feel about `set -x` here? Might make logging of what the deploy process is doing nicer in the event of something going wrong.

damienmg (approved):

UebelAndre (approved):

publish-new-version.sh
acmcarther: Now done!

publish-new-version.sh
acmcarther: Done.

Added explanation for why we're doing this as well.

UebelAndre (approved):

UebelAndre:
@acmcarther When the version is updated should the [`Cargo.lock`](https://github.com/google/cargo-raze/blob/master/tools/Cargo.lock) file be updated too in the `/tools` directory?

acmcarther:
Re https://github.com/google/cargo-raze/pull/217#issuecomment-694394354
 
I'm not really sure. I don't consider the contents of `tools/` to be part of the package proper though, so i think it's OK to punt on this question for now.

That said, this should be a consideration in handling https://github.com/google/cargo-raze/issues/210 I think.

UebelAndre:
As a quick followup. Do you think you could delete some old branches? 😅

acmcarther:
Done!

PR #219, opened 2020-09-15
docs: update instructions for lockfile generation

sudoforge:
`cargo vendor` currently outputs helptext regarding updating
`.cargo/config.toml` in order to replace crates.io sources with the
vendored sources, e.g.:

```
➜ cargo vendor --versioned-dirs
To use vendored sources, add this to your .cargo/config for this
project:

[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "path/to/vendor"
```

If a user adds this to their `.cargo/config.toml` configuration prior to
the first run of `cargo vendor`, then `cargo generate-lockfile` and
`cargo vendor --locked` will fail to update the lockfile.

This patch removes the instructions related to generating the lockfile
with `cargo generate-lockfile` (and the subsequent `--locked` flag to
`cargo vendor`), instead preferring to let `cargo vendor` manage the
lockfile. The updated series of commands works in all cases, including
first runs which already have a `.cargo/config.toml` file replacing
`crates-io` sources. It's also now one command and one flag shorter,
which some people might like too, I guess.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,16 +120,11 @@ targets.
 $ cargo install cargo-raze
 ```
 
-Then, generate a lock file for your dependencies
+Following that, vendor your dependencies from within the cargo/ directory. This
+will also update your `Cargo.lock` file.
 
 ```bash
-$ cargo generate-lockfile
-```
-
-Following that, vendor your dependencies from within the cargo/ directory.
-
-```bash
-$ cargo vendor --versioned-dirs --locked
+$ cargo vendor --versioned-dirs
 ```
 
 Finally, generate your BUILD files, again from within the cargo/ directory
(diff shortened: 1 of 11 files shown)


acmcarther:
Thanks for this PR and sorry for the egregious delay.

My impression is that users should ignore any instructions from `cargo vendor` about changing their Cargo.tomls, but I'm not really sure. Maybe I'm mistaken. In any case, the instructions fail to communicate that if that's what I intended.

Generally the changes here LGTM, though it seems there's a conflict in some readmes.

sudoforge:
> My impression is that users should ignore any instructions from cargo vendor about changing their Cargo.toml

This really isn't about `Cargo.toml`, but rather the local cargo configuration file. This is used by various tools (e.g. editor plugins). The stanza in my original comment is actually necessary for cargo in order to resolve vendored dependencies correctly, which helps make sure language analysis tools aren't throwing errors related to missing depdendencies.

I've rebased this tree to resolve the conflicts.

acmcarther:
Makes sense to me, thank you very much for fixing this!

sudoforge:
@acmcarther any chances there's work progressing on open sourcing the editor integrations used internally?

PR #221, opened 2020-09-16
Don't copy Cargo.toml to tempdir where possible

illicitonion:
I'm not entirely clear on why we're copying out in the first place, but
I added a comment with my best guess.

Avoid doing this where possible, because it breaks support for Cargo
workspaces.

Alternatively, we could try to parse the Cargo.toml for its workspace
members and copy those too, but that seems to be breaking more
abstraction layers.

The limited cases of Workspaces working is probably also worth documenting, though I'm not sure where?

### DIFF
--- a/impl/src/metadata.rs
+++ b/impl/src/metadata.rs
@@ -63,9 +63,25 @@ impl MetadataFetcher for CargoMetadataFetcher {
     assert!(files.toml_path.is_file());
     assert!(files.lock_path_opt.as_ref().map_or(true, |p| p.is_file()));
 
-    // Copy files into a temp directory
     // UNWRAP: Guarded by function assertion
-    let cargo_tempdir = {
+    let toml_dir = files.toml_path.parent().unwrap().to_path_buf();
+
+    // If the Cargo.lock file is specified and is in a different directory to the Cargo.toml,
+    // we need to copy them to a tempdir so that they're adjacent and cargo picks them up.
+    // Note that this breaks Workspace support, as the child Cargo.toml files will not be copied
+    // over into the right place.
+    let can_operate_inplace = if let Some(lock_path) = files.lock_path_opt.as_ref() {
+      lock_path.parent() == files.toml_path.parent()
+    } else {
+      !toml_dir.join("Cargo.lock").exists()
+    };
+
+    // Keep _tempdir alive until the end of the function so that the dir doesn't get deleted until
+    // the cargo-metadata process exits.
+    let (working_directory, _tempdir) = if can_operate_inplace {
+      (toml_dir, None)
+    } else {
+      // Copy files into a temp directory
       let dir = TempDir::new()?;
 
       let dir_path = dir.path();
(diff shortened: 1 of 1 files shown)


UebelAndre:
I've wondered the same thing. I wondered if running `cargo metadata` littered your repo with some artifacts in some context. 

Loosely related, you may want to check out #193 if you're trying to do things with workspaces (just to throw in your 2 cents) 😄

UebelAndre:
I feel like there isn't much worth doing outside of actually solving for workspace support here. Do you have any thoughts on what that would look like? I don't think it's a huge change but I could be wrong 😅

impl/src/metadata.rs
```diff
@@ -63,9 +63,25 @@ impl MetadataFetcher for CargoMetadataFetcher {
...
     assert!(files.lock_path_opt.as_ref().map_or(true, |p| p.is_file()));
 
-    // Copy files into a temp directory
     // UNWRAP: Guarded by function assertion
-    let cargo_tempdir = {
+    let toml_dir = files.toml_path.parent().unwrap().to_path_buf();
+
+    // If the Cargo.lock file is specified and is in a different directory to the Cargo.toml,
+    // we need to copy them to a tempdir so that they're adjacent and cargo picks them up.
+    // Note that this breaks Workspace support, as the child Cargo.toml files will not be copied
```
acmcarther, line 71: What does "it breaks workspace support" mean?

acmcarther (commented):
The code generally looks fine to me, though we probably should hold off for https://github.com/google/cargo-raze/pull/227 or one of its derived PRs will make this obsolete. Not sure though.

IIRC, the copying behavior existed so that raze woudn't modify files in the workspace. I don't know if that's valuable behavior in practice.

LMK if there's a consensus to merge this first.

UebelAndre:
> IIRC, the copying behavior existed so that raze woudn't modify files in the workspace. I don't know if that's valuable behavior in practice.

Do you remember what files might have gotten modified?

I think it'd be great to remove the the logic for copying files into tempdirs if possible since I think that'd simplify the implementation for workspace support.

UebelAndre:
Oh, and yeah, if this could hold off until #227 goes through that'd be helpful. I'll be as responsive as I can to make sure it gets in quickly 😄

acmcarther:
No I've long since forgotten what it actually did.. though the git history might be illuminating. I can take a look at this in a bit.

UebelAndre:
@acmcarther https://github.com/google/cargo-raze/pull/28 appears to be where this was introduced. Does that bring back any memories? There's nothing specifically calling out the `tempdir` stuff in that PR.

acmcarther:
That PR is the first PR where we add an alternative to using cargo's internals to do metadata fetching.

It doesn't really clear anything up beyond what I already remembered, which is that I wanted to run the subcommand away from the actual working directory to avoid any "side effects" like generated files. This was probably more precautionary than deliberate and was probably inspired by my understanding of how bazel itself works.

illicitonion:
https://github.com/google/cargo-raze/pull/227 added `use_tempdir` to `CargoMetadataFetcher` which fills my need here :)

Issue #218, opened 2020-09-15
Cargo Raze ignores binary dependencies


UebelAndre:
I was doing some tests trying to refresh things in `rules_rust` and came across an issue with generating targets for [`@io_bazel_rules_rust//wasm_bindgen/raze/...`](https://github.com/bazelbuild/rules_rust/tree/master/wasm_bindgen/raze). 

There appears to have been a regression introduced in 816ea69b07181337ffe0e035b8e2c6a78bf631f2 (merged by #134) where a lot of the targets are no longer being generated. I've yet to nail down the specific cause. But, if you run `cargo raze` on or after this commit, you get the following output.
```output
Loaded override settings: RazeSettings {
    workspace_path: "//wasm_bindgen/raze",
    target: "x86_64-unknown-linux-gnu",
    crates: {
        "syn": {
            "0.15.44": CrateSettings {
                additional_deps: [],
                skipped_deps: [],
                extra_aliased_targets: [],
                additional_flags: [
                    "--cfg=syn_can_call_macro_by_path",
                ],
                gen_buildrs: true,
                data_attr: None,
                buildrs_additional_environment_variables: [],
                patch_args: [],
                patch_cmds: [],
                patch_cmds_win: [],
                patch_tool: None,
                patches: [],
            },
        },
        "proc-macro2": {
            "0.4.30": CrateSettings {
                additional_deps: [],
                skipped_deps: [],
                extra_aliased_targets: [],
                additional_flags: [
                    "--cfg=use_proc_macro",
                ],
                gen_buildrs: false,
                data_attr: None,
                buildrs_additional_environment_variables: [],
                patch_args: [],
                patch_cmds: [],
                patch_cmds_win: [],
                patch_tool: None,
                patches: [],
            },
        },
        "wasm-bindgen": {
            "0.2.48": CrateSettings {
                additional_deps: [],
                skipped_deps: [],
                extra_aliased_targets: [
                    "cargo_bin_wasm_bindgen",
                ],
                additional_flags: [],
                gen_buildrs: true,
                data_attr: None,
                buildrs_additional_environment_variables: [],
                patch_args: [],
                patch_cmds: [],
                patch_cmds_win: [],
                patch_tool: None,
                patches: [],
            },
        },
    },
    gen_workspace_prefix: "raze",
    genmode: Remote,
    output_buildfile_suffix: "BUILD",
}
Generated remote/BUILD successfully
Generated .//remote/bumpalo-2.6.0.BUILD successfully
[...]

PR #222, opened 2020-09-16
Try to fill in sha256es from local index mirror

illicitonion:
If anything goes wrong, we simply fall back to the current behaviour of
omitting them.

### DIFF
--- a/impl/src/planning/subplanners.rs
+++ b/impl/src/planning/subplanners.rs
@@ -246,6 +246,28 @@ impl<'planner> WorkspaceSubplanner<'planner> {
       .collect();
     files.push(self.files);
 
+    let package_to_checksum = self.populate_sha256_index(&files)?;
+
+    let contexts = catalogs
+      .iter()
+      .map(|catalog| {
+        (*catalog)
+          .metadata
+          .resolve
+          .as_ref()
+          .ok_or_else(|| RazeError::Generic("Missing resolve graph".into()))?
+          .nodes
+          .iter()
+          .sorted_by_key(|n| &n.id)
+          .filter_map(|node| self.create_crate_context(node, &package_to_checksum, catalog))
+          .collect::<Result<Vec<CrateContext>>>()
+      })
+      .collect::<Result<Vec<Vec<CrateContext>>>>()?;
+
+    Ok(contexts.iter().flat_map(|i| i.clone()).collect())
+  }
+
+  fn populate_sha256_index(&self, files: &[&CargoWorkspaceFiles]) -> Result<HashMap<(String, Version), String>> {
     // Gather the checksums for all packages in the lockfile
     // which have them.
     //
@@ -279,23 +301,19 @@ impl<'planner> WorkspaceSubplanner<'planner> {
       }
     }
 
-    let contexts = catalogs
-      .iter()
-      .map(|catalog| {
-        (*catalog)
-          .metadata
-          .resolve
-          .as_ref()
-          .ok_or_else(|| RazeError::Generic("Missing resolve graph".into()))?
-          .nodes
-          .iter()
-          .sorted_by_key(|n| &n.id)
-          .filter_map(|node| self.create_crate_context(node, &package_to_checksum, catalog))
-          .collect::<Result<Vec<CrateContext>>>()
-      })
-      .collect::<Result<Vec<Vec<CrateContext>>>>()?;
+    // If there's no lockfile, the above will miss packages - fill them in here.
+    for entry in &self.crate_catalog.entries {
+      let package = &entry.package;
+      let key = (package.name.to_string(), package.version.clone());
+      if package_to_checksum.get(&key).is_some() {
+        continue;
+      }
+      if let Ok(checksum) = fetch_crate_checksum(&self.settings.index_url, &package.name, &package.version.to_string()) {
+        package_to_checksum.insert(key, checksum);
+      }
+    }
 
-    Ok(contexts.iter().flat_map(|i| i.clone()).collect())
+    Ok(package_to_checksum)
   }
 }
 
@@ -757,5 +775,54 @@ impl<'planner> CrateSubplanner<'planner> {
         .into(),
       )
     }
+  }
+}
+
+
+#[cfg(test)]
+mod tests {
+  use crate::testing::make_workspace;
+  use crate::metadata::CargoMetadataFetcher;
+  use crate::planning::{BuildPlannerImpl, BuildPlanner};
+  use std::path::PathBuf;
+  use crate::settings::{GenMode, RazeSettings};
+  use std::collections::HashMap;
+
+  #[test]
+  fn sha256s() {
+    let settings = RazeSettings {
+      genmode: GenMode::Remote,
+      index_url: "https://github.com/rust-lang/crates.io-index".to_string(),
+
+      workspace_path: String::default(),
+      incompatible_relative_workspace_path: bool::default(),
+      target: None,
+      targets: None,
+      binary_deps: HashMap::default(),
+      crates: HashMap::default(),
+      gen_workspace_prefix: String::default(),
+      output_buildfile_suffix: String::default(),
+      default_gen_buildrs: bool::default(),
+      registry: String::default(),
+    };
+
+    let cargo_toml = r#"[package]
+  name = "test"
+  version = "0.0.1"
+
+  [lib]
+  path = "not_a_file.rs"
+
+  [dependencies]
+  lazy_static = "=1.4.0"
+  "#;
+
+    let (_tempdir, files) = make_workspace(cargo_toml, None);
+
+    let mut fetcher = CargoMetadataFetcher::default();
+    let mut planner = BuildPlannerImpl::new(&mut fetcher);
+    let planned_build = planner.plan_build(&settings, &PathBuf::new(), files, None).unwrap();
+    assert_eq!(planned_build.crate_contexts.len(), 1);
+    assert_eq!(planned_build.crate_contexts[0].sha256, Some("e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646".to_string()));
   }
 }
--- a/impl/src/metadata.rs
+++ b/impl/src/metadata.rs
@@ -206,11 +206,10 @@ pub fn fetch_crate_checksum(index_url: &str, name: &str, version: &str) -> Resul
       .ok_or(anyhow!("Failed to find crate '{}' in index", name))?
   };
 
-  let (_index, crate_version) = crate_index_path
+  let crate_version = crate_index_path
     .versions()
     .iter()
-    .enumerate()
-    .find(|(_, ver)| ver.version() == version)
+    .find(|ver| ver.version() == version)
     .ok_or(anyhow!(
       "Failed to find version {} for crate {}",
       version,


UebelAndre:
Do you know what the release cadence is for https://github.com/frewsxcv/rust-crates-index.git ? Maybe an issue can be created to try and push that through?

illicitonion:
> Do you know what the release cadence is for https://github.com/frewsxcv/rust-crates-index.git ? Maybe an issue can be created to try and push that through?

I mentioned that a release would be handy in a comment on https://github.com/frewsxcv/rust-crates-index/pull/42 - if there isn't one in the next few days I will nudge :)

UebelAndre:
Also, I feel bad doing drive by comments so thanks for the response 😄 but that said, I have to ask, is there a way to write the changes in `planning.rs` so the code isn't so dense/nested? 😅 

I'm not trying to get you to change it but I do think it'd be nicer 😅

kornholi:
I think the bigger issue is that after switching to `cargo-metadata` in #149, the lock file is no longer updated. Running `cargo generate-lockfile` before `cargo raze` works (and the hashes will be correctly propagated), but that forces updates you might not want. It's also an annoying manual step to remember to do.

One alternative option would be to improve `cargo-metadata` to include hashes in its output and skip parsing the lock file at all, but I don't think anyone has made a conscious decision about `cargo-raze` not updating the lock file. It's just something we didn't catch in #149.

UebelAndre:
@kornholi that's a good callout. I've noticed another issue in the switch to `cargo-metadata` in #218 (one I feel is actually pretty major)

kornholi:
Yep, we ran into that issue as well. I never got around to opening the issues on `cargo-metadata` side :( 

Anyways, I'm also hesitant on adding more crates.io-specific code as Cargo supports third-party registries. Raze currently hard codes the crates.io bucket URL in the template, but I think that's the only blocker for supporting them.

UebelAndre:
@kornholi I think that's worth creating an issue ticket for. I only ended up getting a cargo issue opened because some authors/contributors told me there was a bug and I asked them to open it 😃

If anyone is looking for something to do, I'd love to review some proposals to solve that problem or hear some suggestions.

Also, extra apologies to @illicitonion for totally side-railing this PR 😅

illicitonion:
> I think the bigger issue is that after switching to `cargo-metadata` in #149, the lock file is no longer updated. Running `cargo generate-lockfile` before `cargo raze` works (and the hashes will be correctly propagated), but that forces updates you might not want. It's also an annoying manual step to remember to do.

If a `Cargo.lock` already exists, anything still relevant from it will still be used, right?

> 
> One alternative option would be to improve `cargo-metadata` to include hashes in its output and skip parsing the lock file at all, but I don't think anyone has made a conscious decision about `cargo-raze` not updating the lock file. It's just something we didn't catch in #149.

Eric Huss said they were open to doing so in https://internals.rust-lang.org/t/stable-way-of-getting-registry-index-data/13071 but even after implementing, that would realistically be a multi-month wait for release cycles - I think this filling in is still valuable until that could happen (and this code automatically becomes a no-op when that's done).

> Anyways, I'm also hesitant on adding more crates.io-specific code as Cargo supports third-party registries. Raze currently hard codes the crates.io bucket URL in the template, but I think that's the only blocker for supporting them.

I've switched to the version of the code which is generic over all registries, rather than always assuming crates.io.

impl/src/planning.rs
```diff
@@ -493,6 +479,52 @@ impl<'planner> WorkspaceSubplanner<'planner> {
...
+      for package in lockfile.packages {
+        if let Some(checksum) = package.checksum {
+          package_to_checksum.insert(
+            (package.name.to_string(), package.version),
+            checksum.to_string(),
+          );
+        }
+      }
+    }
+
```
acmcarther, line 501: Add a comment here explaining what this second stage is doing.

impl/src/planning.rs
```diff
@@ -493,6 +479,52 @@ impl<'planner> WorkspaceSubplanner<'planner> {
...
+                      package_to_checksum.insert(key, hex::encode(version.checksum()));
+                      break;
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
```
acmcarther, line 523: This construction is a bit messy. Linear logic should not indent endlessly.

You can resolve most of this via early returns or continues and get back to a linear structure

```rust
    let key = (package.name.to_string(), package.version.clone());
    for package in &self.metadata.packages {
      if package_to_checksum.contains_key(&key) {
        // Checksum already populated from lockfile.
        continue;
      }
      let deprefixed_registry_src = package.source.as_ref()
        .map(|src| format!("{}", src)
        .and_then(source.strip_prefix("registry+"));
      if (deprefixed_registry_src.is_none()) {
        // Can't fetch checksum from crate which isn't from the registry
        continue;
      }
      let index = crates_index::BareIndex::from_url(&deprefixed_registry_src.unwrap())
        .and_then(|bare_index| bare_index.open_or_clone());
      if let Err(index_err) = index {
        // Could not open crate index. Maybe should log?
        continue;
      }      
      if let Some(crate_) = index.crate_(&package.name) {
        let want_version = package.version.to_string();
        for version in crate_.versions() {
           if version.version() == &want_version {
              package_to_checksum.insert(key, hex::encode(version.checksum()));
              break;
           }
        }
      }
    }

```

acmcarther (changes requested):
Sorry for the egregious delay here. I'm getting back into the groove of things again.

Some nits about code layout, but generally this LGTM. Is it possible to write a test for this/

UebelAndre:
Does anyone know why the checksums would be missing in the first place? My understanding is that if you're using a workspace then your `Cargo.lock` file gets generated at the workspace root which means updated dependencies won't have checksums or checksums might start getting outdated. It really seems like adding support for cargo workspaces would be the better approach here

impl/src/planning.rs
acmcarther: (this is largely optional, though I feel a bit strongly about it, so I'd hope you feel strongly if you don't want to make the change)

illicitonion:
Sorry for the huge delay in getting back here - on the plus side, in the mean time most of this logic has been added elsewhere! Hopefully this looks better now :)

> Does anyone know why the checksums would be missing in the first place?

If you don't have a lockfile, they'll be missing - I added a test case which fails without this change.

UebelAndre:
Would it be ok if this waited until #276 was merged? I think the implementation for this would be even simpler on top of tat commit. I'm doing something similar for binary dependencies there 😃

illicitonion:
> Would it be ok if this waited until #276 was merged? I think the implementation for this would be even simpler on top of tat commit. I'm doing something similar for binary dependencies there 😃

I think I'd rather merge this now if that's ok - that PR looks pretty large and complex, whereas this one is tiny, and I _think_ that the change #276 to incorporate this should be of the order of moving one line of code, and maybe moving a loop?

UebelAndre:
My general hesitation here is that I'm not sure if this is addressing a the core problem and not just a symptom. I feel like raze should ensure a complete lock file is available and not try to do some lazy (for lack of a better word) fetching when a crate is missing. I think adding workspace support removed one of the larger cases where users would have missing crates. The other case is that they simply forgot to generate one/update which can be solved easily by writing a new conditional in #276. What're your thoughts there?

illicitonion:
If/when we have code which means that we always have a lockfile we can use, I'm happy to delete this tiny patch, but as long as we don't, I'd rather merge it.

UebelAndre:
But the thing is, you wouldn't be deleting this patch, the author of the change to ensure a lockfile is always available would. And I feel this patch would definitely be deleted because if we're guaranteed to have a lockfile, this code would do nothing. I understand the reasoning for wanting a solution in place but I feel this is not a step to solving the real problem. Obviously I can't prevent this from being merged but again, I don't think it's the right solution.

illicitonion:
I understand that the code would need actively deleting if it were obsolete, and I'm happy to be the one who does it. But it's not clear to me that raze _should_ always be generating a `Cargo.lock` file. I'd rather not have to block a useful improvement on all of: a large pending PR, a design discussion around unconditional lockfile generation, and then implementation of that - particularly as we _already_ use these libraries to fetch these checksums. This is really a trivial patch which is very easy to back out when appropriate...

illicitonion:
Is there a reason you think this is actively problematic, rather than a little hacky?

UebelAndre:
I never said raze should always generate a lock file. It should try to make sure one is available though. Ideally raze would warn users that the given lock file is insufficient or generate a new one upon request. I think this approach is problematic because it puts users in a position where they're using half a lock file and whatever else is in their local index which seems like it'd lead to reproducibility issues. I'm specifically asking you to hold for #276 because I genuinely believe it will solve for most of these issues of missing checksums. The correct change for me is to check the validity of the lock file and correct/warn that before raze does any "planning".

illicitonion:
> I never said raze should always generate a lock file. It should try to make sure one is available though. Ideally raze would warn users that the given lock file is insufficient or generate a new one upon request.

I'm a little confused by this - if I pass `None` here: https://github.com/google/cargo-raze/blob/4eed19e0dfd0004ad0c044adbbad702764d4e93d/impl/src/metadata.rs#L47 because I'm happy to do a resolve every time I invoke raze, what are you suggesting raze does? Generate a lockfile for the sake of the metadata, which it will then throw away after it's finished reading the metadata?

> I think this approach is problematic because it puts users in a position where they're using half a lock file and whatever else is in their local index which seems like it'd lead to reproducibility issues.

The lack of a lockfile doesn't lead to them using whatever happens to be in their local index. `cargo` will update its index and perform a full resolve, just as cargo would if you didn't have a lockfile and did a `cargo build`. The only thing this change does is, after that resolve has already selected exact versions, it looks up the sha256 of the crate file. This change doesn't have any effect on what versions are selected - it's only used to fill in metadata _after_ those versions are selected. I'm not sure how this leads to inconsistency based on what's on disk? It will lead to inconsistency based on what's published to the index, but that's what you're asking for in the `cargo` world if you don't have a `Cargo.lock` file.

> I'm specifically asking you to hold for #276 because I genuinely believe it will solve for most of these issues of missing checksums. The correct change for me is to check the validity of the lock file and correct/warn that before raze does any "planning".

Does it generate a lockfile for the case of calling `MetadataFetcher::fetch_metadata` with `None` passed as the `CargoWorkspaceFiles.lock_path_opt`? My (not very detailed) read was that it didn't, and that it only generated a lockfile for binary deps?

Phrased another way: Does the test I add in this PR pass with your PR? Because that's exactly the behaviour I'm trying to support.

UebelAndre:
I need to sit down and read more than just the last thing you said. The changes in #276 are only generating a lock file when binary dependencies are present but it can easily be updated to do so when one is simply missing.

Again, I'll have more attention later but I do want to say I want to solve the problem this PR is solving. I can test your test on top of my changes tonight and report back.

UebelAndre:
I think https://github.com/UebelAndre/cargo-raze/commit/e18f67866c33162cf1f0919a7e9eea533ba6da58 does what you're looking for. The thought there is that anytime you run `cargo-raze` it'll generate a new lockfile (`Cargo.raze.lock`) and continue to use that until it's either missing or the user passes `--update` to force it to update (similar to running `cargo update` in your crate workspace).

I've noticed a few things that I'll also comment on your PR

impl/src/planning/subplanners.rs
```diff
@@ -759,3 +777,52 @@ impl<'planner> CrateSubplanner<'planner> {
...
+
+  [dependencies]
+  lazy_static = "=1.4.0"
+  "#;
+
+    let (_tempdir, files) = make_workspace(cargo_toml, None);
+
+    let mut fetcher = CargoMetadataFetcher::default();
+    let mut planner = BuildPlannerImpl::new(&mut fetcher);
+    let planned_build = planner.plan_build(&settings, &PathBuf::new(), files, None).unwrap();
```
UebelAndre, line 824: Any test in this file should be a direct test on a class in here. The fact that you're calling `planner.plan_build(...` indicates that this test should go into `planning.rs`

impl/src/planning/subplanners.rs
```diff
@@ -759,3 +777,52 @@ impl<'planner> CrateSubplanner<'planner> {
...
+    assert_eq!(planned_build.crate_contexts.len(), 1);
+    assert_eq!(planned_build.crate_contexts[0].sha256, Some("e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646".to_string()));
```
UebelAndre, line 826: This test also reaches out to the internet. It should be updated to avoid this to make it faster and more consistent.

impl/src/planning/subplanners.rs
```diff
@@ -759,3 +777,52 @@ impl<'planner> CrateSubplanner<'planner> {
...
+  use crate::testing::make_workspace;
+  use crate::metadata::CargoMetadataFetcher;
+  use crate::planning::{BuildPlannerImpl, BuildPlanner};
+  use std::path::PathBuf;
+  use crate::settings::{GenMode, RazeSettings};
+  use std::collections::HashMap;
+
+  #[test]
+  fn sha256s() {
+    let settings = RazeSettings {
```
UebelAndre, line 793: There's a helper function for creating a `RazeSettings` object `settings::tests::dummy_raze_settings` that should be used here instead

impl/src/planning/subplanners.rs
```diff
@@ -279,23 +301,19 @@ impl<'planner> WorkspaceSubplanner<'planner> {
...
-      })
-      .collect::<Result<Vec<Vec<CrateContext>>>>()?;
+    // If there's no lockfile, the above will miss packages - fill them in here.
+    for entry in &self.crate_catalog.entries {
+      let package = &entry.package;
+      let key = (package.name.to_string(), package.version.clone());
+      if package_to_checksum.get(&key).is_some() {
+        continue;
+      }
+      if let Ok(checksum) = fetch_crate_checksum(&self.settings.index_url, &package.name, &package.version.to_string()) {
```
UebelAndre, line 311: I generally feel this should be an error and we shouldn't be attempting to update the local index to grab a checksum during planning.

UebelAndre (changes requested):

UebelAndre:
Oh dang! https://github.com/bazelbuild/rules_rust/issues/2#issuecomment-738803545 is some great context for this! I apologize for being the naysayer here. I do still have a branch with some WIP on this and can jump back over to it this weekend 😄

UebelAndre:
Ok, I think I have something that solves for this. The last change on https://github.com/UebelAndre/cargo-raze/tree/checksum updates the metadata fetcher to always generate a lockfile if one isn't provided. If you have time, take a look (it's the last commit on the branch) and let me know if that would work for you.

UebelAndre:
I was finally able to open the PR! https://github.com/google/cargo-raze/pull/341

UebelAndre:
@illicitonion I think the functionality of this PR is satisfied by what was merged in https://github.com/google/cargo-raze/pull/341

Hope it helps 😄

illicitonion:
@UebelAndre Thanks! I tested it out, and it works great :) I've pushed an update to the code dump of rules_rust_external to use that!

PR #224, opened 2020-09-20
Added additional smoketests and improved testing speed

UebelAndre:
This change improves the speed of tests by combining all build steps into 1 command and avoids having to startup Bazel for each set of tests.

There is also now a test for a project that has no dependencies. This tests that raze doesn't generate broken files in this case

Finally, this change adds a `launch_test` macro which is used to create tests for each example binary and launch it as part of a test. This macro was created to try and keep the examples as straight forward as possible (not having to dump some test code in there). This should make it easier to catch regressions.


UebelAndre:
@damienmg Hey, this change is relatively small and I think pretty good for CI. Could I get you to take a look at it? 🙏

source-connect-test (approved):

source-connect-test:
Thanks!

damienmg:
Sorry for that, I used a test account rather than my own account :/ I feel bad. Anyway thanks & LGTM!

UebelAndre:
All good! 😃 thanks!

PR #225, opened 2020-09-20
Pruned vendored examples

UebelAndre:
This is a very large change but I think very beneficial for future contributors.

This change removes all vendored source committed to the repo and only leaves the generated BUILD files (which should be the only relevant thing for this repo to commit). Smoketest will still vendor the directories correctly but the source (and any future source) will be ignored by git. This should enable users to have more focused commits should they need to change or add a vendored example.


UebelAndre:
@damienmg This change is probably the most insane PR I've ever opened lol. In some of the testing I did for supporting Binary dependencies, I wanted to see if I could update all examples. I ended up regenerating `.lock` files (or rather vscode did) and I realized the vendored source code had been checked into the project. I dont' think that should be the case and as a contributor, I think it's nice to have diffs only include relevant changes. 

Sorry to reiterate the description but if you could first checkout #224 and then this, it'd help me out in getting my other (currently unopened) PR in.

UebelAndre:
After playing with this a bit I realize it's kinda a sacrifice for users who may just want to download and run the examples. Maybe all that should have been done here is to hard pin everything. I do think if there is a way to not have all the vendored source code committed to the repo that'd be ideal but since vendoring nukes the generated directories, there's no great way to have users build examples without running raze again.

UebelAndre:
@damienmg Ok, this now contains only the changes relevant to this pull request.

PR #226, opened 2020-09-20
Added support for custom registry urls

UebelAndre:
This change allows for users to change the registry url used in the output Bazel files. The new setting:

```toml
[raze]
registry = "https://crates-io.s3-us-west-1.amazonaws.com/crates/{crate}/{crate}-{version}.crate"
```

Is a templated string that will replace `{crate}` with the name of the crate and `{version}` with the version of the crate. This change should not impact any existing files generated from `v5.0.0` of `cargo-raze`

### DIFF
--- a/impl/src/bazel.rs
+++ b/impl/src/bazel.rs
@@ -567,6 +567,7 @@ mod tests {
         git_data: None,
       },
       sha256: None,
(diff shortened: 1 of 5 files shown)


UebelAndre:
@damienmg Hey, these changes are something that I pulled out of another thing I'm working on to support Binary dependencies. Even if those changes never get merged, I think this is a good quality of life improvement.

damienmg (approved):

PR #228, opened 2020-09-24
Change &String function parameters to &str

sitaktif:
The use of `&String` in function parameters is discouraged in favor of
`&str` as it otherwise forces passing heap-allocated values (which
doesn't include e.g. string literals).

A small contribution before other, more meaningful ones hopefully :-)


smklein:
Hey, thanks for the contribution!

PR #229, opened 2020-09-25
Add some text to help identify missing checksums

UebelAndre:
This adds a small comment in the place of missing checksum so they're easier to identify and track down

### DIFF
--- a/impl/src/templates/remote_crates.bzl.template
+++ b/impl/src/templates/remote_crates.bzl.template
@@ -31,6 +31,8 @@ def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
         type = "tar.gz",
 {%- if crate.sha256 %}
         sha256 = "{{crate.sha256}}",
+{%- else %}
+        # sha256 = TODO
 {%- endif %}
         strip_prefix = "{{crate.pkg_name}}-{{crate.pkg_version}}",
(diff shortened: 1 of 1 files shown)


UebelAndre:
@damienmg hey, would you mind taking a look at this one? Pretty small but nice. Any suggestions other than a 
```
TODO
```
?

damienmg:
I don't think we can do anything in the else case, why would we output a TODO here?

UebelAndre:
It makes it easier to track down which dependencies are missing sha's and update them. I was wondering if others would as well so I thought I'd make a pull request for it.

UebelAndre:
@damienmg I'm fine with not merging this. Just thought it might be nice. Let me know what you think 😄

damienmg:
Here the thing, whoever is getting that file cannot do much about providing a sha, correct? So I don't think we should surface a TODO that is not actionable.

UebelAndre:
Bazel will tell them what the SHA should be. I find it nice to be able to replace the `TODO` and uncomment the line. Though I can  understand the argument that if they're getting `TODO`s then they probably need to `cargo update` or regenerate their `.lock` file.

UebelAndre:
I'll just close this. In an ideal world this would never show up. Better to focus on that 😄

PR #232, opened 2020-09-25
Allow matching triples directly

sitaktif:
Before this change, the get_matching_bazel_triples would only allow matching cfg(...) targets. It seems that the intended behaviour was to allow passing triples (such as `x86_64-unknown-linux-gnu`) directly, as per the `format!("cfg(target = \"{}\")", target)` replacement at the beginning of the function. Unfortunately, the match was missing the right case below (`Predicate::Target(tp)` only matches builtin `target_{arch,os,...}`).

This makes this function more consistent with `is_bazel_supported_platform` above, and it seems to solve a bug where target specific dependencies specified with triples (such as is the case for the `winapi` crate in one of the `complicated_cargo_library` example) would be ignored.

FYI you can see the winapi targeted dependencies here: https://github.com/retep998/winapi-rs/blob/785f7f30a69c70687864ac650aebc123e942c1b3/Cargo.toml#L20-L23

### DIFF
--- a/impl/src/bazel.rs
+++ b/impl/src/bazel.rs
@@ -120,7 +120,7 @@ pub fn is_bazel_supported_platform(target: &str) -> (bool, bool) {
   (is_supported, matches_all)
 }
 
-/** Maps a Rust cfg target to a Bazel supported triples.
+/** Maps a Rust cfg or triple target to Bazel supported triples.
  *
  * Note, the Bazel triples must be defined in:
  * https://github.com/bazelbuild/rules_rust/blob/master/rust/platform/platform.bzl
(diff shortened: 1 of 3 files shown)


impl/src/bazel.rs
```diff
@@ -120,15 +120,15 @@ pub fn is_bazel_supported_platform(target: &str) -> (bool, bool) {
...
+/** Maps a Rust cfg or triple target to Bazel supported triples.
  *
  * Note, the Bazel triples must be defined in:
  * https://github.com/bazelbuild/rules_rust/blob/master/rust/platform/platform.bzl
  */
 pub fn get_matching_bazel_triples(target: &str) -> Result<Vec<String>> {
   let target_exp = match target.starts_with("cfg(") {
     true => target.to_owned(),
-    false => format!("cfg(target=\"{}\")", target),
+    false => format!("cfg(target = \"{}\")", target),
```
sitaktif, line 131: This is just for consistency with the `is_bazel_supported_platform` function above.

UebelAndre (approved):
Nice! Thanks for catching this 😄 

I don't have merge powers but the change looks good to me. Maybe @damienmg can take a look when he's around next 😅

damienmg:
LGTM, thanks a lot!

PR #236, opened 2020-10-08
Fixed rust-analyzer errors

UebelAndre:
This resolved the following issues flagged by rust-analyzer:

Screen Shot 2020-10-08 at 6 26 46 AM

That said though, there should be no behavioral difference in the code.


acmcarther:
LG

PR #237, opened 2020-10-08
Update smoketest

UebelAndre:
This is some cleanup to modernize smoketest's WORKSPACE but is also required for #227


acmcarther:
LGTM!

PR #238, opened 2020-10-08
Update registry url

UebelAndre:
This change is broken out of #227 and is necessary to be compatible with https://github.com/ehuss/cargo-clone-crate


acmcarther:
Thanks, this LGTM.

PR #239, opened 2020-10-08
fetch_metadata does not need '&mut self'

UebelAndre:
There's currently no reason for `fetch_metadata` to require `mut`. Cleaning this up


acmcarther:
LGTM, thanks!

PR #240, opened 2020-10-08
Binary dependencies smoketest

UebelAndre:
This contains the changes in #227 and should not be merged until that PR gets merged and this PR is rebased on top of that.


README.md
```diff
@@ -314,6 +314,24 @@ extra_aliased_targets = [
 Cargo-raze prefixes binary targets with `cargo_bin_`, as although Cargo permits
 binaries and libraries to share the same target name, Bazel disallows this.
 
+#### Crates that only provide binaries
```
acmcarther, line 317: Nit, I very very slightly prefer that this live with the feature PR, but it's probably more work to move this than I'd actually feel is justified.

acmcarther (approved):
I'm LGTMing blind here. too lazy to look through each and every file.

Anything here you think deserves some extra attention? Else, just LGTM.

examples/remote/binary_dependencies/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
+load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+rust_binary(
+    name = "binary_dependencies_bin",
+    srcs = ["src/main.rs"],
+    data = [
+        "//remote/binary_dependencies/cargo:cargo_bin_wasm_bindgen",
```
acmcarther, line 9: Oh actually I do have a comment.

Are there any binary deps with less transitive dependencies? That might cut down on the size of this in the repo

UebelAndre:
The only thing you should take a look at is 067d03220d29d4b6aace47093396060600752717

examples/remote/binary_dependencies/BUILD.bazel
UebelAndre: I can do some digging but this represents the very thing I'd like to address in `rules_rust`. Might I be able to do this in a future PR?

README.md
UebelAndre: This change is only here because it includes the changes from #227 It'll be removed from this PR once that goes through

examples/remote/binary_dependencies/BUILD.bazel
UebelAndre: Or do you know of another create that exhibits this behavior? (I'll try to look tonight either way)

UebelAndre: Found one with 100 less transitive dependencies

UebelAndre:
@acmcarther this PR has been rebased and is ready to merge. The only change worth taking a look at is d96e3ba780dfe40c2d6809caae4ed7e944089bf4

acmcarther:
LGTM

PR #212, opened 2020-09-04
Multi target support

UebelAndre:
Similar to #54 this pull request adds support for generating Bazel targets for multiple platforms. This is still kinda **WIP** but I wanted to share. I'll update this description as the feature evolves but for now am looking for any insight on what may be missing for the initial release of this feature.

Output `BUILD` file:
```python
"""
@generated
cargo-raze crate build file.

DO NOT EDIT! Replaced on runs of cargo-raze
"""

# buildifier: disable=load
load(
    "@io_bazel_rules_rust//rust:rust.bzl",
    "rust_binary",
    "rust_library",
    "rust_test",
)

# buildifier: disable=load
load("@bazel_skylib//lib:selects.bzl", "selects")

package(default_visibility = [
    # Public for visibility by "@raze__crate__version//" targets.
    #
    # Prefer access through "//cbindgen/raze", which limits external
    # visibility to explicit Cargo.toml dependencies.
    "//visibility:public",
])

licenses([
    "notice",  # MIT from expression "MIT"
])

# Generated targets
# Unsupported target "atty" with type "example" omitted

# buildifier: leave-alone
rust_library(
    name = "atty",
    crate_type = "lib",
    deps = [
    ] + selects.with_or({
        # cfg(unix)
        (
            "@io_bazel_rules_rust//rust/platform:aarch64-apple-ios",
            "@io_bazel_rules_rust//rust/platform:aarch64-linux-android",
            "@io_bazel_rules_rust//rust/platform:aarch64-unknown-linux-gnu",
            "@io_bazel_rules_rust//rust/platform:arm-unknown-linux-gnueabi",
            "@io_bazel_rules_rust//rust/platform:i686-apple-darwin",
            "@io_bazel_rules_rust//rust/platform:i686-linux-android",
            "@io_bazel_rules_rust//rust/platform:i686-unknown-freebsd",
            "@io_bazel_rules_rust//rust/platform:i686-unknown-linux-gnu",
            "@io_bazel_rules_rust//rust/platform:powerpc-unknown-linux-gnu",
            "@io_bazel_rules_rust//rust/platform:s390x-unknown-linux-gnu",
            "@io_bazel_rules_rust//rust/platform:x86_64-apple-darwin",
            "@io_bazel_rules_rust//rust/platform:x86_64-apple-ios",
            "@io_bazel_rules_rust//rust/platform:x86_64-linux-android",
            "@io_bazel_rules_rust//rust/platform:x86_64-unknown-freebsd",
[...]

PR #243, opened 2020-10-11
build script was missing platform-specific deps

dae:
Prior to #212 being merged, the following in Cargo.toml was included the resulting BUILD file:

```
[target.'cfg(target_env = "msvc")'.build-dependencies]
vcpkg = { version = "0.2", optional = true }
```

This patch makes it work again, gated with a platform check, eg:

```
cargo_build_script(
    name = "libsqlite3_sys_build_script",
    srcs = glob(["**/*.rs"]),
    crate_root = "build.rs",
    edition = "2018",
    deps = [
        "@raze__cc__1_0_61//:cc",
        "@raze__pkg_config__0_3_18//:pkg_config",
    ] + selects.with_or({
        # cfg(target_env = "msvc")
        (
            "@io_bazel_rules_rust//rust/platform:i686-pc-windows-msvc",
            "@io_bazel_rules_rust//rust/platform:x86_64-pc-windows-msvc",
        ): [
            "@raze__vcpkg__0_2_10//:vcpkg",
        ],
        "//conditions:default": [],
    }),
```

### DIFF
--- a/impl/src/templates/partials/build_script.template
+++ b/impl/src/templates/partials/build_script.template
@@ -18,7 +18,24 @@ cargo_build_script(
       {%- for dependency in crate.default_deps.build_dependencies %}
         "{{dependency.buildable_target}}",
       {%- endfor %}
-    ],
+    ]
+    {%- if crate.targeted_deps -%} 
+    {% for targeted_dep in crate.targeted_deps %} + selects.with_or({
+        # {{ targeted_dep.target }}
+        (
+    {%- for condition in targeted_dep.conditions %}
+            "{{ condition }}",
+    {%- endfor %}
+        ): [
+    {%- for dependency in targeted_dep.deps.build_dependencies %}
+            "{{ dependency.buildable_target }}",
+    {%- endfor %}
+        ],
+        "//conditions:default": [],
+    })
+    {%- endfor -%}
+    {%- else -%}
+    {%- endif %},
     {%- if crate.default_deps.build_proc_macro_dependencies %}
     proc_macro_deps = [
     {%- for dependency in crate.default_deps.build_proc_macro_dependencies %}


dae:
@googlebot I signed it!

dae:
@googlebot I signed it!

impl/src/templates/partials/build_script.template
```diff
@@ -18,7 +18,11 @@ cargo_build_script(
       {%- for dependency in crate.default_deps.build_dependencies %}
         "{{dependency.buildable_target}}",
       {%- endfor %}
-    ],
+    ]
+    {%- if crate.targeted_deps %} 
+    {%- include "templates/partials/targeted_build_dependencies.template" -%},
```
acmcarther, line 23: I think I weakly prefer the contents of this template to just be inlined here, unless you foresee another possible usage of it. Open to veto from @UebelAndre on this.

acmcarther (approved):
Generally LGTM, though I'd like a second opinion from @UebelAndre on this if available.

impl/src/templates/partials/build_script.template
UebelAndre: Yeah, I'd say either inline this or update `targeted_dependencies.template` to handle this case as well.

Inlining is probably the cleaner option (but I never was that great with jinja style templating)

UebelAndre:
This does appear to be a regression so this LGTM! Thanks for fixing that 😄 🙏

UebelAndre:
LGTM! Thanks for updating the PR 😄

UebelAndre:
@acmcarther will have to do final sign off though

dae:
The trailing comma was missing in the targeted deps case; I've moved it outside the endif to fix this.

acmcarther:
LGTM! Thanks for putting this together!

PR #244, opened 2020-10-11
platform-specific aliases were missing

### LINKED ISSUE #242: reqwests fails to compile with git HEAD
Carrying on from https://github.com/google/cargo-raze/pull/212#issuecomment-705507497

To reproduce, checkout https://github.com/ankitects/cargo-raze-bug-examples, checkout HEAD^, and run repro.sh. Tested on a Mac.

dae:
Closes #242

### DIFF
--- a/examples/remote/complicated_cargo_library/cargo/remote/libloading-0.5.2.BUILD.bazel
+++ b/examples/remote/complicated_cargo_library/cargo/remote/libloading-0.5.2.BUILD.bazel
@@ -61,6 +61,8 @@ rust_library(
     ],
     crate_features = [
     ],
+    aliases = {
+    },
 )
 # Unsupported target "markers" with type "test" omitted
 # Unsupported target "windows" with type "test" omitted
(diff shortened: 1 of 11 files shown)


acmcarther:
Travis says this is not producing Buildifier compliant BUILD files. I'm not totally clear why... perhaps it's causing
```
aliases = {
}
```

to be generated in some cases? Running ./smoke-test.sh and inspecting `examples/vendored/complicated_cargo_library/cargo/vendor/MacTypes-sys-2.1.0/BUILD.bazel` might be illuminating.

dae:
I wasn't able to remove aliases in all cases - suggestions welcome: https://github.com/google/cargo-raze/pull/244/commits/91c3999000dbc071ac224e2407fd0f82cc4685ba

UebelAndre:
I think it's fine to have aliases there so long as there's no impact on behavior or new buildifier defects being generated. I feel the templates are a bit hairy and should be streamlined. Maybe you could write a unit test that checks for some particular output so cleaning up the rendering logic would be safer to do in the future? It's up to @acmcarther though. Otherwise LGTM.

acmcarther:
Yeah, adding empty aliases SGTM.

LGTM. Thanks again!

PR #245, opened 2020-10-11
support MSVC targets

dae:
gnu has been removed in favour of MSVC in rules_rust, so cargo-raze will need to be updated to match:
https://github.com/bazelbuild/rules_rust/commit/dfce1232e0286623732871802964f76ac2b12098

### DIFF
--- a/examples/WORKSPACE
+++ b/examples/WORKSPACE
@@ -4,11 +4,11 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "io_bazel_rules_rust",
-    sha256 = "1c7da8ca29382c44b4431790983b4070d66278df7498c0ffe9390214b31a9d9f",
-    strip_prefix = "rules_rust-51b6a6d65cf609d10bd3b9cfe0a847a3bc7b0588",
+    sha256 = "50a772198877e21a61823fa292d28539f8bc99d72463e55b5b09942394ec370e",
+    strip_prefix = "rules_rust-9a8ef691b8e8f682d767189c38339cbee16d0a16",
     urls = [
         # Master branch as of 2020-09-24
-        "https://github.com/bazelbuild/rules_rust/archive/51b6a6d65cf609d10bd3b9cfe0a847a3bc7b0588.tar.gz",
+        "https://github.com/bazelbuild/rules_rust/archive/9a8ef691b8e8f682d767189c38339cbee16d0a16.tar.gz",
     ],
 )
 
--- a/examples/WORKSPACE
+++ b/examples/WORKSPACE
@@ -7,7 +7,7 @@ http_archive(
     sha256 = "50a772198877e21a61823fa292d28539f8bc99d72463e55b5b09942394ec370e",
     strip_prefix = "rules_rust-9a8ef691b8e8f682d767189c38339cbee16d0a16",
     urls = [
-        # Master branch as of 2020-09-24
+        # Master branch as of 2020-10-16
         "https://github.com/bazelbuild/rules_rust/archive/9a8ef691b8e8f682d767189c38339cbee16d0a16.tar.gz",
     ],
 )
--- a/examples/remote/binary_dependencies/cargo/remote/ansi_term-0.11.0.BUILD.bazel
+++ b/examples/remote/binary_dependencies/cargo/remote/ansi_term-0.11.0.BUILD.bazel
@@ -38,8 +38,8 @@ rust_library(
     ] + selects.with_or({
         # cfg(target_os = "windows")
         (
-            "@io_bazel_rules_rust//rust/platform:i686-pc-windows-gnu",
-            "@io_bazel_rules_rust//rust/platform:x86_64-pc-windows-gnu",
+            "@io_bazel_rules_rust//rust/platform:i686-pc-windows-msvc",
+            "@io_bazel_rules_rust//rust/platform:x86_64-pc-windows-msvc",
         ): [
             "@remote_binary_dependencies__winapi__0_3_8//:winapi",
         ],
--- a/examples/remote/binary_dependencies/cargo/remote/atty-0.2.14.BUILD.bazel
+++ b/examples/remote/binary_dependencies/cargo/remote/atty-0.2.14.BUILD.bazel
@@ -61,8 +61,8 @@ rust_library(
     }) + selects.with_or({
         # cfg(windows)
         (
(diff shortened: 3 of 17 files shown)


acmcarther:
I think this change results in updates to the examples, so we need to include the generated contents of ./smoke-test.sh in this PR.

(I'm not totally sure here, I'm basing this on seeing no output from the script on master).

UebelAndre:
Yeah, `./smoke-test.sh` needs to be ran and then the resulting files need to pass a builidifier check. These steps should probably be addded to `./CONTRIBUTING.md`

examples/remote/complicated_cargo_library/Cargo.toml
```diff
@@ -19,6 +19,25 @@ workspace_path = "//remote/complicated_cargo_library/cargo"
 gen_workspace_prefix = "remote_complicated_cargo_library"
 genmode = "Remote"
 incompatible_relative_workspace_path = true
+targets = [
```
UebelAndre, line 22: This should not be necessary, By specifying no targets, all targets should be included. Or did I miss something in this change?

examples/vendored/complicated_cargo_library/Cargo.toml
```diff
@@ -17,6 +17,26 @@ workspace_path = "//vendored/complicated_cargo_library/cargo"
 incompatible_relative_workspace_path = true
 gen_workspace_prefix = "vendored_complicated_cargo_library"
 
+targets = [
```
UebelAndre, line 20: Similar feedback as the `remote` variant

impl/src/bazel.rs
```diff
@@ -35,6 +35,8 @@ static SUPPORTED_PLATFORM_TRIPLES: &'static [&'static str] = &[
   "x86_64-apple-darwin",
   "x86_64-pc-windows-gnu",
   "x86_64-unknown-linux-gnu",
+  "i686-pc-windows-msvc",
```
UebelAndre, line 38: These cannot be added until they're added in https://github.com/bazelbuild/rules_rust/blob/master/rust/platform/platform.bzl

Raze currently relies on the platforms defined in `rules_rust` to do it's configuration switching. If you feel these targets should be supported, can you first get a pull request merged there? Then I'd ask that you update the revision of `rules_rust` specified in `smoke-test.sh` since that ultimately determines what gets output in the examples.

smoke_test/remote/complicated_cargo_library/Cargo.toml
```diff
@@ -19,6 +19,25 @@ workspace_path = "//remote/complicated_cargo_library/cargo"
 gen_workspace_prefix = "remote_complicated_cargo_library"
 genmode = "Remote"
 incompatible_relative_workspace_path = true
+targets = [
```
UebelAndre, line 22: This should not be necessary

smoke_test/vendored/complicated_cargo_library/Cargo.toml
```diff
@@ -17,6 +17,26 @@ workspace_path = "//vendored/complicated_cargo_library/cargo"
 incompatible_relative_workspace_path = true
 gen_workspace_prefix = "vendored_complicated_cargo_library"
 
+targets = [
```
UebelAndre, line 20: This should not be necessary

UebelAndre (changes requested):
Looks good so far but I think a PR will first need to go through to address https://github.com/bazelbuild/rules_rust/blob/master/rust/platform/platform.bzl#L19-L23

impl/src/bazel.rs
dae: Sorry, I left some comments in the commit but should have posted them here as well. This is not intended to be merged as-is: https://github.com/google/cargo-raze/pull/245/commits/efe960419f3aead5b88edcfb97b023abf219597b

acmcarther:
What's going on here? I'l review if needed...  ¯\\_(ツ)_/¯

UebelAndre:
@acmcarther this is blocked by https://github.com/bazelbuild/rules_rust/pull/425

dae:
Looks like there's an issue with the main branch's crate deps at the moment, and I'm unable to compile 0.6.0 locally. Need to sign off now; will pick this up again tomorrow.

UebelAndre:
It might be worth updating Travis to also build and test on windows.

UebelAndre:
You'll likely need to update the pin of `rules_rust` in `smoke-test.sh` now that  https://github.com/bazelbuild/rules_rust/pull/425 has gone through

UebelAndre:
@googlebot I consent.

UebelAndre:
@dae You need to apply the following diff as well
```diff
diff --git a/smoke-test.sh b/smoke-test.sh
index be8a95df..a8d8c2dd 100755
--- a/smoke-test.sh
+++ b/smoke-test.sh
@@ -40,11 +40,11 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "io_bazel_rules_rust",
-    sha256 = "1c7da8ca29382c44b4431790983b4070d66278df7498c0ffe9390214b31a9d9f",
-    strip_prefix = "rules_rust-51b6a6d65cf609d10bd3b9cfe0a847a3bc7b0588",
+    sha256 = "50a772198877e21a61823fa292d28539f8bc99d72463e55b5b09942394ec370e",
+    strip_prefix = "rules_rust-9a8ef691b8e8f682d767189c38339cbee16d0a16",
     urls = [
-        # Master branch as of 2020-09-24
-        "https://github.com/bazelbuild/rules_rust/archive/51b6a6d65cf609d10bd3b9cfe0a847a3bc7b0588.tar.gz",
+        # Master branch as of 2020-10-16
+        "https://github.com/bazelbuild/rules_rust/archive/9a8ef691b8e8f682d767189c38339cbee16d0a16.tar.gz",
     ],
 )
 

```

dae:
Thanks, I'd been looking at the summary diff, assumed that change was in the latter commit instead of the former, and was in the process of digging into why the build was failing :-)

UebelAndre:
All good, I should have split that up. I always try to keep regenerated examples as a separate commit at the end. But I should have split that change out from the original (ie. made my PR a 3 commit PR)

dae:
No problem, I appreciate the help.

acmcarther:
LGTM

UebelAndre:
@acmcarther do you have time today to make a new release for this change?

acmcarther:
After I wrote that script it only takes me like one minute to do this, so no problem.

Pushed, 0.6.1.

UebelAndre:
Thank you so much!

PR #248, opened 2020-10-13
Move util functions out of `bazel` and into `util`

UebelAndre:
This is the first commit of work I'm doing in https://github.com/UebelAndre/cargo-raze/tree/sorting 
The goal of that branch is to rearrange parts of the code to be in better alignment with the originally intended code layout.

It seems as though `bazel.rs` is intended to contain rendering logic so the utility functions used for planning should not go in here. This PR moves them into `util.rs`.

### DIFF
--- a/impl/src/bazel.rs
+++ b/impl/src/bazel.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 
 use tera::{self, Context, Tera};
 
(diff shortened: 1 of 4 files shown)


UebelAndre:
@acmcarther This is the first of many I'd like to do. Again, please check out after reviewing this pull request and let me know what your general thoughts are on https://github.com/UebelAndre/cargo-raze/tree/sorting (see `impl/src`)

(warning: that branch has a bunch of changes but it's only moved code)

acmcarther:
LGTM

PR #250, opened 2020-10-14
Moved RazeError into it's own `error` module

UebelAndre:
This is the next commit in https://github.com/UebelAndre/cargo-raze/tree/sorting following #248 

It looks like most of `util` is actually just used for `planning`. I think this points that out and means it can likely be removed or moved into the "future" `planning` module.


UebelAndre:
@acmcarther Part two 😅

impl/src/error.rs
```diff
@@ -0,0 +1,84 @@
...
+// limitations under the License.
+
+use std::fmt;
+
+pub const PLEASE_FILE_A_BUG: &str =
+"Please file an issue at github.com/google/cargo-raze with details.";
+
+#[derive(Debug)]
+pub enum RazeError {
+Generic(String),
```
acmcarther, line 22: Is this indented correctly?

impl/src/error.rs
```diff
@@ -0,0 +1,84 @@
...
+Config {
+  field_path_opt: Option<String>,
+  message: String,
+},
+}
+
+impl std::error::Error for RazeError {}
+
+impl fmt::Display for RazeError {
+fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
acmcarther, line 41: Same here?

impl/src/lib.rs
```diff
@@ -12,12 +12,22 @@
...
 // limitations under the License.
 
-pub mod bazel;
 pub mod context;
-pub mod license;
+pub mod error;
 pub mod metadata;
-pub mod planning;
-pub mod rendering;
+pub mod planning {
```
acmcarther, line 18: This seems weird.

Why not src/planning/mod.rs? Would also allow you to ditch the weird "planning::planning"

impl/src/lib.rs
```diff
@@ -12,12 +12,22 @@
...
+  mod checks;
+  mod crate_catalog;
+  mod license;
+  mod planning;
+  mod subplanners;
+  pub use planning::*;
+}
+pub mod rendering {
+  pub mod bazel;
+  mod rendering;
```
acmcarther, line 28: I think we should avoid this weird rendering::rendering thing if possible. Use `mod.rs` instead?

acmcarther (changes requested):
This PR name is misleading. The more important change in this PR is the breaking up of planning into multiple sub modules. The RazeError change seems incidental.

A few minor comments otherwise

impl/src/error.rs
UebelAndre: Interesting, I thought I ran `rustfmt` here. I'm not sure how that indentation was created.

acmcarther:
I think there's a balance to be had in organizing code. Some folks tend toward very coarse grained projects with few modules, some go the other way and have many may modules.

I tend toward the "have few modules" camp because I don't think rust's module system is actually very useful (and instead, just imposes maintenance burden), but I feel that way about a lot of things in rust, and I think the consensus is against me so  ¯\\\_(ツ)\_/¯

Anyway, the guiding principle should be "is this change useful", not "does this better align with some platonic ideal usage of the module system". This PR generally looks fine to me, but I wanted to make that point to make sure that you feel this change in organization is not only "more idiomatic", but is also "useful", particularly for folks new to the code to more quickly build an intuition about it.

UebelAndre:
OH GOD! This is the wrong branch!

UebelAndre:
@acmcarther sorry, please take a look at #251

impl/src/lib.rs
UebelAndre: it is, I didn't know if that would be more offensive to people (though I feel it's the correct thing to do)

UebelAndre:
> I think there's a balance to be had in organizing code. Some folks tend toward very coarse grained projects with few modules, some go the other way and have many many modules.
> 
> I tend toward the "have few modules" camp because I don't think rust's module system is actually very useful (and instead, just imposes maintenance burden), but I feel that way about a lot of things in rust, and I think the consensus is against me so ¯\_(ツ)_/¯
> 
> Anyway, the guiding principle should be "is this change useful", not "does this better align with some platonic ideal usage of the module system". This PR generally looks fine to me, but I wanted to make that point to make sure that you feel this change in organization is not only "more idiomatic", but is also "useful", particularly for folks new to the code to more quickly build an intuition about it.

My preference is always to make the critical code path as clear as possible. I feel very strongly that given an entry point, it should be easy to read through the code and get the gist of what it's doing. However, this becomes incredibly difficult when you have 1000+ line files like `planning.rs`. It also makes it harder to understand what has tests. Using directory modules for large modules is better because it makes it easier to identify and read the critical path.

PR #251, opened 2020-10-14
Moved `RazeError` into an `error` module

UebelAndre:
This is the next commit in https://github.com/UebelAndre/cargo-raze/tree/sorting following #248 

It looks like most of `util` is actually just used for `planning`. I think this points that out and means it can likely be removed or moved into the "future" `planning` module.


acmcarther (approved):
LGTM.

My diatribe from https://github.com/google/cargo-raze/pull/250#issuecomment-708554640 applies here as well (though I feel `error` is a very compelling choice for a separate module).

Will wait for CI

impl/src/error.rs
```diff
@@ -0,0 +1,84 @@
...
+      Self::Generic(s) => write!(f, "Raze failed with cause: \"{}\"", s),
+      Self::Internal(s) => write!(
+        f,
+        "Raze failed unexpectedly with cause: \"{}\". {}",
+        s, PLEASE_FILE_A_BUG
+      ),
+      Self::Config {
+        field_path_opt,
+        message,
+      } => match field_path_opt {
```
acmcarther, line 52: Optional: Can we flatten these matches by moving the subexpression into the match?

```
      Self::Config {
        Some(path),
        message,
      } => {}
      Self::Config {
        None,
        message,
      } => {}
```

?

acmcarther (changes requested):
Minor request, else, LGTM

impl/src/error.rs
UebelAndre: I don't think I understand the request here. Can you elaborate a bit more?

UebelAndre: Also, if possible, I'd like to prevent making code changes when moving code around so the history is easier to follow. Can I put this into another PR?

acmcarther: You're basically asking me not to request changes for your PRs. In general, hard no (come on man, it's a _code review_), but in this case since the code was this way before anyway, OK.

Per the [rust book](https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html), Rust supposedly supports complex match expressions, where the pattern can consist of "nested" elements, meaning that a Type(Field: Optional(Int)) could be matched with Type { Field: Some(5) }. I'm asking you to make this change here, instead of pattern matching into another pattern match.

acmcarther: (Will LGTM and merge after ack)

UebelAndre: I'm definitely **not** asking you to **not** request changes. As I originally asked, I wasn't sure how to implement the changes you had requested. 

I was specifically thinking that because this moves code, that it would be beneficial to keep changes to minimum so if the change were to introduce a regression, that it's easier to find if it were done in a separate PR. I'll read that link ASAP and see if it makes more sense to me

UebelAndre: I think I understand the changes you were asking for and I think I've correctly implemented them. Is this correct?

acmcarther:
Thank you for making the change! I'm not sure generally when a given piece of code is done some way because of personal taste or because a feature which makes it (to my eyes) cleaner is obscure. I'm kind of a luddite for many kinds of syntax (e.g. I pushed back on the `::{a, b, c}` import syntax for a _really_ long time), but I guess I should stop projecting. Most people are cool with new syntax.

Will merge on green.

acmcarther (approved):

UebelAndre:
> Thank you for making the change! I'm not sure generally when a given piece of code is done some way because of personal taste or because a feature which makes it (to my eyes) cleaner is obscure. I'm kind of a luddite for many kinds of syntax (e.g. I pushed back on the `::{a, b, c}` import syntax for a _really_ long time, and I'm still bitter about `?` and `try!` being deprecated), but I guess I should stop projecting. Most people are cool with new syntax.
> 
> Will merge on green.

I'm pretty new to Rust so basically the way it is now is all I know.

I just wanna say that I appreciate all your feedback and have learned a lot because of it. I would never ask you to **not** review my changes or make suggestions and I hope to never come off as annoying or stubborn. I think you're an amazing engineer and you've created a tool that really inspires me. So I thank you for that and want to continue to improve this project and learn everything I can along the way :smile:

Issue #220, opened 2020-09-15
Updated `genmode` default to "Remote"


UebelAndre:
I think the common case for `cargo raze` is to use `Remote` generation. Maybe we could update the default? When I was first learning to use the tool, the extra steps for vendoring were a bit confusing. Maybe the documentation just needs to be updated but again, I think remote `genmode` is more common.

Would love some input here and if there aren't any objections, I'll open a pull request at some point.

acmcarther:
We discussed this in #252. The approach being taken now is (1) warn when genmode is absent, then one or two releases later (2) make genmode mandatory.

(this sidesteps this issue, largely).

PR #252, opened 2020-10-15
Added warning for missing 'genmode' setting

UebelAndre:
This implements #220

It seems that the `Remote` genmode has more functionality and is the more common use case. It should be the default.


acmcarther:
This needs a more graceful rollout. We should warn for at least one release when users fail to specify the genmode that the default is changing.

UebelAndre:
Should users always specify a `genmode`?

acmcarther:
Yeah, maybe we should force genmode to be mandatory. One 0.X.0 release with a console warning, then it's an error in 0.(X+1).0?

UebelAndre:
@acmcarther How's this?

impl/src/bin/cargo-raze.rs
```diff
@@ -144,16 +146,18 @@ fn main() -> Result<()> {
...
-      write_to_file_loudly(&path, &contents)?;
+      write_to_file(&path, &contents, options.flag_verbose > 0)?;
     }
   }
 
   Ok(())
 }
 
-fn write_to_file_loudly(path: &Path, contents: &str) -> Result<()> {
+fn write_to_file(path: &Path, contents: &str, verbose: bool) -> Result<()> {
```
acmcarther, line 156: This should probably be a separate PR or at least mentioned in the PR description.

impl/src/settings.rs
```diff
@@ -288,6 +288,7 @@ fn default_raze_settings_field_gen_workspace_prefix() -> String {
 }
 
 fn default_raze_settings_field_genmode() -> GenMode {
+  eprintln!("WARNING: The [raze] setting `genmode` should be explicitly set to either `\"Remote\"` or `\"Vendored\"`");
```
acmcarther, line 291: Mention that this will be mandatory soon (that's the most important thing to convey).

Also, if you're open to it, i'd actually prefer a different implementation, which adds an "Unspecified" variant to GenMode, and maps that to Vendored in the implementation (for now), and prints this warning here: https://github.com/google/cargo-raze/blob/7cd1f6fee5aa3ca539106d5030281f3a4e950b1e/impl/src/bin/cargo-raze.rs#L72-L73

I don't feel strongly about this and I think there are good arguments against dealing with the Unspecified variant everywher.

acmcarther (changes requested):
Two nits

impl/src/settings.rs
acmcarther: Oh, actually, here is better: https://github.com/google/cargo-raze/blob/852a983b3956941e4ae1ceb2fbe4a858ad931c9f/impl/src/settings.rs#L330-L350

impl/src/bin/cargo-raze.rs
UebelAndre: Which would you prefer? I can do either

impl/src/settings.rs
UebelAndre: Yeah, I can take care of this later today.

impl/src/bin/cargo-raze.rs
UebelAndre: I made #263

impl/src/settings.rs
UebelAndre: Done

impl/src/planning/subplanners.rs
```diff
@@ -122,6 +122,8 @@ impl<'planner> WorkspaceSubplanner<'planner> {
       GenMode::Vendored => {
         checks::check_all_vendored(self.crate_catalog.entries(), &self.settings.workspace_path)?;
       },
+      // TODO: Remove when `Unspecified` is removed.
```
acmcarther, line 125: Do you know if serde support requiring a field? I looked earlier and didn't see anything... though this might be a dumb question that's obvious if you've used it recently.

If they don't support required fields, then we may never be able to remove this.

acmcarther (approved):

impl/src/planning/subplanners.rs
UebelAndre: Good question, I could have sworn it did but now I'm not so sure... Need to look

UebelAndre: Yeah, good call, I don't think it does support flagging something as a required field. I've updated the code to account for this.

impl/src/bin/cargo-raze.rs
```diff
@@ -134,6 +134,8 @@ fn main() -> Result<()> {
 
       bazel_renderer.render_remote_planned_build(&render_details, &planned_build)?
     }, /* exhaustive, we control the definition */
+    // Settings should always have `genmode` set to one of the above fields
+    GenMode::Unspecified => unreachable!(),
```
acmcarther, line 138: We can probably just return empty list here. In fact, that might actually be an interesting behavior because it means we've just done planning but we're not going to write anything.

(i'm going to make roughly the same comment everywhere)

impl/src/planning/crate_catalog.rs
```diff
@@ -121,6 +121,8 @@ impl CrateCatalogEntry {
         "vendor/{}/{}",
         &self.package_ident, settings.output_buildfile_suffix,
       ),
+      // Settings should always have `genmode` set to one of the above fields
+      GenMode::Unspecified => unreachable!(),
```
acmcarther, line 125: I'd actually prefer an explicit panic here (or local_build_path to be updated to return Result<>)

My second preference would be to explain in the comment why this is unreachable rather than just assert that it's unreachable (in fact, I don't think it is -- going through validate_settings may not be only conceivable path to this point).

impl/src/planning/crate_catalog.rs
```diff
@@ -140,6 +142,8 @@ impl CrateCatalogEntry {
           format!("{}/vendor/{}", settings.workspace_path, &self.package_ident)
         }
       },
+      // Settings should always have `genmode` set to one of the above fields
+      GenMode::Unspecified => unreachable!(),
```
acmcarther, line 146: roughly the same here -- First preference is Result<String>, second preference is explicit panic, final and least preferred is to attach a better explanation for why this is "unreachable".

impl/src/planning/crate_catalog.rs
```diff
@@ -167,6 +171,8 @@ impl CrateCatalogEntry {
           )
         }
       },
+      // Settings should always have `genmode` set to one of the above fields
+      GenMode::Unspecified => unreachable!(),
```
acmcarther, line 175: Same thing here.

(I generally would _never_ use `unreachable` unless there's something in this function that makes that code unreachable. Having the setting be inaccessible because some far code prevents it doesn't feel very unreachable to me... or at least, it's not a very strong guarantee).

impl/src/planning/subplanners.rs
```diff
@@ -122,6 +122,8 @@ impl<'planner> WorkspaceSubplanner<'planner> {
       GenMode::Vendored => {
         checks::check_all_vendored(self.crate_catalog.entries(), &self.settings.workspace_path)?;
       },
+      // Settings should always have `genmode` set to one of the above fields
+      GenMode::Unspecified => unreachable!(),
```
acmcarther, line 126: We can just do nothing here, rather than "unreachable!()".

impl/src/settings.rs
```diff
@@ -346,6 +347,11 @@ fn validate_settings(settings: &mut RazeSettings) -> Result<(), RazeError> {
     settings.workspace_path.pop();
   }
 
+  if settings.genmode == GenMode::Unspecified {
+    eprintln!("WARNING: The [raze] setting `genmode` is unspecified. Not specifying `genmode` is deperacated. Please explicitly set it to either \"Remote\" or \"Vendored\"");
+    settings.genmode = GenMode::Vendored;
```
acmcarther, line 352: No requested change, just a continuation of my comment on "unreachable!".

You used "unreachable!" in that other code because of this code here, but I think that this code is too far removed from that code, and there's no sure guarantee that someone won't change this code and suddenly make that code reachable. Hence, my preference for different handling of this case.

---

Actually, on further inspection, it turns out that current code *definitely* could reach that other code. `validate_settings` is only called during `load_settings`. CrateSettings is actually a public struct, and therefore its fields could have any of their possible values.

acmcarther (changes requested):
Nit about `unreachable`.

impl/src/planning/subplanners.rs
UebelAndre: I feel like this communicates the intent though can you share more on why you're feeling it would better to do nothing over `unreachable`?

UebelAndre: This might have been answered in your other comment

impl/src/settings.rs
UebelAndre: What should happen in this case?

impl/src/planning/subplanners.rs
acmcarther: For me intent is secondary to defensive programming. I interpret defensive programming in this case to mean "do the least surprising thing", and "don't expect upstream code to stay static (meaning unchanging)".

That tells me two things. Since the validation logic is (somewhat far) upstream of us, to be defensive, we should handle all possible inputs within reason. This variant is an explicit possible variant, and there's really very little preventing future changes to the code from toggling genmode into this state for reasons we can't currently predict.

This particular case is my favorite kind of case, because we can do "nothing" and the behavior won't be surprising. In this case, the switch is only used to enable special per-genmode validation logic. Since we're not generating anything, we have no validation to do.

impl/src/settings.rs
acmcarther: I think this code here should stay the same. Basically, everywhere else I'm saying "don't rely on that code!".

UebelAndre: Hey, I think I addressed everything in your other comments which is why I just marked them as resolved. Let me know what you think

acmcarther (approved):
LGTM!

Will merge on green though I might miss it and have to come back.

PR #254, opened 2020-10-16
Updated windows platforms from `gnu` to `msvc`

UebelAndre:
Not trying to steal the thunder from @dae but there is currently an incompatibility between `cargo-raze` and head of `rules_rust` This PR is the same change as #245 but with updated smoke-tests (the only thing broken about the PR). I'd love to see this get merged (or #245 get updated and merged) and then a release be made.

### DIFF
--- a/examples/WORKSPACE
+++ b/examples/WORKSPACE
@@ -4,11 +4,11 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "io_bazel_rules_rust",
-    sha256 = "1c7da8ca29382c44b4431790983b4070d66278df7498c0ffe9390214b31a9d9f",
-    strip_prefix = "rules_rust-51b6a6d65cf609d10bd3b9cfe0a847a3bc7b0588",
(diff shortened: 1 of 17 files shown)


acmcarther:
I would prefer to submit @dae's PR, but I don't feel strongly about this, particularly if we're fixing an important breakage. I'll check in on both of these in a couple of hours and submit this one if there isn't forward movement.

UebelAndre:
I would prefer to have that PR merged as well, the fix is quite small but yeah, I don't know when he'll be around next. There was a breakage in `rules_rust` that was fixed by regenerating some dependencies using these changes and I feel it's important to get this fix out in a timely manner to ensure the rules and this tool are compatible with each other.

Definitely your call to make @acmcarther

dae:
Sorry for the delay, I've pushed Uebel's change into my branch as well. Please feel free to merge either - fixing the breaking build is more important than credit for a few character change. :-)

UebelAndre:
@acmcarther #245 is ready to merge. Closing this ticket as a duplicate

PR #255, opened 2020-10-16
Split the `rendering` module into a directory

UebelAndre:
Follow up of #251 

I think this makes it crystal clear that `bazel.rs` is expected to contain the rendering logic for rendering Bazel files. I had first thought it was a utility module similar to `util.rs`.


impl/src/rendering/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub mod bazel;
+mod rendering;
+
+pub use rendering::*;
```
acmcarther, line 18: Inline the contents of rendering.rs? I don't see a need for a separate file now.

UebelAndre: I find it easier to work with the code base with the file still named as `rendering.rs`. My feeling is that if you have a reason to a directory for a module, `mod.rs` should just be used to describe the structure. Otherwise it gets annoying having to identify which `mod.rs` you're actually working in should you have multiple modules (which I would like to do with `planning.rs`. Do you still feel I should inline the contents of that file?

acmcarther: Can you find any examples of other projects that follow this convention?

I'm _very_ sympathetic to the concern of having multiple hard-to-distinguish mod.rs files, but I more strongly weigh convention (and avoidance of the "rendering/rendering" stutter). If you got this preference from some other projects though, I'm open to it.

UebelAndre: I'll look. I was first given this suggestion in the Rust Discord channel by one of the regulars there but I can't say that I've seen it in other projects (but I also haven't spent a lot of time reading through other projects)

UebelAndre: It's hard to tell.

### inlined
https://github.com/hyperium/hyper/tree/master/src/client/connect
https://github.com/amethyst/amethyst/tree/master/amethyst_audio/src/components

### separated
https://github.com/rustwasm/wasm-bindgen/tree/master/src/cache
https://github.com/servo/servo/tree/master/components/layout/display_list

The thing I definitely don't see is `module::module` Even in cases where there's basically one implementation [(wasm-bindgen)](https://github.com/rustwasm/wasm-bindgen/tree/master/src/cache) there isn't a source that matches the name of the directory. I'd much rather keep implementations out of `mod.rs` if they're going to be longer than 2 functions and a couple of tests. Could you suggest a name to rename that file to instead? If not I can reluctantly inline it 😅 (but again, would love to avoid using `mod.rs`)

acmcarther: This is probably my least favorite part of rust, and is part of the reason I don't really write rust anymore. We're already spending a lot of effort on thinking about how these files and modules should be organized far out of proportion of how much it actually matters. Part of our thought process becomes "how do I deal with this terrible `mod.rs` naming convention that makes my files hard to deal with in my editor".

Hence, my earlier position of generally avoiding the rust module system when possible. Unfortunately, the rust language bakes these file-to-module mapping conventions into the operations of the compiler itself.

Just awful, not only because of this thing here.

---

I've been doing some googling, and I've been lead to believe that a file structure like this might work

- ./rendering.rs
- ./rendering/bazel.rs

Can you try that?

UebelAndre: Yeah, this works. Take a look now.

acmcarther:
Awesome, whew. This LGTM, just waiting for green.

impl/src/rendering/mod.rs
acmcarther: marking resolved.

PR #257, opened 2020-10-16
Split the `planning` module into multiple submodules

UebelAndre:
This is the superset of all the restructured changes I was thinking of making.


UebelAndre:
@acmcarther Can you check this PR out and let me know what your general thoughts are. My plan was to make a PR for each of these but It'd probably help for you to see the final product of everything too.

acmcarther:
This looks in general OK to me, though I don't think the subplanner types deserve their own module. I don't think that's a sufficiently robust abstraction to merit separation from the rest of the planning logic.

I don't really feel strongly about it though.

UebelAndre:
The motivation here was to remove all the cruft from the `planning` module that that distracted from the critical path. I feel like the subplanners are pretty well defined and so should go in their own module. The thing I like most about this change is that I feel `planning.rs` now better represents the work that it's attempting to do. You can read through the implementation and tests and better reason about things.

acmcarther (approved):
I would make the point that something "distracting from the critical path" is not an objective judgment, but rather a function of your comfort with both the code and with the language, and your reticence (or lackthereof) with imposed structure vis a vis the module system. I would raise the counterpoint that there is a cost to be paid by readers having to look up the definition of a function or type used in one file but defined in another, particularly if the type is only used in one place. This principle applies here in planning (where the functionality being moved out is directly related to planning). 

You might make the argument that we can convey this relationship by making the code into a "submodule" of planning. I don't have a great counter argument for that except to convey my distaste for excessive (rust module related) structure, but that isn't not really an objective measure we can use to judge things.

_TL;DR: there is a balance to be made between keeping like things co-located, and keeping individual files concise and focused, and decisions made on this are largely personal value judgments._

---

That all said, I don't really feel strongly about this refactoring, so I'll LGTM it. I just wanted to make that point.

acmcarther:
I guess, i can further contextualize this with a picture into what I look at day to day:

Some source files I work on regularly are 5-15k LOC in C++, Java, and Go. Obviously these are not exemplars of quality or good practice (in fact, they're the opposite), but regardless, my experience dealing with those awful files has greatly increased my tolerance for how much stuff is happening in a single file, and I think that was beneficial for me.

UebelAndre:
What more can be done to clean up the codebase?

PR #258, opened 2020-10-17
Moved components of the `planning` module into a directory

UebelAndre:
This is the next commit after #255


UebelAndre:
@acmcarther the diff here will get smaller once #255 is merged.

acmcarther:
I'm not really convinced that moving license into planning enhances clarity, but I don't feel strongly about it.

LGTM.

acmcarther (approved):

acmcarther:
I guess this can just be merged without being synced, so I'll do that.

UebelAndre:
`planning` is the only module that uses `license` so I thought that's where it should go. If this doesn't make sense lets move it. But this is the first change in a larger change I want to make

PR #259, opened 2020-10-17
Moved `checks` out of `planning.rs`

UebelAndre:
This is the next change after #258


UebelAndre:
@acmcarther I feel like this is the module that set me on this path lol.

UebelAndre:
This was done in #257

Issue #192, opened 2020-08-04
Generated BUILD files are not properly formatted.


PiotrSikora:
Generated `BUILD` files are not properly formatted per [buildifier](https://github.com/bazelbuild/buildtools/tree/master/buildifier).

```
$ buildifier -lint=fix cargo/BUILD
$ git diff cargo/BUILD
```
```
diff --git a/cargo/BUILD b/cargo/BUILD
index a8639e5..832c24e 100644
--- a/cargo/BUILD
+++ b/cargo/BUILD
@@ -3,26 +3,31 @@ cargo-raze workspace build file.
 
 DO NOT EDIT! Replaced on runs of cargo-raze
 """
+
 package(default_visibility = ["//visibility:public"])
 
 licenses([
-  "notice" # See individual crates for specific licenses
+    "notice",  # See individual crates for specific licenses
 ])
+
 alias(
     name = "chrono",
     actual = "@raze__chrono__0_4_13//:chrono",
     tags = ["cargo-raze"],
 )
+
 alias(
     name = "hashbrown",
     actual = "@raze__hashbrown__0_7_2//:hashbrown",
     tags = ["cargo-raze"],
 )
+
 alias(
     name = "log",
     actual = "@raze__log__0_4_11//:log",
     tags = ["cargo-raze"],
 )
+
 alias(
     name = "wee_alloc",
     actual = "@raze__wee_alloc__0_4_5//:wee_alloc",
```
```
$ buildifier -lint=fix cargo/remote/log-0.4.11.BUILD
$ git diff cargo/remote/log-0.4.11.BUILD
```
```
diff --git a/cargo/remote/log-0.4.11.BUILD b/cargo/remote/log-0.4.11.BUILD
index ec0b13c..5ed228c 100644
--- a/cargo/remote/log-0.4.11.BUILD
+++ b/cargo/remote/log-0.4.11.BUILD
@@ -3,45 +3,43 @@ cargo-raze crate build file.
 
 DO NOT EDIT! Replaced on runs of cargo-raze
 """
-package(default_visibility = [
-  # Public for visibility by "@raze__crate__version//" targets.
-  #
-  # Prefer access through "//cargo", which limits external
-  # visibility to explicit Cargo.toml dependencies.
-  "//visibility:public",
-])
-
-licenses([
-  "notice", # MIT from expression "MIT OR Apache-2.0"
-])
 
 load(
     "@io_bazel_rules_rust//rust:rust.bzl",
     "rust_library",
[...]

PR #260, opened 2020-10-17
Made generated files more Buildifier compliant

UebelAndre:
This addresses https://github.com/google/cargo-raze/issues/192#issuecomment-693285873 and was already almost merged but was changed at the last minute. If Buildifier expects `BUILD.{name}.bazel` then `cargo-raze` should be compliant. Vendored files are already compliant by default with the default `output_buildfile_suffix` being `BUILD.bazel`

edit: Linked the wrong comment. Fixed that.


PiotrSikora (approved):
Thanks!

UebelAndre:
Maybe a smoother/larger change would be to add a new field that controls the formatting of the "Remote" output files and rename `output_buildfile_suffix` to indicate that it only controls the "Vendored" output files. I feel the only benefit in doing this would be to maintain legacy behavior. I personally much prefer that these files output the standardized filenames and so am fine with having the "Remote" outputs always output `BUILD.{crate}.bazel`.

@acmcarther would probably have strong feelings about this. Hoping to hear from him.

PiotrSikora:
The `BUILD.{crate}.bazel` format was introduced quite recently (I think in `v0.5.0`), those files were previously named `{crate}.BUILD`, so I don't think there should be too much concern about maintaining legacy behavior.

UebelAndre:
@PiotrSikora see 210083ea7d25bd8f6ee56cc3efd66bfad134f0af 

Man, that was a tough one. Friendly reminder to always remember to challenge one's assumptions 😄

UebelAndre:
@acmcarther This PR is ready for review

UebelAndre:
@PiotrSikora I want to explicitly thank you for opening your PR for sorting out the other attributes. That was SUUUUPER helpful in getting us on the same page 🙏 😄

PiotrSikora:
@UebelAndre awesome! Thanks for fixing this!

PiotrSikora:
Note: This should probably ship as `v0.7.0`, due to changes in artifact filenames.

impl/src/planning.rs
```diff
@@ -229,8 +230,8 @@ impl CrateCatalogEntry {
   pub fn local_build_path(&self, settings: &RazeSettings) -> String {
     match settings.genmode {
       GenMode::Remote => format!(
-        "remote/{}.{}",
-        &self.package_ident, settings.output_buildfile_suffix,
```
acmcarther, line 233: We're removing this setting?

I'm OK with this, but maybe we should update the documentation of `output_buildfile_suffix` in documentation to mention that it doesn't do anything.

acmcarther (changes requested):
LGTM % request to update `settings.rs`.

impl/src/planning.rs
UebelAndre: It's still used in "Vendored" mode. My thought was update the variable. Your thoughts there? I do agree that something should be done though.

UebelAndre: I updated the documentation for that variable. I think it could honestly be deprecated though.

UebelAndre:
@acmcarther this is ready for another review

impl/src/planning/subplanners.rs
```diff
@@ -76,6 +76,7 @@ struct DependencySet {
   aliased_deps: Vec<DependencyAlias>,
 }
 
+/** A set of dependencies that a crate has for a specific target/cfg */
```
acmcarther, line 79: This change should probably not be part of this PR.

impl/src/context.rs
```diff
@@ -17,9 +17,9 @@ use serde::Serialize;
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize)]
 pub struct BuildableDependency {
+  pub buildable_target: String,
```
acmcarther, line 20: What's up with these changes? The don't seem relevant to the PR (and, in this case, i'm not sure why you're making them).

acmcarther (changes requested):
PR generally looks good to me, but you've got more changes here that aren't related to the PR.

impl/src/context.rs
UebelAndre: These are required to have the outputs be sorted correctly

impl/src/planning/subplanners.rs
UebelAndre: I figured it's harmless. I can remove it if you feel strongly though. What other PR would it go in otherwise?

impl/src/context.rs
UebelAndre: since Buildifier will expect the files to be sorted by `buildable_target` (the thing that gets renderd) instead of `name`. See #261 and https://github.com/google/cargo-raze/pull/260#issuecomment-710762137 for details about the issue.

impl/src/planning/subplanners.rs
acmcarther: I've made comment similar to this one elsewhere. Basically, I feel pretty strongly that PRs should be focused on whatever their title and description are focused on, and incidental changes should be moved to smaller PRs. This is most important when dealing with a release process where commits might get cherry picked or whatever (so, not really this project), but I still consider it good practice.

I'm generally lenient on this practice for junior engineers or folks who aren't familiar enough with git or mercurial (... such that they might shoot themselves in the foot trying to handle it...) but I've seen you deftly split PRs into smaller reviewable units, so I'd appreciate a clean separation here.

Just create a "clean up and document subplanner code" PR.

impl/src/context.rs
acmcarther: Oh I think I understand. To be clear, we're relying on rust sorting behavior for the resulting ordering of the generated BUILD rule. This is important to ensure that the output is BUILDIFIER compliant.

Perhaps capture this in struct documentation to prevent someone from changing this later without understanding why these fields are ordered in this way?

impl/src/planning/subplanners.rs
UebelAndre: Removed these changes and made https://github.com/google/cargo-raze/pull/264

impl/src/context.rs
UebelAndre: I added documentation for the struct and a note about buildifier compliance

impl/src/context.rs
```diff
@@ -15,11 +15,16 @@
 use crate::settings::CrateSettings;
 use serde::Serialize;
 
+/** A struct containing information about a crate's dependency that's buildable in Bazel 
+ * 
+ * Note: The ordering of the this struct's fields is important to ensure that the output
```
acmcarther, line 20: (I'm not sure if i'm being reasonable here, sorry if I'm not)

I think you should be as specific as possible here:

```Note: BUILDifier-compliant BUILD file generation depends on correct sorting of collections of this struct by `buildable_target`. Do not add fields before that field.```

(this is part of a general philosophy of comment and code writing I've acquired -- my north star is to find the perfect balance of terse and detailed. One useful heuristic is to start with the most specific explanation, and then decide whether that needs to be cut down. Typically, a good specific explanation happens also to be terse)

**Review by acmcarther — APPROVED** [review]

File: impl/src/context.rs (hunk 4 above)
Comment (UebelAndre) [reply impl/src/context.rs:20]: Yeah, this is a good suggestion. I can remember to be more specific about things like this. It's definitely something that will help everyone.

Comment (UebelAndre) [reply impl/src/context.rs:20]: Updated the comment.

File: impl/src/rendering/templates/partials/rust_library.template [hunk 5]
```diff
@@ -10,10 +9,10 @@ alias(
 )
 
 {% endif -%}
-# buildifier: leave-alone
 rust_library(
     name = "{{ target_name_sanitized }}",
-    crate_type = "{{ target.kind }}",
+{% include "templates/partials/common_attrs.template" %}
+    # buildifier: leave-alone
```
UebelAndre, line 15: @PiotrSikora For this and the `rust_binary` template, I'm not sure how to solve this. I think it'd be better to move all sorting into the templates if possible but I don't know how realistic that is. Is this something you could take a look at?

UebelAndre:
There was still one case that was not being handled. I updated the PR to hopefully pass now but it required adding more `leave-alone`s. @PiotrSikora I added a comment on this PR pointing that out.

@acmcarther Up to you if you still want to merge this. I think this is already a big improvement but don't know if that's just because I sank a bunch of time into it 😅

impl/src/rendering/templates/partials/rust_library.template
PiotrSikora: Could we merge and sort `crate.default_deps.dependencies` and `crate.raze_settings.additional_deps` into some temporary list before generating those targets?

acmcarther:
I'm open to merging this and implementing correct sorting of the merged default_deps and additional_deps in a separate PR, but I don't feel strongly about it and would defer to @PiotrSikora's preference.

impl/src/rendering/templates/partials/rust_library.template
UebelAndre: Maybe? I think it'd be nice to be able to express this within the templating but it also would make sense to have a single variable with all dependencies. Is this something you have time to try?

PiotrSikora: I know nothing about Jinja, but does something like this work?
```
{%- set deps = [] %}
{%- for dependency in crate.default_deps.dependencies %}
{%-   deps.append(dependency.buildable_target) %}
{%- endfor %}
{%- for dependency in crate.raze_settings.additional_deps %}
{%-   deps.append(dependency) %}
{%- endfor %}
{%- for dependency in deps|sort %}
    "{{dependency}}",
{%- endfor %}
```

UebelAndre: I tried something similar to that. The templating tool being used here is https://github.com/Keats/tera which has a slightly different set of rules. I ran into a bunch of syntax errors trying to figure this out and got frustrated and added back `leave-alone` 😓 

In theory something like that should work and should be super simple but for whatever reason I wasn't able to get it working right away.

PiotrSikora: @UebelAndre This works:
```
--- a/impl/src/rendering/templates/partials/rust_library.template
+++ b/impl/src/rendering/templates/partials/rust_library.template
@@ -17,10 +17,7 @@ rust_library(
         {%- if crate.build_script_target %}
         ":{{ crate_name_sanitized }}_build_script",
         {%- endif %}
-        {%- for dependency in crate.default_deps.dependencies %}
-        "{{dependency.buildable_target}}",
-        {%- endfor %}
-        {%- for dependency in crate.raze_settings.additional_deps %}
+        {%- for dependency in crate.default_deps.dependencies | map(attribute="buildable_target") | concat(with=crate.raze_settings.additional_deps) | sort %}
         "{{dependency}}",
         {%- endfor %}
     ]
```
Could you add this and remove `# buildifier: leave-alone`?

But I don't think we have any example that actually uses both.

UebelAndre: There's still one defect
```diff
diff --git a/examples/vendored/complicated_cargo_library/cargo/vendor/regex-0.2.11/BUILD.bazel b/examples/vendored/complicated_cargo_library/cargo/vendor/regex-0.2.11/BUILD.bazel
index ba728e02..a5c72393 100644
--- a/examples/vendored/complicated_cargo_library/cargo/vendor/regex-0.2.11/BUILD.bazel
+++ b/examples/vendored/complicated_cargo_library/cargo/vendor/regex-0.2.11/BUILD.bazel
@@ -66,12 +66,12 @@ rust_library(
     ],
     version = "0.2.11",
     deps = [
+        "//vendored/complicated_cargo_library/cargo:specs",
         "//vendored/complicated_cargo_library/cargo/vendor/aho-corasick-0.6.10:aho_corasick",
         "//vendored/complicated_cargo_library/cargo/vendor/memchr-2.3.3:memchr",
         "//vendored/complicated_cargo_library/cargo/vendor/regex-syntax-0.5.6:regex_syntax",
         "//vendored/complicated_cargo_library/cargo/vendor/thread_local-0.3.6:thread_local",
         "//vendored/complicated_cargo_library/cargo/vendor/utf8-ranges-1.0.4:utf8_ranges",
-        "//vendored/complicated_cargo_library/cargo:specs",
     ],
 )

```

PiotrSikora: OK, I think we need to keep `# buildifier: leave-alone` for this one. `/` precedes `:` in the ASCII table, so the output is now sorted correctly, but it doesn't match `buildifier`'s sorting, which seems to be aware of the packages' hierarchy. Sadly, this means that we won't be able to solve this using templates.

Please re-add `# buildifier: leave-alone` above `deps` and this should be good to merge. Thanks!

UebelAndre: I did my best 😅

maybe in the future we can have the tool check to see if the buildifier binary is available and run that on the generated files.

I dropped the last two commits which is effectively reapplying `leave-alone`.

UebelAndre:
@PiotrSikora @acmcarther How does the change look now?

PiotrSikora (approved):
@UebelAndre looks great, thanks!

PR #265, opened 2020-10-21
Always clean the "remote" output directory

UebelAndre:
When using `cargo-raze` with `genmode = "Remote"` I always have to either hand inspect the verbose output or delete the `remote` directory in order to find out what files are actually relevant to my build now. This change makes it so the `remote` output directory is always cleaned before rendering new build files to it.


UebelAndre:
@acmcarther if you still have bandwidth today, I'd love to have a review and a release for this change.

impl/src/bin/cargo-raze.rs
```diff
@@ -120,8 +120,12 @@ fn main() -> Result<()> {
     GenMode::Vendored => bazel_renderer.render_planned_build(&render_details, &planned_build)?,
     GenMode::Remote => {
       if !dry_run {
-        // Create the "remote" directory if it doesn't exist
-        fs::create_dir_all(render_details.path_prefix.as_path().join("remote"))?;
+        // Clean out the "remote" directory and guarantee that it exists
+        let remote_dir = render_details.path_prefix.as_path().join("remote");
+        if remote_dir.exists() {
+          fs::remove_dir_all(&remote_dir)?;
```
PiotrSikora, line 126: I agree that this would lead to a better user experience.

However, I imagine that some users might add files that were not generated by `cargo-raze` to that directory. Would it make sense to remove only files matching `*BUILD*` pattern?

UebelAndre: What kind of things do you think users might put in that directory?

PiotrSikora: I don't have good answer for that, but running `rm -rf /some/directory` is something that I try to avoid if possible, and there is a clear pattern that we can match against here.

I don't feel too strongly about this change, though, so feel free to leave it as-is.

UebelAndre: I leave this up to @acmcarther

acmcarther: I second this. Delete only BUILD files matching a regex?.

("i was too lazy to X" and "rm -rf" is a humongous red flag, btw)

acmcarther (changes requested):
Just seconding PiotrSikora's comment.

impl/src/bin/cargo-raze.rs
UebelAndre: Doing the simpler thing is not a bad thing (no matter the motivation). I'll try to get this addressed later today though 😀

UebelAndre:
After some reflection I've realized I don't actually know what the impact of any change to this crate is. Would it be possible to put together a list of users? I think it'd greatly help me and future contributors in this regard. 🙏

acmcarther:
I don't know who uses this crate either, and I suspect you'd be hard pressed to generate such a list. I doubt crates.io maintains any metadata.

The impact of a given change can be estimated without knowing the end users though. My suggested best practice is to assume users are doing the hackiest, most unsound thing while also assuming that they're trying to use the dependency as a critical component in their space program, and while assuming that they don't follow PR discussions. That is to say, break things only after a huge amount of lead time, avoid breaking things when an alternative exists, and generally, don't assume that people read the directions.

acmcarther:
Do you think there might be a chance that users might be maintaining hand-written build files in this directory? If so, they might be very upset by us deleting them. Maybe over-cautious, but perhaps we should put this behind a config setting?

I don't feel really strongly about this and I'm open to arguments which indicate that no user would actually do that.

UebelAndre:
I'd really like to know about that. I would love to make sure raze is expressive enough to be able to properly generate the necessary output file. Can you think of some cases (maybe take some time and post some thoughts tomorrow. I'll be doing the same)?

UebelAndre:
After some thought, I feel like users should have the ability to generate whatever they need through raze with `additional_build_file` and other such settings unless there's some bug they're working around. That said though, I can add a `--clean` argument to gate this behavior. Sound reasonable?

dfreese:
> After some thought, I feel like users should have the ability to generate whatever they need through raze with `additional_build_file` and other such settings unless there's some bug they're working around. That said though, I can add a `--clean` argument to gate this behavior. Sound reasonable?

From my end that sounds reasonable and reflects my teams usage of raze, which is in a shell script that deletes the remote directory first.

UebelAndre:
> > After some thought, I feel like users should have the ability to generate whatever they need through raze with `additional_build_file` and other such settings unless there's some bug they're working around. That said though, I can add a `--clean` argument to gate this behavior. Sound reasonable?
> 
> 
> 
> From my end that sounds reasonable and reflects my teams usage of raze, which is in a shell script that deletes the remote directory first.

I have the same thing setup and so do others I've talked to. If users are writing similar shell scripts to use `cargo-raze` then the tool itself should likely be updated to include the common functionality.

acmcarther:
Above discussion is convincing enough for me to LGTM deleting BUILD files from remote.

acmcarther (approved):

PR #266, opened 2020-10-22
Updated dependencies

UebelAndre:
`cfg-expr` added support for rust 1.47.0 targets


UebelAndre:
@dfreese any chance you could look at this one as well?

dfreese:
> `cfg-expr` added support for rust 1.47.0 targets

Does that require a bump in the MSV of cargo-raze?  I don't think we have that listed anywhere, and CI runs against 1.47.0, so that doesn't help.

UebelAndre:
> Does that require a bump in the MSV of cargo-raze? I don't think we have that listed anywhere, and CI runs against 1.47.0, so that doesn't help.

No, there should be no need to change that. This only adds support for platforms introduced in 1.47.0 but the crate does not require that to compile.

```output
cargo +1.45.1 test -q

running 47 tests
...............................................
test result: ok. 47 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out


running 1 test
.
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

dfreese (approved):

PR #267, opened 2020-10-23
Checksums for binary dependencies shouldn't be fetched in genmode=vendored

UebelAndre:
`Vendored` genmode relies on the behavior of `cargo vendor` and shouldn't do anything special for "binary dependencies".

### DIFF
--- a/impl/src/planning/subplanners.rs
+++ b/impl/src/planning/subplanners.rs
@@ -244,13 +244,17 @@ impl<'planner> WorkspaceSubplanner<'planner> {
       }
     }
 
-    // Additionally, the binary dependencies need to have their checksums added as well
-    for (pkg, info) in self.settings.binary_deps.iter() {
-      let version = semver::Version::parse(info.req())?;
-      package_to_checksum.insert(
-        (pkg.clone(), version.clone()),
-        fetch_crate_checksum(&self.settings.index_url, pkg, &version.to_string())?,
-      );
+    // Additionally, the binary dependencies need to have their checksums added as well in 
+    // Remote GenMode configurations. Vendored GenMode relies on the behavior of `cargo vendor`
+    // and doesn't perform any special logic to fetch binary dependency crates.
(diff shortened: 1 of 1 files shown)


UebelAndre:
@acmcarther Hey, just FYI, this is part of a larger set of changes I'm working on and would love to get these in before the weekend so I can hopefully have them done and in a pull-request on monday 😅 🙏

impl/src/planning/subplanners.rs
```diff
@@ -245,12 +245,15 @@ impl<'planner> WorkspaceSubplanner<'planner> {
...
-    for (pkg, info) in self.settings.binary_deps.iter() {
-      let version = semver::Version::parse(info.req())?;
-      package_to_checksum.insert(
-        (pkg.clone(), version.clone()),
-        fetch_crate_checksum(&self.settings.index_url, pkg, &version.to_string())?,
-      );
+    if self.settings.genmode == GenMode::Remote {
+      for (pkg, info) in self.settings.binary_deps.iter() {
+        let version = semver::Version::parse(info.req())?;
+        let key = fetch_crate_checksum(&self.settings.index_url, pkg, &version.to_string()).unwrap();
```
dfreese, line 251: Was there a reason `?` was changed to `unwrap()`?

impl/src/planning/subplanners.rs
```diff
@@ -245,12 +245,15 @@ impl<'planner> WorkspaceSubplanner<'planner> {
     }
 
     // Additionally, the binary dependencies need to have their checksums added as well
```
dfreese, line 247: can you update the comment to be specific to remote, and mention why vendored shouldn't do this?

impl/src/planning/subplanners.rs
UebelAndre: This was leftover from debugging. Sorry about that. Unintended change

dfreese (approved):

PR #268, opened 2020-10-24
settings.index_url can now be a path

UebelAndre:
This is primarily to unblock other work I'm doing but it did reveal that `test_binary_dependencies_remote_genmode` was a test that relied on remote resources. This both makes the test more deterministic and much faster.


impl/src/metadata.rs
```diff
@@ -194,12 +194,20 @@ fn cargo_generate_lockfile(
 }
 
 pub fn fetch_crate_checksum(index_url: &str, name: &str, version: &str) -> Result<String> {
-  let index = crates_index::BareIndex::from_url(index_url)?;
-
-  let crate_index = index
-    .open_or_clone()?
-    .crate_(name)
-    .ok_or(anyhow!("Failed to find crate '{}' in index", name))?;
+  let crate_index = match url::Url::parse(index_url) {
```
acmcarther, line 197: This could be regular conditional if you don't actually need the contents of the result.

```let index_path_is_url = url::Url::parse(index_url).is_ok()```

My rule is to use the absolute least powerful expression type that works for the job.

File: impl/src/testing.rs [hunk 2]
```diff
@@ -24,6 +41,29 @@ dependencies = [
     "
 }
 
+pub fn named_toml(name: &str, version: &str) -> String {
```
acmcarther, line 44: Would prefer slightly more verbose names for these two.

"named_toml_contents" "named_lock_contents".

impl/src/testing.rs
```diff
@@ -56,3 +96,107 @@ pub fn make_workspace(
...
+      .header("content-type", "application/x-tar")
+      .body_from_file(&tar_path.display().to_string());
+  });
+
+  dir
+}
+
+/** A helper macro for passing a `crates` to  `mock_crate_index` */
+#[macro_export]
+macro_rules! index_crates {
```
acmcarther, line 160: This can be a function, right?

(again, use the _least powerful_ available tool for the job)

impl/src/testing.rs
```diff
@@ -56,3 +96,107 @@ pub fn make_workspace(
 pub fn make_basic_workspace() -> (TempDir, CargoWorkspaceFiles) {
   make_workspace(basic_toml(), Some(basic_lock()))
 }
+
+/** Creates mock responses for fetching crate metadata and source code from a crate registry */
```
acmcarther, line 100: Nit: More precise would be "Configures the given mock_server (representing a crates.io remote) to return mock responses for the given crate and version"

impl/src/testing.rs
```diff
@@ -56,3 +96,107 @@ pub fn make_workspace(
...
+pub fn mock_remote_crate(name: &str, version: &str, mock_server: &MockServer) -> TempDir {
+  // Crate info mock response
+  mock_server.mock(|when, then| {
+    when.method(GET).path(format!("/api/v1/crates/{}", name));
+    then.status(200).json_body(json!({
+        "crate": {
+            "id": name,
+            "name": name,
+            "versions": [
+                283716
```
acmcarther, line 110: What is this and where does it come from?

impl/src/testing.rs
```diff
@@ -56,3 +96,107 @@ pub fn make_workspace(
...
+}
+
+/** Create a mock cache in a temporary direcotry that contains a set of given crates */
+pub fn mock_crate_index(crates: &HashMap<String, String>) -> TempDir {
+  let index_url = TempDir::new().unwrap();
+
+  for (name, version) in crates {
+    assert!(
+      name.len() >= 5,
+      "Mock crate names should be at least 5 characters"
```
acmcarther, line 176: Why?

impl/src/testing.rs
```diff
@@ -56,3 +96,107 @@ pub fn make_workspace(
...
+    );
+    let path = index_url
```
acmcarther, line 178: Path to what?

impl/src/testing.rs
```diff
@@ -56,3 +96,107 @@ pub fn make_workspace(
...
+        "vers": version,
+        "deps": [],
+        "cksum": "8a648e87a02fa31d9d9a3b7c76dbfee469402fbb4af3ae98b36c099d8a82bb18",
+        "features": {},
+        "yanked": false,
+        "links": null
+      })
+      .to_string(),
+    )
+    .unwrap();
```
acmcarther, line 198: I'm seeing a whole lot of unwrap here. Maybe this function should just return a result.

I realize this is for testing, but it still feels sloppy. Maybe the caller can just unwrap.

impl/src/planning.rs
```diff
@@ -135,6 +135,7 @@ mod tests {
 
   use super::*;
   use cargo_metadata::PackageId;
+  use httpmock::MockServer;
```
acmcarther, line 138: We already had this dependency? I'm surprised to hear that.

acmcarther (changes requested):
Almost entirely nits.

I'm still grumpy about these pattern matches.

impl/src/planning.rs
UebelAndre: Why is that surprising?

impl/src/metadata.rs
UebelAndre: I feel like this is actually a valid use for `match` How else can I conditionally set `crate_index` so I don't have to duplicate the logic at the end of this function? Or is that the wrong problem to solve here?

(I'm not deliberately trying to use `match` everywhere, but it does seem to allow me to express more of what I want. I know your gripe with it and am happy to employ alternatives, just need to learn what those are 😅)

impl/src/testing.rs
UebelAndre: it came from the actual data of `cargo-raze` version `0.5.0`. This value simply needs to match the `id` in `versions`.

impl/src/testing.rs
UebelAndre: Because I hadn't implemented support for characters less than that length in the mock. I always parse the index path of two pairs of characters below this. so `some-crate` would be locatable  at `so/me/some-crate`

impl/src/testing.rs
UebelAndre: This would require require a conversion from `io::Error` to likely `anyhow::Error` which seemed unnecessary. Can you elaborate on why unwrap in this case is also sloppier than `?`?

UebelAndre: After some sleep I realized I was doing something silly 😅 

Updated. But still curious as to why `?` would be less sloppy in testing. I find it better to know exactly what and where something broke with virtually no digging required.

impl/src/metadata.rs
acmcarther:
```rust
let index_path_is_url = url::Url::parse(index_url).is_ok()
let crate_index_path = if index_path_is_url {
  crates_index::Index::new(index_url)
    .crate_(name)
    .ok_or(anyhow!("Failed to find crate '{}' in index", name))?
} else {
  crates_index::BareIndex::from_url(index_url)?
    .open_or_clone()?
    .crate_(name)
    .ok_or(anyhow!("Failed to find crate '{}' in index", name))?
};
```

I don't really feel strongly about this, but it's equivalent.

impl/src/testing.rs
acmcarther: Add a comment? I don't feel strongly about this.

impl/src/metadata.rs
UebelAndre: Oh wow, I didn't know you can do that with an `if`. So in every case would you prefer that style over using `match`?

impl/src/testing.rs
acmcarther: Looks like what the real repo does is:
1. Single char names go in 1/
2. Double char names go in 2/
3. Three char names go in 3/

Open to replicating this here?
```rust
// Returns the subdirectory within the index for the given crate name. 
fn mock_index_subdir(crate_name str&) -> PathBuf {
  if (crate_name.len() > 4) {
    let mut p = PathBuf::from(format!("{}/",name.as_str()[0..2]));
    p.push(format!("{}/", name.as_str()[2..4]));
    return p;
  }
  return PathBuf::from("{}/", crate_name.len());
}
```

UebelAndre: Can do, I'll take care of this tonight

impl/src/metadata.rs
acmcarther: I prefer `if` when it suffices. If the match is "doing work", meaning it's serving both as control flow and as destructuring (e.g. Ok(value_i_need)), then I prefer the match. I also prefer the match for enums, or when there's more than two values, unless an early return or series of early returns can be used. Then I prefer the early returns.

impl/src/planning.rs
acmcarther: dunno, I guess I'm just surprised that we have a mock http server dependency. Many tests could be implemented using it to expand test coverage.

I guess you added it in https://github.com/google/cargo-raze/pull/227

¯\\\_(ツ)\_/¯

UebelAndre:
@acmcarther this is ready for review.

impl/src/testing.rs
UebelAndre: Updated

impl/src/testing.rs
UebelAndre: Updated

impl/src/metadata.rs
UebelAndre: Updated

impl/src/testing.rs
UebelAndre: This ended up getting rolled back to contain unwraps but I think this is better addressed in https://github.com/google/cargo-raze/pull/276

impl/src/testing.rs
```diff
@@ -24,6 +41,31 @@ dependencies = [
     "
 }
 
+pub fn named_toml_contents(name: &str, version: &str) -> String {
+  vec![
```
GregBowyer, line 45: Nit: [indoc](https://docs.rs/indoc/1.0.3/indoc/) can make this better with dtolnay awesome.

```rust
indoc::formatdoc! { r#"
    [package]
    name =  {name}
    version = {version}

    [lib]
    path = "not_a_lib.rs"
"#, name = name, version=version }
```

impl/src/testing.rs
```diff
@@ -56,3 +98,119 @@ pub fn make_workspace(
...
+            }
+        ],
+    }));
+  });
+
+  // Create archive contents
+  let dir = tempfile::TempDir::new().unwrap();
+  let tar_path = dir.as_ref().join(format!("{}.tar.gz", name));
+  {
+    std::fs::create_dir_all(dir.as_ref().join("archive")).unwrap();
```
GregBowyer, line 134: Nit: Maybe import `std::fs` and `std::fs::File` ?

GregBowyer (approved):

impl/src/testing.rs
UebelAndre: updated

impl/src/testing.rs
UebelAndre: updated

acmcarther (approved):
LGTM

PR #272, opened 2020-10-30
[Lint] Change inline test texts to use `indoc!`

GregBowyer:
This is a _small_ quality of life change. Make it possible to use `indoc!` to declare inline strings.

This is going to be used to make tests for #127 since the inline toml is large and has a few escapes that can get hairy.


dfreese (approved):
Seems generally like a readability win.

PR #274, opened 2020-10-31
Accept `[package.metadata.raze]`, superseding `[raze]`

### LINKED ISSUE #273: cargo: unused manifest key; use package metadata?
When I extend my `Cargo.toml` with

```toml
[raze]
workspace_path = "//third_party/rust"
genmode = "Remote"
```

per the documentation, `cargo raze` works great! But all `cargo`
commands also now print a warning about the `raze` key:

```
$ cargo check
warning: unused manifest key: raze
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
```

[The Cargo docs mention that “tools which would like to store package
configuration in `Cargo.toml`” can use the `package.metadata` table to
ignore that warning.][d] I think that this would just amount to changing
the table from `[raze]` to `[package.metadata.raze]`:

```toml
[package.metadata.raze]
workspace_path = "//third_party/rust"
genmode = "Remote"
```

When I do so, Cargo indeed no longer complains, but now `cargo-raze`
can’t find my configuration:

```
$ cargo raze
Error: Raze failed with cause: "missing field `raze` at line 20 column 1"
```

Perhaps `cargo-raze` could be modified to accept `package.metadata.raze`
as well as `raze`, and then maybe to drop support for bare `raze` at
such a time when that breaking change would be considered acceptable?

[d]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table

wchargin:
Cargo sets aside the [`[package.metadata]`][md] namespace for use by
third-party tools like `cargo-raze`. By specifying configuration under
`[package.metadata.raze]` rather than `[raze]`, we avoid an “unused
manifest key: raze” warning on every Cargo command.

This patch teaches `cargo-raze` to accept either `[raze]` or the new
`[package.metadata.raze]`. In case of conflict, it prints a warning and
uses the value of `[package.metadata.raze]`. In the future, we can
remove the `[raze]` key entirely if desired.

Fixes #273.

Changes outside of `impl/` generated by:

```
git ls-files -z ':!impl' | xargs -0 sed -i 's/\[raze/[package.metadata.raze/'
```

[md]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table

Test Plan:
With this patch, running `./smoke-test.sh` passes both before and after
updating all the examples with the above `sed` command. Before the `sed`
update, the smoke test prints the deprecation warning, as desired.

wchargin-branch: toml-package-metadata-raze

### DIFF
--- a/impl/src/settings.rs
+++ b/impl/src/settings.rs
@@ -26,7 +26,18 @@ pub type CrateSettingsPerVersion = HashMap<Version, CrateSettings>;
  */
 #[derive(Debug, Clone, Deserialize)]
 pub struct CargoToml {
-  pub raze: RazeSettings,
+  pub raze: Option<RazeSettings>,
+  pub package: Option<PackageToml>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct PackageToml {
+  pub metadata: Option<MetadataToml>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct MetadataToml {
+  pub raze: Option<RazeSettings>,
 }
 
 /** The configuration settings for `cargo-raze`, included in Cargo.toml. */
@@ -361,7 +372,7 @@ pub fn load_settings<T: AsRef<Path>>(cargo_toml_path: T) -> Result<RazeSettings,
     Ok(handle) => handle,
     Err(err) => {
       return Err(RazeError::Generic(err.to_string()));
-    },
+    }
   };
 
   let mut toml_contents = String::new();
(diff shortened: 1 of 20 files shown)


impl/src/settings.rs
```diff
@@ -370,7 +381,35 @@ pub fn load_settings<T: AsRef<Path>>(cargo_toml_path: T) -> Result<RazeSettings,
     return Err(RazeError::Generic(err.to_string()));
   }
   let mut settings = match toml::from_str::<CargoToml>(&toml_contents) {
```
UebelAndre, line 383: Do you know if `package.metadata.raze` settings can be parsed from `cargo metadata`?

Specifically, could something like the `CargoMetadataFetcher` be used to load these settings more holistically?

wchargin: Indeed, they can. It looks like this:

```shell
$ cargo metadata --format-version 1 | jq '.. | .raze? | select(.)'
{
  "genmode": "Remote",
  "workspace_path": "//third_party/rust",
  "crates": {
    "crc": {
      "1.8.1": {
        "gen_buildrs": true
      }
    },
    "indexmap": {
      "1.6.0": {
        "additional_flags": [
          "--cfg=has_std"
        ]
      }
    },
    "proc-macro2": {
      "1.0.24": {
        "additional_flags": [
          "--cfg=use_proc_macro",
          "--cfg=wrap_proc_macro"
        ]
      }
    },
    "prost-build": {
      "0.6.1": {
        "gen_buildrs": true
      }
    },
    "thiserror": {
      "1.0.21": {
        "gen_buildrs": true
      }
    }
  }
}
```

The path is `.packages[].metadata.raze`, so you’d have to pluck it from
the right package.

I haven’t looked at `CargoMetadataFetcher`. If you’d prefer that
approach, I can do so (but no promises on when I’ll get around to it).

Note that `[raze]` does not show up in `cargo metadata`, so as long as
you want to keep both around you would need two parsing paths. So
I might propose a plan of (a) parse both from `Cargo.toml`, (b) wait
until comfortable dropping compat, (c) remove `CargoToml::raze`,
(d) migrate to parsing from `cargo metadata` now that that’s easier.

UebelAndre: I was more trying to find out if this information could be parsed all from `cargo metadata`.

wchargin: Looks like the `[package.metadata.raze]` info can; the `[raze]` info can’t.

acmcarther (approved):
PR description makes a compelling case for this change, and this PR keeps the original codepath around so that we don't suddenly break people.

Thanks for making this change, it LGTM.

UebelAndre:
@acmcarther could you do a release for this change?

wchargin:
Thanks for the reviews!

dfreese:
> @acmcarther could you do a release for this change?

It would be nice to land https://github.com/google/cargo-raze/pull/127, then I'd be fine releasing 0.7.0 to include this.

dfreese:
Released in 0.7.0

Issue #262, opened 2020-10-18
Feature Request: Support Cargo Workspaces


UebelAndre:
I decided to make a top level issue to track all information about [cargo workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html). This tool should be updated to be able to generate Bazel files for projects that make use of cargo workspaces.

UebelAndre:
If anyone was interested in implementing this I'd be forever grateful!!

acmcarther:
Do you have opinions about the desired behavior? Do you think we should represent the workspaces somehow in the generated output, or just generate "the minimal thing that works"?

UebelAndre:
I would be happy with "the minimal thing that works". I do think the best approach for the time being would be to create aliases in the appropriate subpackages that contain aliases to the dependencies, but outside of that, I don't think it should matter how the dependencies are rendered.

My thought is that the outputs would look something like the following:

```output
├── BUILD.bazel
├── Cargo.lock
├── Cargo.toml
├── WORKSPACE.bazel
├── cargo
│   ├── BUILD.bazel
│   └── remote
│       └── BUILD.deps.bazel
├── lib_a
│   ├── BUILD.bazel
│   ├── Cargo.toml
│   ├── cargo
│   │   └── BUILD.bazel
│   └── src
│       └── lib.rs
└── lib_b
    ├── BUILD.bazel
    ├── Cargo.toml
    ├── cargo
    │   └── BUILD.bazel
    └── src
        └── lib.rs
```

where `./cargo` is where all build files are generated and the directories `./lib_a/cargo` and ./lib_b/cargo` would have generated aliases for their correct dependencies.

That said though, I do feel at some point `raze` should interact with `bazel-gazelle` at some point such that users no longer have to manage dependencies in two places (their BUILD files and the `Cargo.toml` files). Without giving it too much though, I feel the source of truth should remain the `Cargo.toml` files and I like this because it makes it easier for existing Rust projects to become integrated with Bazel (though there may be better solutions for this).

So long as I can run `cargo-raze` on a project that's using cargo workspaces and have all my dependencies correctly generated and a way to relatively easily hook them up, I'll be thrilled.

UebelAndre:
I would say if I got to pick the next feature `cargo-raze` got, it would definitely be this one.

nikclayton-dfinity:
FWIW, I've been trying another approach with a project that uses workspaces.

Orginally, the directory tree looked like (just the Rust parts)

```
[...]

PR #276, opened 2020-11-01
Workspace support

UebelAndre:
The following pull requests are included in this change set and should be closed first:
- https://github.com/google/cargo-raze/pull/275
- https://github.com/google/cargo-raze/pull/268

This pull requests adds support for [cargo workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) and closes https://github.com/google/cargo-raze/issues/262

### Notable Changes:
1. There is now a new setting `workspace_member_dir` which is used to determine the location where a package's dependencies will get aliased. Previously, a BUILD file would be rendered into `workspace_path` which would contain aliases for direct dependencies. Because we now account for there being multiple packages (ie, in a workspace) a directory matching the name specified by `workspace_member_dir` will be created in each package directory (where a `Cargo.toml` file lives) containing a BUILD file with the aliases. This is intended to make it easier for users to specify the dependencies they care about as well as maintain the previous behavior (`workspace_member_dir` should simply match the package name specified by `workspace_path` to result in no change for existing projects).
2. Binary dependencies are treated as workspace members. This is done by downloading the binary dependency's source into the mock workspace when `raze` is gathering metadata and modifying the workspace root's `Cargo.toml` to include a [workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html) definition that includes the binary dependency. This allows `raze` to share a single lockfile for all dependencies and ensure all dependencies use compatible versions with each other. (this actually solves a bug I noticed when generating `wasm-bindgen-cli` for `rules_rust`. [This dependency](https://github.com/bazelbuild/rules_rust/blob/master/wasm_bindgen/raze/Cargo.lock#L37-L44) will update if the file is regenerated and cause a build error). `Cargo-raze` will now output a `Cargo.raze.lock` file when binary dependencies are specified. This file is authoritative over subsequent raze runs.
3. Add support for defining Raze Settings via `[workspace.metadata.raze]`. This will take precedence over package definitions (`[package.metadata.raze]`) if it's defined. This should provide a consistent central location for settings even in projects that only provide a single root package.

Sorry about the large PR. I did a pass to split out what I could into other pull requests but I found it difficult to do so.


UebelAndre:
@acmcarther sorry about the large PR. Looking forward to your feedback. I'm really hoping you like some of the changes here 😄

UebelAndre:
@wchargin do you have any suggestion on how `[package.metadata.raze]` would work in the context of a workspace? I'm getting warnings output in #277

UebelAndre:
> @wchargin do you have any suggestion on how `[package.metadata.raze]` would work in the context of a workspace? I'm getting warnings output in #277

Ah, found https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspacemetadata-table

Sorry to bother 😅

impl/src/metadata.rs
```diff
@@ -50,25 +97,132 @@ pub struct CargoWorkspaceFiles {
...
+        no_deps_metadata.workspace_root.join("Cargo.lock"),
+        temp_dir.as_ref().join("Cargo.lock"),
+      )?;
+    }
+
+    // Copy over the Cargo.toml files of each workspace member
+    let re = Regex::new(r".+\(path\+file://(.+)\)")?;
+    for member in no_deps_metadata.workspace_members.iter() {
+      // Get a path to the workspace member directory
+      let path = {
```
acmcarther, line 161: More descriptive name?

impl/src/metadata.rs
```diff
@@ -50,25 +97,132 @@ pub struct CargoWorkspaceFiles {
...
+          None => continue,
+        };
+
+        PathBuf::from(re_match.as_str())
+      };
+
+      let toml_path = path.join("Cargo.toml");
+
+      // If no Cargo.toml file is found, skip this
+      if !toml_path.exists() {
```
acmcarther, line 178: What does it mean if there's no toml path? Should we log? Is it user error? Is it sometimes expected?

impl/src/metadata.rs
```diff
@@ -50,25 +97,132 @@ pub struct CargoWorkspaceFiles {
...
+
+              let new_path = temp_dir.as_ref().join(diff);
+
+              if let Some(parent) = new_path.parent() {
+                fs::create_dir_all(parent)?;
+              }
+
+              self.make_symlink(&path, &new_path)?;
+            },
+            _ => {},
```
acmcarther, line 209: What situations are caught here? when does this entry not contain Ok?

impl/src/metadata.rs
```diff
@@ -50,25 +97,132 @@ pub struct CargoWorkspaceFiles {
...
+  // Create a symlink file on unix systems
+  #[cfg(target_family = "unix")]
+  fn make_symlink(&self, src: &Path, dest: &Path) -> Result<()> {
+    std::os::unix::fs::symlink(src, dest)
+      .with_context(|| "Failed to create symlink for generating metadata")
+  }
+
+  // Create a symlink file on windows systems
+  #[cfg(target_family = "windows")]
+  fn make_symlink(&self, src: &Path, dest: &Path) -> Result<()> {
```
acmcarther, line 126: Thanks for handling this for all platforms!

impl/src/metadata.rs
```diff
@@ -50,25 +97,132 @@ pub struct CargoWorkspaceFiles {
...
+
+    Ok((temp_dir, no_deps_metadata.workspace_root))
+  }
+
+  fn fetch_crate_src(&self, dir: &Path, name: &str, version: &str) -> Result<PathBuf> {
+    let registry_url = {
+      let mut r_url = self.registry_url.clone();
+      r_url.set_path("");
+      r_url.to_string()
+    };
```
acmcarther, line 223: Comment above this block? I found the `set_path("")` suspicious.

impl/src/metadata.rs
```diff
@@ -92,189 +246,203 @@ impl CargoMetadataFetcher {
     Ok(crate_dir)
   }
 
-  pub fn fetch_crate_src(
+  fn inject_binaries_into_workspace(
```
acmcarther, line 249: I think this needs a comment -- the name is provocative but not really self explanatory.

impl/src/metadata.rs
```diff
@@ -92,189 +246,203 @@ impl CargoMetadataFetcher {
...
+    let value = toml::Value::try_from(&manifest)?;
+    match std::fs::write(root_toml, toml::to_string(&value)?) {
+      Ok(_) => Ok(()),
+      Err(err) => Err(anyhow!("{:?}", err)),
+    }
+  }
+
+  fn fetch_crate_checksum(&self, name: &str, version: &str) -> Result<String> {
+    let index_path_is_url = url::Url::parse(&self.index_url).is_ok();
+    let crate_index_path = if index_path_is_url {
```
acmcarther, line 281: I think this is the second time this review I"m seeing this code block.

Two things.

1. Should this be moved into a function?
2. Should we just be opening this index in `new` or something and holding onto it?

acmcarther (changes requested):
I got tired half way through the review  (on me, not you, didn't get enough sleep last night), so I'm only half way done. I'll need to continue from about half way through impl/src/metadata.rs on my next review.

Looks good so far. Thank you for your continued work here and throughout the repo!

impl/Cargo.toml
```diff
@@ -34,6 +34,8 @@ docopt = "1.1.0"
 glob = "0.3.0"
 itertools = "0.9.0"
 log = "0.4.11"
+pathdiff = "0.2.0"
```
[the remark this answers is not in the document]
UebelAndre: There's very likely an smarter implementation that I didn't think of that avoids this dependency all together. I have the same concern but felt for the time being, it'd be better to use a tested crate than hand write something.

impl/src/bin/cargo-raze.rs
```diff
@@ -65,20 +66,26 @@ Options:
 "#;
 
 fn main() -> Result<()> {
+  // Parse options ---
```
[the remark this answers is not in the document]
UebelAndre: Was just trying to highlight the "sections" within this tool. Any suggestions? (I can still remove the `---`)

impl/src/metadata.rs
UebelAndre: Where is the second instance of this code block?

impl/src/metadata.rs
UebelAndre: https://docs.rs/glob/0.3.0/glob/ suggests there would be errors in some cases. I opted to ignore them but could escalate that error. Your thoughts?

UebelAndre:
@acmcarther Hey, I've addressed some stuff here and had questions on a few things but is otherwise ready for another pass 😀

impl/src/metadata.rs
UebelAndre: Updated implementation

impl/src/rendering/bazel.rs
```diff
@@ -220,18 +219,37 @@ impl BuildRenderer for BazelRenderer {
...
-      .render_aliases(&workspace_context, &crate_contexts)
-      .map_err(|e| RazeError::Rendering {
-        crate_name_opt: None,
-        message: unwind_tera_error!(e),
-      })?;
+    for member_path in planned_build.workspace_context.workspace_members.iter() {
+      let all_packages: Vec<CrateContext> = planned_build
+        .crate_contexts
+        .iter()
+        .filter_map(|ctx| {
```
acmcarther, line 226: I think this can be expressed with filter:  ctx.is_binary_dependency || ctx.workspace_member_dependents.contains(member_path)

impl/src/rendering/bazel.rs
```diff
@@ -275,18 +294,36 @@ impl BuildRenderer for BazelRenderer {
...
 
-    let alias_file_path = path_prefix.as_path().join(buildfile_suffix);
-    let rendered_alias_build_file = self
-      .render_remote_aliases(&workspace_context, &crate_contexts)
-      .map_err(|e| RazeError::Rendering {
-        crate_name_opt: None,
-        message: unwind_tera_error!(e),
-      })?;
+    for member_path in planned_build.workspace_context.workspace_members.iter() {
+      let all_packages: Vec<CrateContext> = planned_build
```
acmcarther, line 298: This block seems to duplicate the code above.

impl/src/planning/subplanners.rs
```diff
@@ -232,70 +231,19 @@ impl<'planner> WorkspaceSubplanner<'planner> {
...
+        .crate_catalog
+        .metadata
+        .resolve
+        .as_ref()
+        .ok_or_else(|| RazeError::Generic("Missing resolve graph".into()))?
+        .nodes
+        .iter()
+        .sorted_by_key(|n| &n.id)
+        .filter_map(|node| self.create_crate_context(node, &self.crate_catalog))
+        .collect::<Result<Vec<CrateContext>>>()?,
```
acmcarther, line 245: I'd prefer more imperative framing of this. I don't think the fluent writing of this is is very readable.

impl/src/metadata.rs
```diff
@@ -50,25 +97,132 @@ pub struct CargoWorkspaceFiles {
...
+      let path = {
+        let capture = match re.captures(&member.repr) {
+          Some(capture) => capture,
+          None => continue,
+        };
+
+        let re_match = match capture.get(1) {
+          Some(re_match) => re_match,
+          None => continue,
+        };
```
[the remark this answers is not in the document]
acmcarther: Why not my version?

impl/src/bin/cargo-raze.rs
```diff
@@ -89,77 +96,81 @@ fn main() -> Result<()> {
     lock_path_opt,
   };
 
+  let baze_root = find_bazel_workspace_root().unwrap_or(std::env::current_dir()?);
```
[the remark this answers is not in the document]
acmcarther: why was this marked resolved.

acmcarther (commented):
This is going to take a very long time to review.

impl/src/bin/cargo-raze.rs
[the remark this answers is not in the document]
UebelAndre: I changed it but must have stomped the changes out when rebasing or some thing. I'll update this again

UebelAndre:
> This is going to take a very long time to review.

Why is that? Is there something I can do to help with that?

impl/src/bin/cargo-raze.rs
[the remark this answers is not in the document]
UebelAndre: I don't know why it doesn't show up in this diff but I added a comment above it explaining what the variable is. Is that not sufficient?

impl/src/settings.rs
```diff
@@ -384,17 +402,38 @@ pub fn load_settings<T: AsRef<Path>>(cargo_toml_path: T) -> Result<RazeSettings,
...
+    Ok(CargoToml {
+      workspace:
+        Some(PackageToml {
+          metadata: Some(MetadataToml {
+            raze: Some(raze),
+          }),
+        }),
+      raze: top_level_raze,
+      ..
+    }) => {
```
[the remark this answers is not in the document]
UebelAndre: The logic for loading settings is something I'm hoping to improve in a future PR. I want to make update this to be smarter by parsing `workspace.metadata.raze` and `package.metadata.raze` and combining the results. Can we please defer this to another PR?

impl/src/rendering/bazel.rs
UebelAndre: Can you elaborate?

impl/src/planning/subplanners.rs
UebelAndre: This is only a slight modification to what was there before. Can this be deferred or can you write a suggestion?

UebelAndre:
https://github.com/google/cargo-raze/pull/276#discussion_r520806557
@acmcarther what are you referring to here?

impl/src/testing.rs
```diff
@@ -102,13 +142,21 @@ pub fn make_basic_workspace() -> (TempDir, CargoWorkspaceFiles) {
...
 /**
  * Configures the given mock_server (representing a crates.io remote) to return
  * mock responses for the given crate and version .
  */
-pub fn mock_remote_crate(name: &str, version: &str, mock_server: &MockServer) -> TempDir {
+pub fn mock_remote_crate<'server>(
+  name: &str,
+  version: &str,
+  mock_server: &'server MockServer,
+) -> (TempDir, Vec<MockRef<'server>>) {
```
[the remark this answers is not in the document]
UebelAndre: This is a pattern that existed in the project before I started working on this project: https://github.com/google/cargo-raze/blob/master/impl/src/testing.rs#L72-L99

Could this not be considered acceptable in test code? I'm happy to update documentation for this function to explain the return type.

UebelAndre:
> > This is going to take a very long time to review.
> 
> Why is that? Is there something I can do to help with that?

Oh, is it just that you're busy @acmcarther ? Totally understandable if so.

Also, this PR is ready for another review

impl/src/bin/cargo-raze.rs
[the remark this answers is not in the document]
acmcarther: Comment was regarding name of variable.

impl/src/bin/cargo-raze.rs
```diff
@@ -89,77 +96,81 @@ fn main() -> Result<()> {
...
+      for entry in glob::glob(&build_glob)? {
+        if let Ok(path) = entry {
+          fs::remove_file(path)?;
+        }
+      }
+    }
+  }
+
+  for output in bazel_file_outputs.iter() {
+    if options.flag_dryrun.unwrap_or(false) {
```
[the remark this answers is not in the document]
acmcarther: Ignored this comment?

impl/src/metadata.rs
```diff
@@ -19,16 +19,67 @@ use std::{
...
 
+use glob::glob;
+use pathdiff::diff_paths;
+use regex::Regex;
+use rustc_serialize::hex::ToHex;
 use tempfile::TempDir;
+use url::Url;
 
-use rustc_serialize::hex::ToHex;
+pub use cargo_metadata::{DependencyKind, Metadata, Node, Package, PackageId};
```
acmcarther, line 34: Why pub use?

impl/src/metadata.rs
```diff
@@ -19,16 +19,58 @@ use std::{
...
+#[derive(Debug, Clone)]
+pub struct RazeMetadata {
+  pub metadata: Metadata,
+  pub checksums: HashMap<String, String>,
+  pub workspace_root: PathBuf,
+  pub lockfile: Option<Lockfile>,
+}
+
+impl RazeMetadata {
+  pub fn new(
```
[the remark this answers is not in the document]
acmcarther: Ignored this comment?

impl/src/metadata.rs
```diff
@@ -19,16 +19,58 @@ use std::{
...
+  ) -> RazeMetadata {
+    RazeMetadata {
+      metadata,
+      checksums,
+      workspace_root,
+      lockfile,
+    }
+  }
+
+  pub fn get_checksum(&self, name: &str, version: &str) -> Option<&String> {
```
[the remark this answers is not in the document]
acmcarther: Ignored this comment?

impl/src/metadata.rs
```diff
@@ -19,16 +19,58 @@ use std::{
...
+    self.checksums.get(&format!("{}-{}", name, version))
+  }
+}
+
+/** A struct containing information about a binary dependency */
```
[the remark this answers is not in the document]
acmcarther: Ignored this comment?

impl/src/metadata.rs
```diff
@@ -37,7 +79,12 @@ const SYSTEM_CARGO_BIN_PATH: &str = "cargo";
  * <https://github.com/rust-lang/cargo/pull/5122>
  */
 pub trait MetadataFetcher {
-  fn fetch_metadata(&self, files: &CargoWorkspaceFiles) -> Result<Metadata>;
+  fn fetch_metadata(
+    &self,
+    files: &CargoWorkspaceFiles,
+    binary_dep_info: Option<&HashMap<String, cargo_toml::Dependency>>,
+    override_lockfile: Option<PathBuf>,
```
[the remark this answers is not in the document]
acmcarther: Ignored this comment?

impl/src/metadata.rs
```diff
@@ -50,25 +97,132 @@ pub struct CargoWorkspaceFiles {
 /** A workspace metadata fetcher that uses the Cargo Metadata subcommand. */
 pub struct CargoMetadataFetcher {
   cargo_bin_path: PathBuf,
-  use_tempdir: bool,
+  registry_url: Url,
+  index_url: String,
```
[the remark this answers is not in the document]
acmcarther: Ignored this comment?

impl/src/metadata.rs
```diff
@@ -50,25 +97,132 @@ pub struct CargoWorkspaceFiles {
...
 }
 
 impl CargoMetadataFetcher {
-  pub fn new<P: Into<PathBuf>>(cargo_bin_path: P, use_tempdir: bool) -> CargoMetadataFetcher {
+  pub fn new<P: Into<PathBuf>>(
+    cargo_bin_path: P,
+    registry_url: Url,
+    index_url: &str,
```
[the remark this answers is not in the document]
acmcarther: Ignored this comment?

impl/src/metadata.rs
```diff
@@ -50,25 +97,132 @@ pub struct CargoWorkspaceFiles {
...
-    &self,
-    registry_url: &str,
-    dir: &Path,
-    name: &str,
-    version: &str,
-  ) -> Result<PathBuf> {
-    let mut url = url::Url::parse(registry_url)?;
+  // Create a symlink file on unix systems
+  #[cfg(target_family = "unix")]
+  fn make_symlink(&self, src: &Path, dest: &Path) -> Result<()> {
```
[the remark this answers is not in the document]
acmcarther: Ignored this comment?

impl/src/metadata.rs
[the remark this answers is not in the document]
acmcarther: There must be something wrong with the UI. Try opening the diff view. Do you see the full comment chain then?

Reproducing the comment:

This above can optionally be condensed.
```
let filename = {
  let filename_match =  re.captures(&member.repr).and_then(|c| c.get(1));
  if (filename_match.is_none()) {
    continue;
  }
  PathBuf::from(filename_match.unwrap()); // UNWRAP: Guarded above
};
```
(i don't really feel strongly about the forumulation of the unwrap -- more important is that you can use a map to avoid the second match)

impl/src/metadata.rs
acmcarther: Ignored this comment?

impl/src/metadata.rs
acmcarther: (is this because you wanted to access it from other modules through this module? if so, I'd really prefer you just import it like normal. If it's because you want the library to export this type, I also find that questionable -- I don't think it's a great idea to export other libraries types through our interface)

UebelAndre: I was thinking of this from the perspective of external crates (things that might import `cargo-raze`). The `RazeMetadata` struct is the source of truth so I thought whatever it was using should be re exported. I thought this was a common pattern since I'd run into it in the `cargo_lock` and `cargo_metadata` crates in how they handle `semver::Version`. Should I still change this?

impl/src/bin/cargo-raze.rs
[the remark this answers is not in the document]
UebelAndre: Updated the variable name.

impl/src/bin/cargo-raze.rs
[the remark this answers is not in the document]
UebelAndre: Sorry, I don't know why this disappeared from the overview.

impl/src/rendering/bazel.rs
```diff
@@ -301,17 +338,26 @@ impl BuildRenderer for BazelRenderer {
...
+        path: path_prefix.as_path().join("Cargo.raze.lock"),
+        contents: lockfile.to_string(),
+      });
+    }
+
+    // Ensure there is always a `BUILD.bazel` file to accompany `crates.bzl`
+    let crates_bzl_pkg_file = path_prefix.as_path().join("BUILD.bazel");
+    if !file_outputs
+      .iter()
+      .any(|output| output.path.eq(&crates_bzl_pkg_file))
```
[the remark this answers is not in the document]
UebelAndre: Can-do

impl/src/metadata.rs
```diff
@@ -19,16 +19,70 @@ use std::{
...
+use regex::Regex;
+use rustc_serialize::hex::ToHex;
 use tempfile::TempDir;
+use url::Url;
 
-use rustc_serialize::hex::ToHex;
+// These attriburtes are the source of truth for the Metadata used by Cargo raze.
+// Making them public allows for dependents to always ensure they're using the 
+// same versions `cargo-raze` is.
+pub use cargo_metadata::{Metadata, Package, PackageId};
```
acmcarther: Not sure why the comment disappeared, but RE your last response.

Lets not reexport this. If an end users needs this, we can decide to reexport it at that time once a specific case has been made.

impl/src/testing.rs
```diff
@@ -215,5 +266,9 @@ pub fn mock_crate_index(crates: &HashMap<String, String>) -> TempDir {
     .unwrap();
   }
 
-  index_url_mock_dir
+  if mock_dir.is_none() {
```
acmcarther, line 269: It seems this function always returns Some(x). Update it to not return Option anymore?

UebelAndre:
@acmcarther Don't know why comments are being hidden (maybe the rebasing?) but I think I've gone over everything now.

impl/src/metadata.rs
GregBowyer: I would think its best for now to not be pub, I dont feel there is a good surface API in cargo-raze for exporting.

impl/src/testing.rs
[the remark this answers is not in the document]
acmcarther: Hmm.

That is an entirely valid and powerful rebuttal, but I think that case is slightly different because it returns two self-descriptive types. These are not quite self descriptive.

impl/src/bin/cargo-raze.rs
```diff
@@ -76,16 +80,21 @@ fn main() -> Result<()> {
...
   let settings = load_settings("Cargo.toml")?;
   if options.flag_verbose.unwrap_or(false) {
     println!("Loaded override settings: {:#?}", settings);
   }
 
-  let mut metadata_fetcher: Box<dyn MetadataFetcher> = match options.flag_cargo_bin_path {
-    Some(ref p) => Box::new(CargoMetadataFetcher::new(p, /*use_tempdir: */ true)),
+  // Fetch metadata
+  let metadata_fetcher: Box<dyn MetadataFetcher> = match options.flag_cargo_bin_path {
+    Some(ref p) => Box::new(CargoMetadataFetcher::new(
```
GregBowyer, line 91: Nit: I might be a bit famous for making variable names smaller (scope-declares-verbosity) but even by my standards I think this can be something like `path`

impl/src/metadata.rs
UebelAndre: I think that's worth exploring but I think that should be a focused effort. I've made these not public.

impl/src/rendering/bazel.rs
acmcarther: L222
```
    for member_path in planned_build.workspace_context.workspace_members.iter() {
      let all_packages: Vec<CrateContext> = planned_build
        .crate_contexts
        .iter()
        .filter(|ctx| {
          ctx.is_binary_dependency || ctx.workspace_member_dependents.contains(member_path)
        })
        .map(|ctx| ctx.clone())
        .collect();
      let rendered_alias_build_file = self
        .render_aliases(&workspace_context, &all_packages)
        .map_err(|e| RazeError::Rendering {
          crate_name_opt: None,
          message: unwind_tera_error!(e),
        })?;

      file_outputs.push(FileOutputs {
        path: render_details
          .cargo_root
          .join(member_path)
          .join(&render_details.workspace_member_output_dir)
          .join("BUILD.bazel"),
        contents: rendered_alias_build_file,
      });
    }
```

L292
```
      let all_packages: Vec<CrateContext> = planned_build
        .crate_contexts
        .iter()
        .filter_map(|ctx| {
          if ctx.is_binary_dependency {
            Some(ctx.clone())
          } else if ctx.workspace_member_dependents.contains(member_path) {
            Some(ctx.clone())
          } else {
            None
          }
        })
        .collect();
      let rendered_alias_build_file = self
        .render_remote_aliases(&workspace_context, &all_packages)
        .map_err(|e| RazeError::Rendering {
          crate_name_opt: None,
          message: unwind_tera_error!(e),
        })?;

      file_outputs.push(FileOutputs {
        path: render_details
          .cargo_root
          .join(member_path)
          .join(&render_details.workspace_member_output_dir)
          .join("BUILD.bazel"),
        contents: rendered_alias_build_file,
      });
```

impl/src/metadata.rs
UebelAndre: Oh, and thanks @GregBowyer for checking out this PR!

I'm particularly interested in hearing feedback on the functionality in here. Would love to get your thoughts/review 😄

impl/src/bin/cargo-raze.rs
```diff
@@ -181,52 +192,34 @@ fn write_to_file(path: &Path, contents: &str, verbose: bool) -> Result<()> {
...
-    )
-    .unwrap(),
-    std::path::Path::new("direct").join("path")
-  );
+/** Helper function for gathering inputs for `genmode = "Remote"` builds */
+fn gather_remote_genmode_inputs<'settings>(
+  bazel_root: &Path,
+  settings: &'settings RazeSettings,
+) -> RemoteGenModeInputs<'settings> {
+  if settings.genmode == GenMode::Remote {
```
acmcarther, line 206: Nit: This is cleaner structured as an early return:

```rust
if settings.genmode != GenMode::Remote {
    return RemoteGenModeInputs {
      override_lockfile: None,
      binary_deps: None,
    }
}
// the rest

File: impl/src/metadata.rs (hunk 27 above)
Comment (UebelAndre) [reply impl/src/metadata.rs]: This has been updated and is no longer public.

**Review by acmcarther — COMMENTED** [review]
functions are too long man

too long

File: impl/src/metadata.rs (hunk 24 above)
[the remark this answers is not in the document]
Comment (GregBowyer) [reply impl/src/metadata.rs:113]: Typically from my experience yes, `&String` would suggest that you are giving a reference directly to an owned thing, at which point since I dont get ownership of it in any meaningful way (`Box<String>` / `String` / `&mut String`) I am curious why not have the more primal type.

In the case of avoiding a copy but allowing it if needed I would more expect to see `Cow<'a, str>` as the type/

My two cents, I have not seen much `&String` in the wild

**Comment by UebelAndre** [comment]
https://github.com/google/cargo-raze/pull/276#discussion_r521759396
Ah! I see, I thought you were saying I had duplicate code within the same function.

**Comment by acmcarther** [comment]
Humble request to stop rebasing the commits unless you have a good reason. It's wrecking the review history.

I try to click https://github.com/google/cargo-raze/pull/276/files/558e3cb7b2259a369880d26642269f376801b73c#diff-c8fe41d4c65c6cc6fc8160d444cb04842ac218c135768f6e4309b18ec27ee784 because you asked me to provide an example, and the link is broken

File: impl/src/planning/crate_catalog.rs [hunk 31]
```diff
@@ -198,39 +184,35 @@ impl CrateCatalog {
...
-      .cloned()
-      .collect::<HashSet<_>>();
-    let workspace_crates = metadata
-      .workspace_members
+    let workspace_crates: Vec<&Node> = resolve
+      .nodes
       .iter()
-      .cloned()
-      .collect::<HashSet<_>>();
+      .filter_map(|node| {
```
GregBowyer, line 190: Could just be 

```rust
filter(|node| metadata.workspace_members.contains(&node.id))
```

no ?

impl/src/planning/subplanners.rs
```diff
@@ -143,40 +134,44 @@ impl<'planner> WorkspaceSubplanner<'planner> {
       workspace_path: self.settings.workspace_path.clone(),
       gen_workspace_prefix: self.settings.gen_workspace_prefix.clone(),
       output_buildfile_suffix: self.settings.output_buildfile_suffix.clone(),
+      workspace_members: self
+        .metadata
+        .metadata
+        .workspace_members
+        .iter()
+        .filter_map(|pkg_id| {
```
GregBowyer, line 142: I might be missing something but why not

```rust
self
.metadata
.metadata // <-- Also why this stutter?
.packages
.iter()
.find(|pkg| pkg.id == &pkg_id) // <-- Also why the javaesq `.eq(`
.map(|pkg| get_workspace_member_path(&pkg.manifest_path, &self.metadata.metadata.workspace_root))
```

impl/src/planning/subplanners.rs
GregBowyer: Minor: `Ok(` wrap on something that returns `Result` ? Why not unfold one level?

impl/src/planning/subplanners.rs
```diff
@@ -366,13 +314,40 @@ impl<'planner> CrateSubplanner<'planner> {
...
+      .crate_catalog_entry
+      .workspace_member_dependents
+      .iter()
+      .filter_map(|pkg_id| {
+        if let Some(workspace_member) = self
+          .crate_catalog
+          .metadata
+          .packages
+          .iter()
+          .find(|pkg| pkg.id.eq(pkg_id))
```
GregBowyer, line 327: Same as elsewhere is there a reason this is not `==` am I missing something or being dumb?

impl/src/rendering/bazel.rs
```diff
@@ -220,18 +219,32 @@ impl BuildRenderer for BazelRenderer {
...
+    for member_path in planned_build.workspace_context.workspace_members.iter() {
+      let all_packages: Vec<CrateContext> = planned_build
+        .crate_contexts
+        .iter()
+        .filter(|ctx| {
+          ctx.is_binary_dependency || ctx.workspace_member_dependents.contains(member_path)
+        })
+        .map(|ctx| ctx.clone())
+        .collect();
+      let rendered_alias_build_file = self
```
GregBowyer: Nit: I would put a line break between the two bindings to avoid everything becoming a wall of text over time.

impl/src/settings.rs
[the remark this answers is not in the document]
GregBowyer: Yeah I can be the king of crazy patterns and even for me this is a bit much.

impl/src/testing.rs
```diff
@@ -45,6 +46,48 @@ dependencies = [
...
+
+[lib]
+path = "not_a_file.rs"
+
+[dependencies]
+proc-macro2 = "1.0.24"
+"#
+}
+
+pub fn advanced_lock() -> &'static str {
```
GregBowyer, line 63: I would use `indoc` to make the function / const indented. I would also consider a const here

UebelAndre:
> Humble request to stop rebasing the commits unless you have a good reason. It's wrecking the review history.
> 
> I try to click https://github.com/google/cargo-raze/pull/276/files/558e3cb7b2259a369880d26642269f376801b73c#diff-c8fe41d4c65c6cc6fc8160d444cb04842ac218c135768f6e4309b18ec27ee784 because you asked me to provide an example, and the link is broken

Ah, sorry, I'll add new commits from here on without rebasing 😅

impl/src/testing.rs
GregBowyer: maybe 

```rust
mock_dir.or_else(|| Some(index_url_mock_dir))
```

impl/src/bin/cargo-raze.rs
UebelAndre: Updated

UebelAndre:
> [#276 (comment)](https://github.com/google/cargo-raze/pull/276#discussion_r521759396)
> Ah! I see, I thought you were saying I had duplicate code within the same function.

This is updated

UebelAndre:
https://github.com/google/cargo-raze/pull/276#discussion_r521757623
So would documentation suffice or what should I do?

impl/src/testing.rs
UebelAndre: No, it does return `None` some in cases
```
test planning::tests::test_binary_dependencies_vendored_genmode ... FAILED
test planning::tests::test_binary_dependencies_remote_genmode ... FAILED
```
I added a panic on that line and these tests tripped it.

impl/src/planning/subplanners.rs
UebelAndre: No, this needs to iterate workspace members specifically.

The 'stutter' is is the name `metadata` being used in two different contexts. One is the glob of metadata output by the metadata fetcher and the inner member is the `cargo_metadata::Metadata`.

impl/src/planning/subplanners.rs
UebelAndre: I found this to be clearer than de-referencing references to allow for `==` but that may be personal experience. I've since updated these.

impl/src/rendering/bazel.rs
UebelAndre: Done

impl/src/testing.rs
UebelAndre: This was done to be consistent with what was there before (`basic_toml` and `basic_lock`) but I can update it.

UebelAndre: How would I apply `const` here? I thought that was implicit

UebelAndre: I've updated these now.

impl/src/bin/cargo-raze.rs
UebelAndre: Sorry, can you elaborate?

impl/src/metadata.rs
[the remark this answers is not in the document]
acmcarther: I didn't mean &String, i meant String. Move the value.

The motivation here is basically that we already need an owned string here, but we take an &str because generally `&str` is what you want.

Here though that can result in an extra allocation if the end user no longer needed the string anyway.

impl/src/settings.rs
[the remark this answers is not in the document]
UebelAndre: What can I do do simplify this without having to write a large amount of new logic?

impl/src/testing.rs
GregBowyer:
```rust
pub const fn advanced_lock()
```

or 

```rust
const ADVANCED_LOCK: &str = indoc! { r#" str str str "#}
```

impl/src/testing.rs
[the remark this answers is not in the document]
UebelAndre: Updated

impl/src/planning/subplanners.rs
UebelAndre: Done

UebelAndre:
@acmcarther @GregBowyer Ready for your eyes again

impl/src/bin/cargo-raze.rs
GregBowyer: So I _personally_ follow the practise of making my variable names as descriptive as possible without going to the point of utter verbosity.

Basically the closer a variable is in scope to where it will be used the shorter I feel I can get away with.

For example `x` / `y` is ok for say

```rust
some_iter.map(|x| x.thing)
```

whereas say for your code since `p` is a little further away from its use I might have mental forgotten what `p` relates to by the time it gets used, so in that case I would go with `path`

If it was, yet further away I might go longer and use `metadata_path` (and so on and so forth).

I will make exceptions if the context while short is just utterly obvious or ubiquitous e.g. https://github.com/simh/simh/blob/master/VAX/vax_cpu.c#L631 (I work on this :/) where sure `R` is very short but if you read any documentation on a vax its _the name_ for registers.

What I aim to avoid is either this https://github.com/simh/simh/blob/master/VAX/vax_cpu.c#L3792 (Great `i` is used everywhere and means a million little contextual things) or the other way https://chromium.googlesource.com/native_client/nacl-gcc/+/ng/master/libgo/go/net/fd_windows.go#36 (`hasLoadSetFileCompletionNotificationModes`)

In this context I suggest that `p` is a little too short as I found I had forgotten what `p` was by the time it was used.

UebelAndre: That was a good read. Thanks 😄 

I've updated the variable name.

impl/src/settings.rs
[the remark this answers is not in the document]
UebelAndre: Suggestions are greatly welcomed 😅 🙏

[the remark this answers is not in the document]
UebelAndre: I've made a change but have no idea if it's better or not 😞

UebelAndre:
@acmcarther Just wanted to thank you for all the attention you gave this PR on Wednesday. Super appreciated! 🙏 

I was wondering if we'd be able to do something similar today. There are more changes I'd like to make but are all done on top of the changes here. I'm very happy with the general state of the project after the work here which were inspired by all of the thoughts you've shared about the design goals of `cargo-raze`. I truly believe that this is my final large PR since I can't think of any major features missing after this change. I would very much like to get this through to unblock other work. As always, looking forward to hearing from you 😄

UebelAndre:
@acmcarther I did one more small update. This is ready for review 😅

UebelAndre:
@acmcarther Hey, If you don't have the bandwidth for reviewing this, would you be able to recommend another reviewer? I totally understand that free time can be hard to come by but I'd really like to get this functionality in 😅 🙏

impl/src/bin/cargo-raze.rs
```diff
@@ -96,83 +105,85 @@ fn main() -> Result<()> {
     lock_path_opt,
   };
 
+  // Determine the path to the Bazel workspace root or fallback to the current working directory
```
acmcarther, line 108: Purely aesthetic thing, but it looks like your inline comments here are meant to describe blocks of code, but there are newlines sprinkled throughout which makes the association not obvious.

If 
```rust
 // Determine the path to the Bazel workspace root or fallback to the current working directory
```
is meant to describe L109 - L117, I think you should remove the newlines:

```rust
  let cargo_raze_working_dir = find_bazel_workspace_root().unwrap_or(std::env::current_dir()?);
  let remote_genmode_inputs = gather_remote_genmode_inputs(&cargo_raze_working_dir, &settings);
  let metadata = metadata_fetcher.fetch_metadata(
    &files,
    remote_genmode_inputs.binary_deps,
    remote_genmode_inputs.override_lockfile,
  )?;
```

The same applies to the `// Do Planning` comment, as well as the `// Render BUILD files` comment, as well as the `// Write BUILD files` comment.

impl/src/bin/cargo-raze.rs
```diff
@@ -96,83 +105,85 @@ fn main() -> Result<()> {
...
+      &output.path,
+      &output.contents,
+      options.flag_verbose.unwrap_or(false),
+    )?;
+  }
+
   Ok(())
 }
 
+/** Helper function for writing rendered files to disk */
```
acmcarther, line 186: This is a pretty extreme nit, but I'm making the comment because it applies in multiple places.

"Helper function" or "helper struct" in the function or type documentation add very little. Trimming them has more impact:

```
/** Writes rendered files to filesystem. */
```

```
/** Represents the inputs specific to the Remote genmode. */
```

```
/** Gathers inputs for the `genmode = "Remote"` builds. */
```

impl/src/metadata.rs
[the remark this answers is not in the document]
acmcarther: I don't feel strongly about this, but I'm really surprised you're not seeing these comments. I just click the `Files Changed` tab. Perhaps you're not expanding the unexpanded files?

impl/src/planning.rs
```diff
@@ -47,251 +45,157 @@ pub trait BuildPlanner {
...
+  fn plan_build(&self, platform_details: Option<PlatformDetails>) -> Result<PlannedBuild>;
 }
 
 /** The default implementation of a `BuildPlanner`. */
-pub struct BuildPlannerImpl<'fetcher> {
-  metadata_fetcher: &'fetcher mut dyn MetadataFetcher,
-  binary_deps_tempdir: Result<TempDir, io::Error>,
+pub struct BuildPlannerImpl {
+  pub metadata: RazeMetadata,
+  pub settings: RazeSettings,
```
acmcarther, line 54: Why is this stuff now pub?

acmcarther (changes requested):
I think you missed my (and GregBowyer's) comments which are included in unexpanded files. At least, that's the case with `metadata.rs`. 

Did you move functions around in metadata.rs? The diff makes it look like you basically deleted everything and rewrote it.

acmcarther:
Sorry for the delay in the review by the way. I recognize the huge effort you've put in both in this PR and in previous PRs.

impl/src/planning.rs
acmcarther: (i'm making this comment not because I have deep wisdom about this, but because it looks superficially like you might be breaking the abstraction to take an implementation shortcut)

impl/src/bin/cargo-raze.rs
UebelAndre: Do you think that comment is still necessary with the renamed comment? I really felt the `// Section ---` style clearly communicated that the `main` method was sectional and made those comments distinct from others describing functionality

impl/src/metadata.rs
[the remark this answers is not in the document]
UebelAndre: `checksums` was not pub but I can make it so. I thought keeping it private would drive users to use `get_checksum` instead of directly accessing the field but I suppose it's not that valuable.

impl/src/metadata.rs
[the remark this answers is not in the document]
UebelAndre: I don't think that `override_lockfile` should go with `CargoWorkspaceFiles` since it's intended to be a unique lockfile only generated by `cargo-raze`. The reason this file exists in the first place was `(b)`. That information does not exist within the projects cargo metadata because of the inconsistency in how Cargo handles binary dependencies. Raze generates a new lockfile to make sure subsequent runs are reproducible.

I actually feel that the `CargoWorkspaceFiles` struct should no longer exist. https://github.com/google/cargo-raze/pull/274 introduced a way to cleanly embed Raze settings into cargo metadata and because of this I feel we should actually be using a [`MetadataCommand`](https://docs.rs/cargo_metadata/0.12.0/cargo_metadata/struct.MetadataCommand.html) to load settings. It should no longer matter what `Cargo.toml` and `Cargo.lock` file is being used, we can the metadata from anywhere within a cargo workspace and parse settings. In the context of supporting workspaces, this would allow users to keep `CrateSettings` in the relevant workspace member toml files and have them be joined in an intelligent way to be used for the rest of the `raze` command. Additionally, should we need to locate the cargo files, `workspace_root` should always be the directory where the cargo files live for currently supported use cases (single/root package projects).

Somewhat related to this issue, there's work in https://github.com/UebelAndre/cargo-raze/commit/332ad931f374ea28e3100fe46c92e99dd51b6ae2 which is inspired by conversations in https://github.com/google/cargo-raze/pull/222 but also starts to point out that the `CargoWorkspaceFiles` is not needed. But for now I feel strongly that `override_lockfile` represents something different than `CargoWorkspaceFiles`. Though I agree with your point. I'd like to either delete the need for `override_lockfile` and have everything live in normal Cargo files or delete `CargoWorkspaceFiles` and use Cargo Metadata to gather the necessary information

impl/src/metadata.rs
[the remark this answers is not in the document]
UebelAndre: I've come across both cases. I'm not quite sure what all the trade-offs here are, ownership is the only one I've really had experience with. But I think you're right @acmcarther I think it'd be better to have the signature accept `String` to clearer communicate the intent and not mask any allocations.

impl/src/metadata.rs
UebelAndre: Added documentation and updated this to an error

impl/src/metadata.rs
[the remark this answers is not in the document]
UebelAndre: I made it a `Url` I think it's the cleaner implementation

impl/src/metadata.rs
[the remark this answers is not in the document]
UebelAndre: Moved it back (I think to the right spot)

impl/src/planning.rs
UebelAndre: It didn't need to be. Updated

UebelAndre:
@acmcarther Ok, I think I've found all the comments. Still learning how to use GitHub. I wonder if there's a user setting to always show comments in the "Conversation" tab. I assumed everything showed up there and that I could treat each comment like a TODO list and resolve conversations as appropriate. I'm definitely not ignoring comments. Maybe just going through the feedback a bit too quickly 😅 

This is ready for another review 🙏 I think it's getting closer!! 😄

impl/src/planning.rs
acmcarther: This still looks pub as of most recent commit.

impl/src/bin/cargo-raze.rs
acmcarther: The essential aspect of my comment is that "// some comment" doesn't obviously apply to multiple lines of code once you insert newlines between the lines of code.

The "// whatever ---" seems completely arbitrary. Arbitrary conventions don't really communicate intent very well.

impl/src/planning.rs
UebelAndre: Sorry 😢 I did forget to add that file.

impl/src/bin/cargo-raze.rs
UebelAndre: Ah, I see. So what's the action item from here on?

impl/src/settings.rs
```diff
@@ -369,6 +387,51 @@ fn validate_settings(settings: &mut RazeSettings) -> Result<(), RazeError> {
   Ok(())
 }
 
+/** Parses raze settings from the contents of a `Cargo.toml` file */
+fn parse_raze_settings(toml_contents: CargoToml) -> Result<RazeSettings, RazeError> {
+  // Workspace takes precidence
```
acmcarther, line 392: Nit s/precidence/precedence

impl/src/settings.rs
```diff
@@ -369,6 +387,51 @@ fn validate_settings(settings: &mut RazeSettings) -> Result<(), RazeError> {
...
+  }
+
+  // Next is package
+  if let Some(raze) = toml_contents
+    .package
+    .and_then(|pkg| pkg.metadata.and_then(|data| data.raze))
+  {
+    if toml_contents.raze.is_some() {
+      eprintln!(
+        "WARNING: Both [raze] and [pakcage.metadata.raze] are set. Using [package.metadata.raze] \
```
acmcarther, line 413: Nit: s/pakcage/package

impl/src/settings.rs
```diff
@@ -369,6 +387,51 @@ fn validate_settings(settings: &mut RazeSettings) -> Result<(), RazeError> {
   Ok(())
 }
 
+/** Parses raze settings from the contents of a `Cargo.toml` file */
+fn parse_raze_settings(toml_contents: CargoToml) -> Result<RazeSettings, RazeError> {
+  // Workspace takes precidence
+  if let Some(raze) = toml_contents
+    .workspace
+    .and_then(|pkg| pkg.metadata.and_then(|data| data.raze))
```
acmcarther, line 395: This appears to add a third possible configuration location:

- [raze]
- [package.metadata.raze]
- [workspace.metadata.raze]

This was not mentioned in the PR description. Why is this being added?

impl/src/planning/subplanners.rs
```diff
@@ -366,13 +312,40 @@ impl<'planner> CrateSubplanner<'planner> {
...
+      .crate_catalog_entry
+      .workspace_member_dependents
+      .iter()
+      .filter_map(|pkg_id| {
+        if let Some(workspace_member) = self
+          .crate_catalog
+          .metadata
+          .packages
+          .iter()
+          .find(|pkg| pkg.id == *pkg_id)
```
acmcarther, line 325: Nit: This expression is too long.

```rust
    let workspace_package = self
        .crate_catalog
        .metadata
        .packages
        .iter()
        .find(|pkg| pkg.id == *pkg_id);
    if let Some(workspace_package) = workspace_package {
      // etc
    } else {
      // etc
    }
  ```

**Comment by UebelAndre** [comment]
> I think you missed my (and GregBowyer's) comments which are included in unexpanded files. At least, that's the case with `metadata.rs`.
> 
> Did you move functions around in metadata.rs? The diff makes it look like you basically deleted everything and rewrote it.

I don't remember doing a large shuffle. But I did just restore the placement of one thing and split some functionality from `make_temp_workspace` into it's own function. Trying to keep things minimal now

File: impl/src/planning/subplanners.rs [hunk 43]
```diff
@@ -143,40 +134,44 @@ impl<'planner> WorkspaceSubplanner<'planner> {
...
+        .metadata
+        .workspace_members
+        .iter()
+        .filter_map(|pkg_id| {
+          if let Some(pkg) = self
+            .metadata
+            .metadata
+            .packages
+            .iter()
+            .find(|pkg| pkg.id == *pkg_id)
```
acmcarther, line 148: Nit: Made this comment elsewhere, duplicating here.

Can we avoid these extremely long conditional clauses?

```rust
let workspace_member = self
            .metadata
            .metadata
            .packages
            .iter()
            .find(|pkg| pkg.id == *pkg_id);
if let Some(workspace_member) = workspace_member {
```


```

File: impl/src/planning/subplanners.rs [hunk 44]
```diff
@@ -101,39 +101,30 @@ pub struct WorkspaceSubplanner<'planner> {
...
-  pub(super) files: &'planner CargoWorkspaceFiles,
-  pub(super) binary_dependencies: &'planner Vec<CrateCatalog>,
-  pub(super) binary_deps_files: &'planner HashMap<String, CargoWorkspaceFiles>,
+  pub(super) metadata: &'planner RazeMetadata,
 }
 
 impl<'planner> WorkspaceSubplanner<'planner> {
   /** Produces a planned build using internal state. */
   pub fn produce_planned_build(&self) -> Result<PlannedBuild> {
+    // Check for errors ---
```
acmcarther, line 110: This "---" is not a convention of comment writing, as far as I know. I prefer regular comments and avoiding interstitial newlines to communicate that a comment refers to a whole block of code.

impl/src/metadata.rs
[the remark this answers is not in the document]
acmcarther: ¯\\\_(ツ)_/¯

acmcarther (changes requested):
I have the sneaking suspicion that this is going to to be a breaking change, but the test suite isn't good enough to catch it, and I'm no longer familiar enough with the code to identify it in review.

Oh well I guess. I think this is the last pass of comments.

impl/src/bin/cargo-raze.rs
acmcarther: I'm saying don't do the "---" thing. When a comment applies to multiple lines, you have two options.

1. Keep the lines of code together. Don't add whitespace
2. Move the lines into a function.

impl/src/settings.rs
UebelAndre: This PR was opened before https://github.com/google/cargo-raze/pull/274 was merged and I had to make this change after a rebase. I just slipped my mind to add it to the PR description. This is being added because it makes sense to me when authoring projects using cargo workspaces. Associating raze settings with a package gets confusing when there's no root package. To solve for this case, we should use `[workspace.metadata]`

impl/src/planning/subplanners.rs
UebelAndre: I'll keep this in mind for future changes.

impl/src/planning/subplanners.rs
UebelAndre: Updated

impl/src/planning/subplanners.rs
UebelAndre: Updated

impl/src/settings.rs
UebelAndre: I also updated the PR description.

UebelAndre:
@acmcarther Updated. Back to you 🙏

UebelAndre:
> I have the sneaking suspicion that this is going to to be a breaking change, but the test suite isn't good enough to catch it, and I'm no longer familiar enough with the code to identify it in review.
> 
> Oh well I guess. I think this is the last pass of comments.

This may be a "breaking" change for more obscure projects. Notable the generation of the aliases is controlled by the `workspace_member_dir` setting and is always relative to the location of `Cargo.toml` files. That said though, I think with that setting, users should be able to get new generations with these changes to match previous outputs.

UebelAndre:
Fixed a typo which should fix CI.

UebelAndre:
https://github.com/google/cargo-raze/pull/277 is also green with no changes other than a rebase

UebelAndre:
@acmcarther Hey, are you around to take another quick look? 😅

impl/src/bin/cargo-raze.rs
```diff
@@ -76,16 +80,21 @@ fn main() -> Result<()> {
...
+  // Fetch metadata
+  let metadata_fetcher: Box<dyn MetadataFetcher> = match options.flag_cargo_bin_path {
+    Some(ref cargo_bin_path) => Box::new(CargoMetadataFetcher::new(
+      cargo_bin_path,
+      Url::parse(&settings.registry)?,
+      Url::parse(&settings.index_url)?,
+    )),
     None => Box::new(CargoMetadataFetcher::default()),
   };
-  let mut planner = BuildPlannerImpl::new(&mut *metadata_fetcher);
```
acmcarther, line 98: Still a bunch of whitespace here between L89 "fetch metadata" and L118 "do planning".

impl/src/bin/cargo-raze.rs
```diff
@@ -96,83 +105,84 @@ fn main() -> Result<()> {
...
-      false => None,
-    },
-    settings.incompatible_relative_workspace_path,
-  )?;
+  let planned_build =
+    BuildPlannerImpl::new(metadata.clone(), settings.clone()).plan_build(platform_details)?;
 
-  let planned_build = planner.plan_build(&settings, &prefix_path, files, platform_details)?;
+  // Render BUILD files
   let mut bazel_renderer = BazelRenderer::new();
```
acmcarther, line 129: Whitespace?

impl/src/bin/cargo-raze.rs
UebelAndre: Sorry, I thought removing the comment form the earlier conversation meant this was not necessary. Updated

impl/src/bin/cargo-raze.rs
UebelAndre: Updated

UebelAndre:
@acmcarther Removed the whitespace. Ready for another pass 🙏

acmcarther (approved):

acmcarther:
Will wait for green from CI

UebelAndre:
@acmcarther Just finished 😄

UebelAndre:
@acmcarther thank you so much for all your support!!! 🙏

dae:
@UebelAndre would it make sense for workspace_member_dir to default to "." for compatibility? After updating I found I was getting cargo/cargo/BUILD.bazel instead of cargo/BUILD.bazel, and had to dig around to find out how to change it.

UebelAndre:
@dae I can do that but I originally worried about setting it to `"."` because I was afraid of overriding valuable build files.

Also, what do you think of the name `workspace_member_dir`? It's really just the Bazel package aliases should be rendered into. Is there a  better name for it?

dae:
Would `build_file_output_dir` work, or does it have other uses I have missed?

UebelAndre:
@dae Also, is your project public? I'd love to know more about how raze is used. From what I've gathered, there's two kinds of usage:
1. Cargo projects that are being ported to bazel.
2. Bazel projects that want to introduce dependencies on crates.io crates. 

I've been focusing on the first case for some time which is partly why `workspace_member_dir` ended up in the state that it is. I'd love to get to a point where either users don't have to define bogus `Cargo.toml` files to use this tool. Then for the second point, I'd want to update `rules_rust` to be able to act as the authority and interract with crate registries from within Bazel entirely. No `Cargo.toml` files required. Users should be able to author their targets to define all the dependencies they require and the rules should efficiently handle the rest. I think there's a very real path forward to that goal with this tool and maybe some updates to `rules_gazelle`. 

Anywho, I'd love to see your project structure and to know which camp you fall into (camp 1 or camp 2) to give me better context for future changes.

UebelAndre:
> Would `build_file_output_dir` work, or does it have other uses I have missed?

I feel like that's similar to `workspace_member_dir` in that it doesn't quite communicate the intent. I was thinking something like `package_aliases_dir`. Your thoughts there?

dae:
`aliases_build_file_dir`? `top_build_file_dir`? 

I run an open source project that was recently converted to Bazel, though it does not currently use workspaces: https://github.com/ankitects/anki

dae:
(the point being that when the BUILD file moved, I was hunting for something related to the build file that controlled where it was placed)

UebelAndre:
> (the point being that when the BUILD file moved, I was hunting for something related to the build file that controlled where it was placed)

Would I be correct in saying your primary frustration was in the lack of documentation? This I can understand and I'm going to be writing documentation ASAP if that's the case or not.  I've created https://github.com/google/cargo-raze/pull/306 to start addressing this feedback. We should continue the conversation there.

Also, thanks for showing me your project! 🙏

PR #278, opened 2020-11-03
Updated the bazel version and rules_rust hash used for smoketest

UebelAndre:
This causes the tests to be built with `rustc` version `1.47.0` which is the new default in `rules_rust`.


acmcarther (approved):
LGTM

PR #279, opened 2020-11-03
Moved template header into it's own file

UebelAndre:
Since there is a nearly identical header for two types of files, I thought it'd make sense to create a single source of truth for this.

This is more work that's been broken out of https://github.com/google/cargo-raze/pull/276


UebelAndre:
@acmcarther Hey, you around today to check this out?

acmcarther (approved):
LGTM

PR #280, opened 2020-11-05
Add support for wasm32-wasi target.

PiotrSikora:

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,11 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "io_bazel_rules_rust",
-    sha256 = "1211f57f6cbc5d68aff40d78d2548c8adee8f2921b54dca6ec9a9e7df60db13b",
-    strip_prefix = "rules_rust-224fe6a81b8ffa9c585c75f49a41f9b9924d516d",
+    sha256 = "17dbf791f4dab0fd4496ce5345af35e9ce2f6d011c1c8423436da517d019a3ea",
+    strip_prefix = "rules_rust-2f97db595b05b1ee8cc44bde5bdf03c00bd169fb",
     urls = [
-        # Master branch as of 2020-10-26
-        "https://github.com/bazelbuild/rules_rust/archive/224fe6a81b8ffa9c585c75f49a41f9b9924d516d.tar.gz",
+        # Master branch as of 2020-11-05
+        "https://github.com/bazelbuild/rules_rust/archive/2f97db595b05b1ee8cc44bde5bdf03c00bd169fb.tar.gz",
     ],
 )
 
--- a/examples/WORKSPACE
+++ b/examples/WORKSPACE
@@ -4,11 +4,11 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "io_bazel_rules_rust",
-    sha256 = "1211f57f6cbc5d68aff40d78d2548c8adee8f2921b54dca6ec9a9e7df60db13b",
-    strip_prefix = "rules_rust-224fe6a81b8ffa9c585c75f49a41f9b9924d516d",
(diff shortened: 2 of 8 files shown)


PiotrSikora:
Blocked on https://github.com/bazelbuild/rules_rust/pull/477

PiotrSikora:
@UebelAndre PTAL.

UebelAndre (approved):
Looks good to me! Thank you so much for adding this!! 🙏

UebelAndre:
@dfreese @acmcarther would either of you be able to take a look as well?

dfreese (approved):

wchargin:
As a user of `cargo-raze`, is it possible to disable this? This requires
latest `rules_rust`, which also requires Bazel 3.0.0 rather than 0.17.x,
and thus is non-trivial for us to incorporate.

UebelAndre:
What's the specific issue you're running into? You might be able to prevent that target from showing up by giving an explicit list of platform triples to raze via `targets`: 
https://github.com/google/cargo-raze/blob/master/impl/src/settings.rs#L69-L75

PiotrSikora:
@UebelAndre I believe the issue is with bazelbuild/rules_rust#463, not `wasm32-wasi` itself, and anything that requires changes from `rules_rust` after that PR was merged will be an issue for @wchargin.

UebelAndre:
> @UebelAndre I believe the issue is with [bazelbuild/rules_rust#463](https://github.com/bazelbuild/rules_rust/pull/463), not `wasm32-wasi` itself, and anything that requires changes from `rules_rust` after that PR was merged will be an issue for @wchargin.

Ah, that makes more sense. Perhaps a new issue should be opened for this in `rules_rust` linking the relevant issues then?

wchargin:
If I depend on `rand = "0.7.3"` and generate build files with cargo-raze
0.7.0, there’s a dependency edge to `:wasm32-wasi`, which doesn’t exist
at my `rules_rust` version:

```
$ bazel build //third_party/rust:rand
ERROR: While resolving configuration keys for @raze__getrandom__0_1_15//:getrandom: no such target '@io_bazel_rules_rust//rust/platform:wasm32-wasi': target 'wasm32-wasi' not declared in package 'rust/platform' defined by /HOMEDIR/.cache/bazel/_bazel_wchargin/52a95bbdd50941251730eb33b7476a66/external/io_bazel_rules_rust/rust/platform/BUILD
ERROR: Analysis of target '//third_party/rust:rand' failed; build aborted: Analysis failed
INFO: Elapsed time: 0.139s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
```

Setting `targets` on the top-level `raze` settings does seem to fix the
problem; thanks! It doesn’t work if I set it only for the `getrandom`
crate, but I guess it makes sense to define this globally anyway.

(I don’t really see how this is related to the linker API? That’s the
bit that requires Bazel 3.0.0+, yes, but that seems unrelated to me.)

UebelAndre:
> Setting `targets` on the top-level `raze` settings does seem to fix the
> problem; thanks! It doesn’t work if I set it only for the `getrandom`
> crate, but I guess it makes sense to define this globally anyway.

`targets` is a top level setting (or is expected to be) so yeah, you should set it top level like you see in the example I added to my comment above.

> (I don’t really see how this is related to the linker API? That’s the
> bit that requires Bazel 3.0.0+, yes, but that seems unrelated to me.)

I think there was just some confusion around what the core issue was. I do think we `rules_rust` should do something about this though. I'll ping https://github.com/bazelbuild/rules_rust/issues/415 and see if a release can go out to communicate this incompatibility

wchargin:
And thanks to both of you for such quick and helpful responses!

PiotrSikora:
> (I don’t really see how this is related to the linker API? That’s the
> bit that requires Bazel 3.0.0+, yes, but that seems unrelated to me.)

Indeed, the `wasm32-wasi` target requires very recent `rules_rust`, but I don't think it's going to be the only change that does that, which means that the core issue (at least, in my mind) is your inability to update `rules_rust` because of that PR.

PR #282, opened 2020-11-06
Rework planning to be based off the resolve tree and not directly pkgs

### LINKED ISSUE #144: Can I have two different versions of the same Rust library in one Bazel project?
Hi! Can I have two different versions of the same Rust library in one Bazel project?
For example, some projects within one Bazel repository can depend on `futures-0.1.x`, but some can require `futures-0.3.x`. How to handle such use case using cargo-raze?

### LINKED ISSUE #241: When no target set, flate2 fails to compile with latest HEAD
Carrying on from https://github.com/google/cargo-raze/pull/212#issuecomment-705507497

To reproduce, checkout https://github.com/ankitects/cargo-raze-bug-examples and run repro.sh. Tested on a Mac.

### LINKED ISSUE #269: Erroneous aliases for async-global-executor-1.4.2
Repro:

Add a dependency on the crate async-global-executor-1.4.2 (https://github.com/async-rs/async-global-executor/blob/master/Cargo.toml).

That has this section in the `Cargo.toml` file:

```
[features]
default = ["async-io"]
tokio02 = ["tokio02-crate"]
tokio03 = ["tokio03-crate"]

# ...

[dependencies.tokio02-crate]
package = "tokio"
version = "^0.2"
optional = true
default-features = false
features = ["rt-core"]

[dependencies.tokio03-crate]
package = "tokio"
version = "^0.3"
optional = true
default-features = false
features = ["rt"]
```

However, the file generated by `cargo raze` (v0.6.1, installed a few moments ago) looks like:

```
# boilerplate omitted

# buildifier: leave-alone
rust_library(
    name = "async_global_executor",
    crate_type = "lib",
    deps = [
        "@raze__async_executor__1_3_0//:async_executor",
        "@raze__async_io__1_1_10//:async_io",
        "@raze__futures_lite__1_11_2//:futures_lite",
        "@raze__num_cpus__1_13_0//:num_cpus",
        "@raze__once_cell__1_4_1//:once_cell",
        "@raze__tokio__0_2_22//:tokio",
    ],
    srcs = glob(["**/*.rs"]),
    crate_root = "src/lib.rs",
    edition = "2018",
    rustc_flags = [
        "--cap-lints=allow",
    ],
    version = "1.4.2",
    tags = [
        "cargo-raze",
        "manual",
    ],
    crate_features = [
        "async-io",
        "default",
        "tokio02",
        "tokio02-crate",
    ],
    aliases = {
        "@raze__tokio__0_2_22//:tokio": "tokio03_crate",
    },
)
```

Note the `aliases` section at the bottom, and how `tokio_0_2_22` points to `tokio03_crate`, resulting in the following build error:

```
INFO: From Compiling Rust lib async_global_executor v1.4.2 (1 files):
error[E0433]: failed to resolve: use of undeclared type or module `tokio02_crate`
  --> external/raze__async_global_executor__1_4_2/src/lib.rs:69:22
   |
69 | static TOKIO02: Lazy<tokio02_crate::runtime::Handle> = Lazy::new(|| {
   |                      ^^^^^^^^^^^^^ use of undeclared type or module `tokio02_crate`

error[E0433]: failed to resolve: use of undeclared type or module `tokio02_crate`
  --> external/raze__async_global_executor__1_4_2/src/lib.rs:70:42
   |
70 |     let mut rt = tokio02_crate::runtime::Builder::new()
   |                                          ^^^^^^^ not found in `tokio02_crate::runtime`
```

Changing that `tokio03_crate` to `tokio02_crate` is enough for the build to succeed.

### LINKED ISSUE #270: Incorrect `rand` alias generation for `libsecp256k1` crate
Add `libsecp256k1 = "0.3.5"` to the dependencies section of the `Cargo.toml` and run `cargo raze`.

With v0.6.1 this generates `libsecp256k1-0.3.5.BUILD.bazel` with the following entry (hopefully irrelevant parts removed):

```
...

rust_library(
    name = "secp256k1",
    crate_type = "lib",

    ...
    
    aliases = {
        "@raze__rand__0_7_3//:rand": "rand_test",
    },
)
```

Note the `aliases` entry. This causes `rand` to appear as `rand_test` when building, and you get the following error:

```
% bazel build //cargo:libsecp256k1
...
INFO: From Compiling Rust lib secp256k1 v0.3.5 (15 files):
error[E0432]: unresolved import `rand`
  --> external/raze__libsecp256k1__0_3_5/src/lib.rs:34:5
   |
34 | use rand::Rng;
   |     ^^^^ use of undeclared type or module `rand`

error: aborting due to previous error
```

Edit the generated build file to remove or comment out the incorrect alias and the crate can build.

GregBowyer:
There are a range of subtle and irritating bugs that stem from
attempting to work with renames from the package set provided by cargo
metadata.

Fortunately, recent versions of cargo metadata provide a limited version
of the resolve nodes, both as the original node ids and as a newer set
providing both rename and targetting information.

As such we can (hopefully) simplify crate -> dep node resolution, remove
a bunch of subtle aliasing bugs.

Fixes #241, fixes #269, fixes #270, resolves #144, resolves #187

### DIFF
--- a/impl/src/context.rs
+++ b/impl/src/context.rs
@@ -12,7 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::path::PathBuf;
+use std::{
+  collections::BTreeSet,
+  hash::{Hash, Hasher},
+  path::PathBuf,
+};
 
 use crate::settings::CrateSettings;
 use semver::Version;
@@ -29,10 +33,23 @@ pub struct BuildableDependency {
   pub is_proc_macro: bool,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+#[derive(Debug, Clone, Eq, PartialOrd, Ord, Serialize)]
 pub struct DependencyAlias {
   pub target: String,
   pub alias: String,
+}
+
+// We only want equality for the first member as it can have multiple names pointing at it.
+impl Hash for DependencyAlias {
+  fn hash<H: Hasher>(&self, state: &mut H) {
+    self.target.hash(state);
+  }
+}
+
+impl PartialEq for DependencyAlias {
+  fn eq(&self, other: &Self) -> bool {
+    self.target == other.target
+  }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize)]
@@ -81,7 +98,7 @@ pub struct SourceDetails {
   pub git_data: Option<GitRepo>,
 }
 
-#[derive(Debug, Clone, Serialize, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Default, Debug, Clone, Serialize, PartialEq, Eq, PartialOrd, Ord)]
 pub struct CrateDependencyContext {
   pub dependencies: Vec<BuildableDependency>,
   pub proc_macro_dependencies: Vec<BuildableDependency>,
@@ -92,7 +109,7 @@ pub struct CrateDependencyContext {
   // build_data_dependencies can only be set when using cargo-raze as a library at the moment.
   pub build_data_dependencies: Vec<BuildableDependency>,
   pub dev_dependencies: Vec<BuildableDependency>,
-  pub aliased_dependencies: Vec<DependencyAlias>,
+  pub aliased_dependencies: BTreeSet<DependencyAlias>,
 }
 
 impl CrateDependencyContext {
@@ -110,7 +127,7 @@ impl CrateDependencyContext {
 pub struct CrateTargetedDepContext {
   pub target: String,
   pub deps: CrateDependencyContext,
-  pub conditions: Vec<String>,
+  pub platform_targets: Vec<String>,
 }
 
 #[derive(Debug, Clone, Serialize)]
--- a/impl/src/metadata.rs
+++ b/impl/src/metadata.rs
@@ -438,7 +438,7 @@ impl RazeMetadataFetcher {
           let version = info.req();
           let src_dir = self.fetch_crate_src(cargo_dir.as_ref(), &name, version)?;
           checksums.insert(
-            package_ident(name, version),
+            package_ident(name, &version),
             self.fetch_crate_checksum(name, version)?,
           );
           if let Some(dirname) = src_dir.file_name() {
--- a/impl/src/planning.rs
+++ b/impl/src/planning.rs
@@ -20,7 +20,7 @@ use anyhow::Result;
 use cargo_lock::Lockfile;
 
 use crate::{
-  context::{CrateContext, WorkspaceContext},
+  context::{CrateContext, DependencyAlias, WorkspaceContext},
   metadata::RazeMetadata,
   settings::RazeSettings,
   util::PlatformDetails,
@@ -32,8 +32,13 @@ use subplanners::WorkspaceSubplanner;
 /// A ready-to-be-rendered build, containing renderable context for each crate.
 #[derive(Debug)]
 pub struct PlannedBuild {
+  /// The overall context for this workspace
   pub workspace_context: WorkspaceContext,
+  /// The creates to build for
   pub crate_contexts: Vec<CrateContext>,
+  /// And aliases that are defined at the workspace root
+  pub workspace_aliases: Vec<DependencyAlias>,
+  /// The version lock used if present
   pub lockfile: Option<Lockfile>,
 }
 
@@ -89,6 +94,7 @@ mod tests {
   use super::*;
   use cargo_metadata::PackageId;
   use indoc::indoc;
+  use itertools::Itertools;
   use semver::{Version, VersionReq};
 
   fn dummy_resolve_dropping_metadata() -> RazeMetadata {
@@ -245,22 +251,22 @@ mod tests {
     assert!(krate_position.is_some());
 
     // Get crate context using computed position
-    let krate_context = crates_with_aliased_deps[krate_position.unwrap()].clone();
+    let crate_ctx = crates_with_aliased_deps[krate_position.unwrap()].clone();
 
     // There are two default dependencies for cargo-raze-alias-test, log^0.4 and log^0.3
-    // However, log^0.3 is aliased to old_log_ while log^0.4 isn't aliased. Therefore, we
-    // should only see one aliased dependency (log^0.3 -> old_log_) which shows that the
+    // However, log^0.3 is aliased to old_log while log^0.4 isn't aliased. Therefore, we
+    // should only see one aliased dependency (log^0.3 -> old_log) which shows that the
     // name and semver matching for aliased dependencies is working correctly
-    assert!(krate_context.default_deps.aliased_dependencies.len() == 1);
-    assert_eq!(
-      krate_context.default_deps.aliased_dependencies[0].target,
-      "@raze_test__log__0_3_9//:log"
-    );
-    assert_eq!(
-      krate_context.default_deps.aliased_dependencies[0].alias,
-      "old_log_"
-    );
+    let dep = crate_ctx
+      .default_deps
+      .aliased_dependencies
+      .iter()
+      .exactly_one()
+      .unwrap();
+    assert_eq!(dep.target, "@raze_test__log__0_3_9//:log");
+    assert_eq!(dep.alias, "old_log");
   }
+
   #[test]
   fn test_plan_build_produces_proc_macro_dependencies() {
     let mut settings = dummy_raze_settings();
@@ -423,7 +429,6 @@ mod tests {
     let context = planned_build
       .crate_contexts
       .iter()
-      .inspect(|x| println!("{}{}", x.pkg_name, x.pkg_version))
       .find(|ctx| ctx.pkg_name == "some-binary-crate" && ctx.pkg_version == version)
       .unwrap();
 
--- a/impl/src/planning/subplanners.rs
+++ b/impl/src/planning/subplanners.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use std::{
-  collections::{HashMap, HashSet},
+  collections::{BTreeSet, HashMap, HashSet},
   io, iter,
   path::{Path, PathBuf},
   str::FromStr,
@@ -21,7 +21,7 @@ use std::{
 
 use anyhow::{anyhow, Context, Result};
 use cargo_lock::SourceId;
-use cargo_metadata::{DependencyKind, Node, Package};
+use cargo_metadata::{DepKindInfo, DependencyKind, Node, Package};
 use cargo_platform::Platform;
 use itertools::Itertools;
 
@@ -43,39 +43,11 @@ use super::{
   PlannedBuild,
 };
 
-/// A set of named dependencies (without version) derived from a package manifest.
-struct DependencyNames {
-  // Dependencies that are required for all buildable targets of this crate
-  normal_dep_names: Vec<String>,
-  // Dependencies that are required for build script only
-  build_dep_names: Vec<String>,
-  // Dependencies that are required for tests
-  dev_dep_names: Vec<String>,
-  // Dependencies that have been renamed and need to be aliased in the build rule
-  aliased_dep_names: HashMap<String, (semver::VersionReq, String)>,
-}
+/// Named type to reduce declaration noise for deducing the crate contexts
+type CrateContextProduction = (Vec<CrateContext>, Vec<DependencyAlias>);
 
-// TODO(acmcarther): Remove this struct -- move it into CrateContext.
-/// A set of dependencies that a crate has broken down by type.
-struct DependencySet {
-  // Dependencies that are required for all buildable targets of this crate
-  normal_deps: Vec<BuildableDependency>,
-  proc_macro_deps: Vec<BuildableDependency>,
-  // Dependencies that are required for build script only
-  build_deps: Vec<BuildableDependency>,
-  // Dependencies that proc macros and are required for the build script only
-  build_proc_macro_deps: Vec<BuildableDependency>,
-  // Dependencies that are required for tests
-  dev_deps: Vec<BuildableDependency>,
-  // Dependencies that have been renamed and need to be aliased in the build rule
-  aliased_deps: Vec<DependencyAlias>,
-}
-
-/// A set of dependencies that a crate has for a specific target/cfg
-struct TargetedDependencySet {
-  target: String,
-  dependencies: DependencySet,
-}
+/// Utility type alias to reduce declaration noise
+type DepProduction = HashMap<Option<String>, CrateDependencyContext>;
 
 /// An internal working planner for generating context for an individual crate.
 struct CrateSubplanner<'planner> {
@@ -103,11 +75,12 @@ impl<'planner> WorkspaceSubplanner<'planner> {
   /// Produces a planned build using internal state.
   pub fn produce_planned_build(&self) -> Result<PlannedBuild> {
     // Produce planned build
-    let crate_contexts = self.produce_crate_contexts()?;
+    let (crate_contexts, workspace_aliases) = self.produce_crate_contexts()?;
 
     Ok(PlannedBuild {
       workspace_context: self.produce_workspace_context(),
       crate_contexts,
+      workspace_aliases,
       lockfile: self.metadata.lockfile.clone(),
     })
   }
@@ -155,7 +128,7 @@ impl<'planner> WorkspaceSubplanner<'planner> {
     &self,
     node: &Node,
     catalog: &CrateCatalog,
-  ) -> Option<Result<CrateContext>> {
+  ) -> Option<Result<(CrateContext, bool)>> {
     let own_crate_catalog_entry = catalog.entry_for_package_id(&node.id)?;
     let own_package = own_crate_catalog_entry.package();
 
@@ -193,7 +166,11 @@ impl<'planner> WorkspaceSubplanner<'planner> {
       sha256: &checksum_opt.map(|c| c.to_owned()),
     };
 
-    Some(crate_subplanner.produce_context(&self.metadata.cargo_workspace_root))
+    let res = crate_subplanner
+      .produce_context(&self.metadata.cargo_workspace_root)
+      .map(|x| (x, own_crate_catalog_entry.is_workspace_crate()));
+
+    Some(res)
   }
 
   fn crate_settings(&self, package: &Package) -> Result<Option<&CrateSettings>> {
@@ -225,8 +202,8 @@ impl<'planner> WorkspaceSubplanner<'planner> {
   }
 
   /// Produces a crate context for each declared crate and dependency.
-  fn produce_crate_contexts(&self) -> Result<Vec<CrateContext>> {
-    self
+  fn produce_crate_contexts(&self) -> Result<CrateContextProduction> {
+    let contexts = self
       .crate_catalog
       .metadata
       .resolve
@@ -236,25 +213,64 @@ impl<'planner> WorkspaceSubplanner<'planner> {
       .iter()
       .sorted_by_key(|n| &n.id)
       .filter_map(|node| self.create_crate_context(node, &self.crate_catalog))
-      .collect::<Result<Vec<CrateContext>>>()
+      .collect::<Result<Vec<_>>>()?;
+
+    let root_ctxs = contexts
+      .iter()
+      .filter_map(|(ctx, is_workspace)| match is_workspace {
+        true => Some(ctx.default_deps.aliased_dependencies.clone()),
+        false => None,
+      })
+      .flatten()
+      .collect::<BTreeSet<_>>();
+
+    let contexts = contexts.into_iter().map(|(ctx, _)| ctx).collect_vec();
+    let aliases = self.produce_workspace_aliases(root_ctxs, &contexts);
+
+    Ok((contexts, aliases))
+  }
+
+  fn produce_workspace_aliases(
+    &self,
+    root_dep_aliases: BTreeSet<DependencyAlias>,
+    all_packages: &[CrateContext],
+  ) -> Vec<DependencyAlias> {
+    let renames = root_dep_aliases
+      .iter()
+      .map(|rename| (&rename.target, &rename.alias))
+      .collect::<HashMap<_, _>>();
+
+    all_packages
+      .iter()
+      .filter(|to_alias| to_alias.lib_target_name.is_some())
+      .filter(|to_alias| to_alias.is_workspace_member_dependency)
+      .flat_map(|to_alias| {
+        let pkg_name = to_alias.pkg_name.replace("-", "_");
+        let target = format!("{}:{}", &to_alias.workspace_path_to_crate, &pkg_name);
+        let alias = renames
+          .get(&target)
+          .map(|x| x.to_string())
+          .unwrap_or(pkg_name);
+        let dep_alias = DependencyAlias { alias, target };
+
+        to_alias
+          .raze_settings
+          .extra_aliased_targets
+          .iter()
+          .map(move |extra_alias| DependencyAlias {
+            alias: extra_alias.clone(),
+            target: format!("{}:{}", &to_alias.workspace_path_to_crate, extra_alias),
+          })
+          .chain(std::iter::once(dep_alias))
+      })
+      .sorted()
+      .collect_vec()
   }
 }
 
 impl<'planner> CrateSubplanner<'planner> {
   /// Builds a crate context from internal state.
   fn produce_context(&self, cargo_workspace_root: &Path) -> Result<CrateContext> {
-    let (
-      DependencySet {
-        build_deps,
-        build_proc_macro_deps,
-        proc_macro_deps,
-        dev_deps,
-        normal_deps,
-        aliased_deps,
-      },
-      targeted_deps,
-    ) = self.produce_deps()?;
-
     let package = self.crate_catalog_entry.package();
 
     let manifest_path = PathBuf::from(&package.manifest_path);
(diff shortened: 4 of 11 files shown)


GregBowyer:
@UebelAndre sorry if this stomps on anything you have in progress

UebelAndre:
It undoubtably will 😢. While I do like the changes here, I would greatly appreciate it if I could get #276 through first. I've had to rebase and resolve conflicts a few times already ☹️ 

If I could just get it reviewed, I can promise to have it updated as fast as I possibly can to make sure it gets closed as soon as possible 😃

impl/src/planning/subplanners.rs
```diff
@@ -279,6 +267,8 @@ impl<'planner> WorkspaceSubplanner<'planner> {
       }
     }
 
+    let mut root_ctx = None;
```
GregBowyer, line 270: This bit, and its devil children I dont fully like, it would be nice maybe to just grab the root contect without having to discover it through resolving deps.

I could not think of or find a way to do it, its probably ok for now, and maybe a better future approach might be to move to guppy as a dep

impl/src/planning/subplanners.rs
```diff
@@ -323,48 +350,30 @@ impl<'planner> CrateSubplanner<'planner> {
...
-        conditions: generate_bazel_conditions(&target_triples)?,
-      });
-    }
-
-    filtered_deps.sort();
+    let mut targeted_deps = deps
+      .into_iter()
+      .map(|(target, deps)| {
+        let target = target.unwrap().to_string();
+        let platform_targets = get_matching_bazel_triples(&target, &self.settings.targets)?
```
GregBowyer, line 366: I noticed in working with this, the targetting functions are a bit scattered and not really that reused.

As such I removed a bunch of the additional functions, smushing things down towards a single function, and in avoiding allocation-o-rama (which we do a bit frequently in cargo-raze) made it return an iterator.

The part for the generation of bazel conditions I moved into the template itself, as it seemed like a lot of code was written just to essentially prepend a string onto the platform targets.

impl/src/planning/subplanners.rs
```diff
@@ -412,32 +414,21 @@ impl<'planner> CrateSubplanner<'planner> {
...
+    let mut dep_production: HashMap<Option<&'planner str>, CrateDependencyContext> = HashMap::new();
 
     let all_skipped_deps = self
       .crate_settings
       .iter()
       .flat_map(|pkg| pkg.skipped_deps.iter())
       .collect::<HashSet<_>>();
 
-    for dep_id in &self.node.dependencies {
+    for dep in &self.node.deps {
```
GregBowyer, line 427: This wonderful part of the metadata gives us *both* _renames_ and _targets_. Irritatingly its still a little tricky to detect renames fully (we need this to finally deduce the aliases - see `is_renamed`)

If https://github.com/rust-lang/cargo/issues/7289 gets solved then a lot of the left-over rename detection code can get murdered

impl/src/planning/subplanners.rs
```diff
@@ -446,182 +437,163 @@ impl<'planner> CrateSubplanner<'planner> {
...
-      };
-
-      if build_dep_names.contains(&dep_package.name) {
-        if buildable_dependency.is_proc_macro {
-          dep_set
-            .build_proc_macro_deps
-            .push(buildable_dependency.clone());
-        } else {
-          dep_set.build_deps.push(buildable_dependency.clone());
+      // TODO(GregBowyer): Reimplement what cargo does to detect bad renames
```
GregBowyer, line 440: The problem manifests as this

```toml
[dependencies]
bytes_new = { version = "0.3.0", package = "bytes" }
bytes_old = { version = "0.3.0", package = "bytes" }
```

Strictly this is an error. Right now cargo metadata will basically lose both `bytes` deps from the resolve graph :gruntle: but a `cargo build` will scream about having the same basic package with multiple names.

I dont have a good solution right now (I dont think we should go back to trying to essentially emulate the resolve graph like before), so I guess a todo :/

impl/src/planning/subplanners.rs
```diff
@@ -446,182 +437,163 @@ impl<'planner> CrateSubplanner<'planner> {
...
+      .dependencies
+      .iter()
+      .filter(|x| x.name == dep_package.name)
+      .filter(|x| x.req.matches(&dep_package.version))
+      .filter(|x| x.source == dep_package.source.as_ref().map(|src| src.to_string()))
+      .find(|x| x.rename.is_some())
+      .map_or(false, |x| x.rename.is_some())
+  }
 
+  fn is_dep_targetted(&self, target: Option<&str>) -> bool {
```
GregBowyer, line 573: NIT: Maybe add a comment in keeping with the other functions?

GregBowyer:
> It undoubtably will cry. While I do like the changes here, I would greatly appreciate it if I could get #276 through first. I've had to rebase and resolve conflicts a few times already frowning_face
> 
> If I could just get it reviewed, I can promise to have it updated as fast as I possibly can to make sure it gets closed as soon as possible smiley

I can hold on your changes and work with that, I think we get #127 in then #276 and then see how bad that breaks this.

UebelAndre:
Yeah, I can rebase on top of that when that goes in.

UebelAndre:
https://github.com/google/cargo-raze/pull/276 is rebased on top of https://github.com/google/cargo-raze/pull/127 now

GregBowyer:
@acmcarther thoughts?

acmcarther:
Was holding off on review because the other PR is in review, but that's taking a while (because I'm picky or something).

I'll review this now.

GregBowyer:
Nah we are all picky ;P

I am sure this one is going to need more work anyhow there are things I cant think of a better way for that I dont like

impl/src/planning/subplanners.rs
```diff
@@ -446,182 +441,163 @@ impl<'planner> CrateSubplanner<'planner> {
...
 
-        // Legacy behavior
-        if let Some(platform_details) = &self.platform_details {
-          if let Some(settings_target) = &self.settings.target {
-            let platform = Platform::from_str(&target_str)?;
+  /// Test to see if a dep has been renamed
+  ///
+  /// Currently cargo-metadata provides rename detection in a few places, we take the names
+  /// from the resolution of a package
+  fn is_renamed(&self, dep_package: &Package) -> bool {
```
acmcarther, line 564: Does it make sense to precompute some of these new properties? I'm lost in the details here, but part of the idea with the crate catalog was to make this type of lookup more efficient.

acmcarther (changes requested):
Not ideal that the formatter changes are co-mingled with the behavior changes. A bit harder to review.

I am so unqualified to review these PRs at this point. I don't remember much about the implementation and anyway it has changed a lot since I wrote significant code here.

Oh well

impl/src/planning.rs
```diff
@@ -36,8 +36,13 @@ use subplanners::WorkspaceSubplanner;
 /** A ready-to-be-rendered build, containing renderable context for each crate. */
 #[derive(Debug)]
 pub struct PlannedBuild {
+  /// The overall context for this workspace
```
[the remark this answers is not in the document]
GregBowyer: Rustdoc thing, but I was noticing in the ther review most comments in carg-raze are f the form `/** comment */` and so I should make the comments that I think (for the consistency)

impl/src/planning/subplanners.rs
GregBowyer: Yeah, I kinda ignore it in cargo-raze because the tool runs for me a few times a week, but when calling three consecutive functions seems to produce vector and string of each other it gets a bit much ;)

GregBowyer:
> Not ideal that the formatter changes are co-mingled with the behavior changes. A bit harder to review.
> 
> I am so unqualified to review these PRs at this point. I don't remember much about the implementation and anyway it has changed a lot since I wrote significant code here.
> 
> Oh well

I would only bother with the first commit in the PR, I can just pull the second commit (the reformatting) and we can do a round of that after all crazy PRs get merged

impl/src/planning/subplanners.rs
GregBowyer: Humm yeah might work, and might be a good way to make the find the root sillyness go away.

impl/src/planning/subplanners.rs
```diff
@@ -290,30 +280,67 @@ impl<'planner> WorkspaceSubplanner<'planner> {
...
+      .iter()
+      .map(|rename| (&rename.target, &rename.alias))
+      .collect::<HashMap<_, _>>();
 
-    Ok(contexts.iter().flat_map(|i| i.clone()).collect())
+    all_packages
+      .iter()
+      .filter(|to_alias| to_alias.is_root_dependency && to_alias.lib_target_name.is_some())
+      .flat_map(|to_alias| {
+        let pkg_name = sanitize_ident(&to_alias.pkg_name);
```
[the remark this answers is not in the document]
GregBowyer: From what I have seen it typically is providing its not too crazy but this is pretty much at the limit and might need to just be a function directly.

UebelAndre:
Crazy PR has been merged. No more crazy PRs from me for the foreseeable future 😅

UebelAndre:
In hindsight, this should have been merged before any crazy PRs 😞 Sorry about pushing for mine over this one.

@GregBowyer do you have bandwidth in the foreseeable future to revive this effort 😅?

GregBowyer:
I have time to pick this up again

GregBowyer:
Ok restarting the patch and the review.

This should be the changes (and just the changes) from the original code. The unit tests will be redone by me to cover the bugs.

There is one test that fails but that is just down to how the test itself is constructed than any logic failure.

I am going to try a few runs locally to see if it does the right thing on a live codebase.

@acmcarther @UebelAndre If we review and get this in can we either call this PR dead or not let it get overridden, I am get fatigued by constantly rewriting this logic (this is the 4th time of writing this now :S)

UebelAndre:
> @acmcarther @UebelAndre If we review and get this in can we either call this PR dead or not let it get overridden, I am get fatigued by constantly rewriting this logic (this is the 4th time of writing this now :S)

@GregBowyer I don't think I understand your request but I'm happy to do whatever I can to make this easier for you 😄 I owe you for being so flexible with my large PR. Just let me know how I can help

impl/src/rendering/templates/partials/targeted_dependencies.template
```diff
@@ -1,8 +1,8 @@
 {% for targeted_dep in crate.targeted_deps %} + selects.with_or({
         # {{ targeted_dep.target }}
         (
-{%- for condition in targeted_dep.conditions %}
-            "{{ condition }}",
+{%- for platform_target in targeted_dep.platform_targets %}
+          "@rules_rust//rust/platform:{{platform_target}}",
```
UebelAndre, line 5: This should use a templated workspace name. ie:
https://github.com/google/cargo-raze/blob/eee9428916139e4329d25d5b1c42de120486463f/impl/src/rendering/templates/crate.BUILD.template#L10

impl/src/util.rs
```diff
@@ -436,48 +400,4 @@ mod tests {
       get_builtin_target_by_triple(triple).unwrap();
     }
   }
-
-  #[test]
-  fn generate_condition_strings() {
```
UebelAndre, line 441: Why are these tests not necessary?

UebelAndre (changes requested):

impl/src/rendering/templates/partials/targeted_dependencies.template
GregBowyer: yep noted when I was updating `Cargo.toml` to the latest

impl/src/util.rs
GregBowyer: The functions for doing targets were broken out into three steps that essentially needed each other to actually work.

I made it one function since the API was scattered, in the process the tests that worked on the intermetidate steps are harder (and probably dont actually need testing as a unit) to test.

As I see it the function should go to being `fn(cargo_target) -> bazel_target`, rather than juggling through several functions. I can beef up the tests for targeting if wanted to cover the testing gap.

GregBowyer:
FWIW assuming I fixerate tests (and add the ones from the 3rd version of this PR) this works for us for our main `Cargo.toml`

UebelAndre:
Hey @GregBowyer just FYI, I think https://github.com/google/cargo-raze/pull/363 introduced some conflicts but they're just Clippy fixes (which is now enabled in CI).

GregBowyer:
@UebelAndre I have bumped to latest (untested), want to do me a review?

UebelAndre:
Sure 😄 I'll take a look tonight

impl/src/planning.rs
```diff
@@ -245,22 +251,22 @@ mod tests {
     assert!(krate_position.is_some());
 
     // Get crate context using computed position
-    let krate_context = crates_with_aliased_deps[krate_position.unwrap()].clone();
+    let krate_ctx = crates_with_aliased_deps[krate_position.unwrap()].clone();
```
UebelAndre, line 254: nit: any reason not to go with `crate_ctx` (understanding `crate` is not usable)?

impl/src/planning/subplanners.rs
```diff
@@ -236,25 +213,63 @@ impl<'planner> WorkspaceSubplanner<'planner> {
...
+    let renames = root_dep_aliases
+      .iter()
+      .map(|rename| (&rename.target, &rename.alias))
+      .collect::<HashMap<_, _>>();
+
+    all_packages
+      .iter()
+      .filter(|to_alias| to_alias.lib_target_name.is_some())
+      .flat_map(|to_alias| {
+        let pkg_name = util::sanitize_ident(&to_alias.pkg_name);
```
UebelAndre, line 247: I feel like the sanitize is worth keeping but wanted to share https://github.com/google/cargo-raze/commit/9d74090f28cf7d546414a0591205efa5219c9bf2 here in case that's relevant. I think `sanatize_indent` might need to be updated

impl/src/planning/subplanners.rs
```diff
@@ -264,59 +279,40 @@ impl<'planner> CrateSubplanner<'planner> {
...
-          &self.settings.rust_rules_workspace_name,
-          &target_triples,
-        )?,
-      });
-    }
+    // Take the default deps that are not bound to platform targets
+    let default_deps = deps.remove(&None).unwrap_or_default();
 
-    filtered_deps.sort();
+    // Build a list of dependencies while addression a potential whitelist of target triples
```
UebelAndre, line 294:
```suggestion
    // Build a list of dependencies while addressing a potential allowlist of target triples
```

impl/src/planning/subplanners.rs
```diff
@@ -439,225 +426,208 @@ impl<'planner> CrateSubplanner<'planner> {
...
       .flat_map(|pkg| pkg.skipped_deps.iter())
       .collect::<HashSet<_>>();
 
-    for dep_id in &self.node.dependencies {
+    // This wonderful part of the metadata gives us both renames and targets.
+    //
+    // Irritatingly its still a little tricky to detect renames fully, we need this to finally
+    // deduce the aliases - see `is_renamed`.
+    //
+    // If rust-lang/cargo#7289 gets solved then a lot of the left-over rename detection code can
```
UebelAndre, line 444:
```suggestion
    // If https://github.com/rust-lang/cargo/issues/7289 gets solved then a lot of the left-over rename detection code can
```

This is optional but I personally prefer full links unless it's to the same repo as the project.

impl/src/planning/subplanners.rs
```diff
@@ -439,225 +426,208 @@ impl<'planner> CrateSubplanner<'planner> {
...
-      build_dep_names: Vec::new(),
-      dev_dep_names: Vec::new(),
-      normal_dep_names: Vec::new(),
-      aliased_dep_names: HashMap::new(),
+    use DependencyKind::*;
+    match dep.kind {
+      Build if is_proc_macro => dep_set.build_proc_macro_dependencies.push(build_dep),
+      Build => dep_set.build_dependencies.push(build_dep),
+      Development => dep_set.dev_dependencies.push(build_dep),
+      // TODO(GregBowyer): Is a proc-macro -sys crate a thing?
```
UebelAndre, line 520: Since writing this, was did you ever find out? I don't think it's a thing but that's likely a result of my limited experience 😅

impl/src/planning/subplanners.rs
```diff
@@ -439,225 +426,208 @@ impl<'planner> CrateSubplanner<'planner> {
...
+          Some((platform_details, settings_target)) => {
             // Skip this dep if it doesn't match our platform attributes
-            if !platform.matches(settings_target, platform_details.attrs().as_ref()) {
-              continue;
-            }
+            // UNWRAP: It is reasonable to assume cargo is not giving us odd platform strings
+            let platform = Platform::from_str(platform).unwrap();
+            platform.matches(settings_target, platform_details.attrs())
           }
+          // TODO (GregBowyer): Make this an enum
```
UebelAndre, line 596: Do you think it'd be a lot of work to make this enum? It sounds like a far better solution and I don't think this is used too widely. Can punt but I think it'd be some great polish.

impl/src/util.rs
```diff
@@ -107,81 +109,46 @@ pub fn is_bazel_supported_platform(target: &str) -> (bool, bool) {
...
+      });
+
+      match res {
+        true => Some(triple),
         false => None,
       }
     })
-    .collect();
+    .filter(move |x| {
+      whitelist
```
UebelAndre, line 139:
```suggestion
      allowlist
```

UebelAndre (changes requested):
A couple of comments and suggestions but otherwise sounds pretty good. I'm going to need to take another day to page some of the nuances here back into my brain and do another pass but still pretty excited for the direction!

UebelAndre:
Since migrating most things to Bazel, ensuring the examples are up to date has not been added to CI. Would you be able to run `bazel run //tools:examples_raze` and commit the results? I suspect a few things might change based these changes.

impl/src/context.rs
```diff
@@ -29,9 +30,11 @@ pub struct BuildableDependency {
   pub is_proc_macro: bool,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+#[derive(Debug, Clone, Eq, PartialOrd, Ord, Serialize, Derivative)]
+#[derivative(Hash, PartialEq)]
```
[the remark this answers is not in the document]
GregBowyer: Yeah I can kill it, I doubt it adds much but its probably not that hard to hand roll the `Hash` and `PartialEq`

impl/src/planning.rs
GregBowyer: None, I was only mildly shortening the name that was already there.

impl/src/planning/subplanners.rs
GregBowyer: Oops I thought I factored that in when I rebased so good catch.

impl/src/planning/subplanners.rs
GregBowyer: I have never seen one in the wild although `-sys` crates are by convention.

I will pull it, I think if it happens it will be a curious bug in the future to figure out (even if such a thing existed would it be a host or target compile for instance)

GregBowyer:
I will pull in the comments (all sane I think) make sure that test failure is not something sinister.

I think I should also take another scan through the bug list to see if there are any other bugs that this solves.

impl/src/util.rs
```diff
@@ -107,81 +109,46 @@ pub fn is_bazel_supported_platform(target: &str) -> (bool, bool) {
...
+) -> Result<impl Iterator<Item = &'static str> + 'a> {
+  let expression = match target.starts_with("cfg(") {
+    true => Expression::parse(target),
+    false => Expression::parse(&format!("cfg(target = \"{}\")", target)),
+  }?;
+
+  let triples = SUPPORTED_PLATFORM_TRIPLES
     .iter()
-    .filter_map(|triple| {
+    .filter_map(move |triple| {
```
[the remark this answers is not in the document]
GregBowyer: Its to let rust know that `expression` is only needed used in the closure when we are filtering the targets, as such it needs to exist for the length of the iterator as bound by `'static` and `'a`. That is the `filter_map` takes ownership of `expression` and will not share it.

This is similar for `allowlist` which is given to the `filter` closure. Since we are not sharing these references we can move them where they are needed. It is important to note that this is not an optimisation but is actually the ownership rules.

impl/src/util.rs
```diff
@@ -49,20 +51,33 @@ static SUPPORTED_PLATFORM_TRIPLES: &[&str] = &[
   "x86_64-unknown-freebsd",
 ];
 
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
```
UebelAndre, line 54: Haven't done a full review yet but are there any paths where this might require `Serialize` and `Deserialize` as well?

UebelAndre: Specifically asking in regards to https://github.com/google/cargo-raze/pull/415 but that's likely out of scope for that change and something I'll go update 😄

GregBowyer:
No need it's not going towards the templates so won't serialise

On Fri, Apr 23, 2021, at 3:08 PM, UebelAndre wrote:
> 

> *@UebelAndre* commented on this pull request.

> 
> In impl/src/util.rs <https://github.com/google/cargo-raze/pull/282#discussion_r619518891>:

> > @@ -49,20 +51,33 @@ static SUPPORTED_PLATFORM_TRIPLES: &[&str] = &[
   "x86_64-unknown-freebsd",
 ];
 
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]

impl/src/util.rs
UebelAndre: https://github.com/google/cargo-raze/pull/282#issuecomment-825969996

impl/src/planning/subplanners.rs
```diff
@@ -439,225 +427,205 @@ impl<'planner> CrateSubplanner<'planner> {
...
+      .filter(|x| x.source == dep_package.source.as_ref().map(|src| src.to_string()))
+      .find(|x| x.rename.is_some())
+      .map_or(false, |x| x.rename.is_some())
+  }
 
-        // Legacy behavior
-        if let Some(platform_details) = &self.platform_details {
-          if let Some(settings_target) = &self.settings.target {
-            let platform = Platform::from_str(&target_str)?;
+  fn is_dep_targetted(&self, target: Option<&String>) -> bool {
```
UebelAndre, line 580: nit: Can you add a small doc for this?

impl/src/planning/subplanners.rs
```diff
@@ -439,225 +427,205 @@ impl<'planner> CrateSubplanner<'planner> {
...
-      .iter()
-      .map(|(target, deps)| TargetedDependencySet {
-        target: target.clone(),
-        dependencies: self._produce_deps(deps).unwrap(),
-      })
-      .collect::<Vec<TargetedDependencySet>>();
-
-    Ok((self._produce_deps(&default_deps)?, targeted_set))
-  }
+  fn process_dep(
```
UebelAndre, line 499: nit: add a doc here as well

UebelAndre (changes requested):
Just another small pass. I'll take some time over the weekend to do a thorough review. The thing I want to look into is the test coverage for this. What's your confidence level that there's enough here to prevent a regression? (again, need to look, there might be enough already but I don't have that context paged in my brain right now 😅)

impl/src/metadata.rs
```diff
@@ -438,7 +438,7 @@ impl RazeMetadataFetcher {
           let version = info.req();
           let src_dir = self.fetch_crate_src(cargo_dir.as_ref(), &name, version)?;
           checksums.insert(
-            package_ident(name, version),
+            package_ident(name, &version),
```
UebelAndre, line 441: Was there an issue with this before?

impl/src/planning/subplanners.rs
```diff
@@ -155,7 +128,7 @@ impl<'planner> WorkspaceSubplanner<'planner> {
     &self,
     node: &Node,
     catalog: &CrateCatalog,
-  ) -> Option<Result<CrateContext>> {
+  ) -> Option<Result<(CrateContext, bool)>> {
```
UebelAndre, line 131: Is this not information that already lives (or could already live) in `CrateContext`? Admittedly that struct could use some cleaning but whether or not the crate is a workspace member seems like it makes sense to store in the context.
https://github.com/google/cargo-raze/blob/ee2effee5b6c95cfcfd5e496677ce61dcd35c472/impl/src/context.rs#L116-L153

impl/src/rendering/templates/partials/targeted_dependencies.template
UebelAndre:
```suggestion
          "@{{ rust_rules_workspace_name }}//rust/platform:{{platform_target}}",
```

impl/src/rendering/templates/partials/build_script.template
```diff
@@ -73,8 +73,8 @@ cargo_build_script(
     {% for targeted_dep in crate.targeted_deps %} + selects.with_or({
         # {{ targeted_dep.target }}
         (
-    {%- for condition in targeted_dep.conditions %}
-            "{{ condition }}",
+    {%- for platform_target in targeted_dep.platform_targets %}
+          "@rules_rust//rust/platform:{{platform_target}}",
```
UebelAndre, line 77:
```suggestion
          "@{{ rust_rules_workspace_name }}//rust/platform:{{platform_target}}",
```

impl/src/testing.rs
```diff
@@ -225,7 +225,7 @@ pub fn mock_remote_crate<'server>(
     let enc = flate2::write::GzEncoder::new(tar_gz, Compression::default());
     let mut tar = tar::Builder::new(enc);
     tar
-      .append_dir_all(package_ident(name, version), dir.as_ref().join("archive"))
+      .append_dir_all(package_ident(name, &version), dir.as_ref().join("archive"))
```
UebelAndre, line 228: Another question about using a reference. Just curious.

UebelAndre (changes requested):
Ok, a few more nits and one or two questions. But otherwise I think this is looking really good! 😄

UebelAndre:
@acmcarther this is looking near completion. Don't know if you're around to review it but it'd be great to get in. If not, maybe @dfreese can take a look? It'd probably be good to get your eyes on it too just so you're aware of the changes (sorry for the ping otherwise 😅)

impl/src/context.rs
```diff
@@ -29,12 +33,25 @@ pub struct BuildableDependency {
...
+  fn hash<H: Hasher>(&self, state: &mut H) {
+    self.target.hash(state);
+  }
+}
+
+impl PartialEq for DependencyAlias {
+  fn eq(&self, other: &Self) -> bool {
+    self.target == other.target
+  }
+}
```
dfreese, line 53: I've been bitten by some nasty bugs where hash or comparison functions don't cover the all of the data in the struct.

Is there another way to structure this, perhaps by using map from target string to alias (or &DependencyAlias)?

impl/src/planning/subplanners.rs
dfreese: +1 to TODO(#cargo-raze-issue): referencing this error.

dfreese (commented):
Haven't had a chance to make a full pass at this, but had one major question, based on past experience.

UebelAndre:
@GregBowyer friendly ping. I feel this is so close to going in 😄

UebelAndre:
@GregBowyer @dfreese another friendly ping 😅 Would love to get this merged and am wondering what outstanding action items are. Happy to help move this forward however I can.

UebelAndre:
Friendly ping again. Happy to help however I can to get this in but would love to see this merged 😄

impl/src/planning/subplanners.rs
```diff
@@ -264,59 +280,40 @@ impl<'planner> CrateSubplanner<'planner> {
...
-          &target_triples,
-        )?,
-      });
-    }
+    let lib_target_name = targets
+      .iter()
+      .find(|target| target.kind == "lib" || target.kind == "proc-macro")
+      .map(|target| target.name.clone());
+
+    let is_proc_macro = targets.iter().any(|target| target.kind == "proc_macro");
```
UebelAndre, line 288:
```suggestion
    let is_proc_macro = targets.iter().any(|target| target.kind == "proc-macro");
```

Aren't string literals fun? 😄

UebelAndre:
@dfreese is this not something you could apply the suggestions for and accept as is? I think it's in a good spot.

dfreese:
> @dfreese is this not something you could apply the suggestions for and accept as is? I think it's in a good spot.

https://github.com/google/cargo-raze/pull/282#discussion_r620890929

Not having all fields included as a part of a hash/comparison function causes some incredibly hard bugs to figure out, so I'd want to see if there's an alternative solution to that first.

UebelAndre:
> > @dfreese is this not something you could apply the suggestions for and accept as is? I think it's in a good spot.
> 
> https://github.com/google/cargo-raze/pull/282#discussion_r620890929
> 
> Not having all fields included as a part of a hash/comparison function causes some incredibly hard bugs to figure out, so I'd want to see if there's an alternative solution to that first.

Is that not something you or I could patch in this PR or in a follow up? I think @GregBowyer has put fourth a huge amount of effort implementing this PR at least 3 times. I'd hate to have this reimplemented a 4th time and there are other changes I think should go in.

PR #283, opened 2020-11-07
cli: implement `--version`

wchargin:
This patch adds a `--version` flag that causes `cargo-raze` to print
version info and exit. The functionality is built into Docopt, so it’s
just a matter of hooking it up to the `CARGO_PKG_VERSION` env var.

Test Plan:
When invoking the binary to test, make sure to pass an extra `raze` arg
so that docopt matches the `cargo raze` (not `cargo-raze`) pattern:

```
cargo run -- raze --version  # or...
cargo install --path . && cargo raze --version
```

Either invocation prints “cargo-raze 0.7.0” and a newline before exiting
successfully. This format is consistent with the `--version` behavior of
`cargo` and `rustc` themselves.

wchargin-branch: cli-version

### DIFF
--- a/impl/src/bin/cargo-raze.rs
+++ b/impl/src/bin/cargo-raze.rs
@@ -53,9 +53,11 @@ Usage:
     cargo raze [--verbose] [--quiet] [--color=<WHEN>] [--dryrun] [--cargo-bin-path=<PATH>] [--output=<PATH>]
     cargo raze <buildprefix> [--verbose] [--quiet] [--color=<WHEN>] [--dryrun] [--cargo-bin-path=<PATH>]
                              [--output=<PATH>]
+    cargo raze --version
 
 Options:
     -h, --help                         Print this message
+    --version                          Print version info and exit
     -v, --verbose                      Use verbose output
     -q, --quiet                        No output printed to stdout
     --color=<WHEN>                     Coloring: auto, always, never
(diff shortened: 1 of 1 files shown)


impl/src/bin/cargo-raze.rs
```diff
@@ -53,9 +53,11 @@ Usage:
     cargo raze [--verbose] [--quiet] [--color=<WHEN>] [--dryrun] [--cargo-bin-path=<PATH>] [--output=<PATH>]
     cargo raze <buildprefix> [--verbose] [--quiet] [--color=<WHEN>] [--dryrun] [--cargo-bin-path=<PATH>]
                              [--output=<PATH>]
+    cargo raze --version
 
 Options:
     -h, --help                         Print this message
+    --version                          Print version info and exit
```
dfreese, line 60: can you add in a -V option as well, since both rustc/cargo allow it?

impl/src/bin/cargo-raze.rs
```diff
@@ -66,6 +68,11 @@ Options:
 
 fn main() -> Result<()> {
   let options: Options = Docopt::new(USAGE)
+    .map(|d| {
+      d.version(Some(
+        concat!("cargo-raze ", env!("CARGO_PKG_VERSION")).to_string(),
+      ))
```
dfreese, line 74: does this cause an automatic exit?

dfreese (approved):

impl/src/bin/cargo-raze.rs
dfreese: (when seeing the version flag)

wchargin: Yep. As with `--help`, there’s a [`Version` error variant][verr], so the
program hits the `|e| e.exit()`.

[verr]: https://docs.rs/docopt/1.1.0/docopt/enum.Error.html#variant.Version

impl/src/bin/cargo-raze.rs
wchargin: Sure. Done.

dfreese (approved):

PR #284, opened 2020-11-07
Remove `test_metadata_deserializes_correctly` which tests unused functionality

UebelAndre:
I suspect this was relevant when raze was using `cargo` directly but this functionality should now be covered by the `cargo_metadata` crate and we shouldn't need to have this file committed.


GregBowyer (approved):

UebelAndre:
@dfreese Do you have a moment to review this PR?

acmcarther:
This was added by https://github.com/google/cargo-raze/pull/28 and yeah, we don't need to bother with this since we're using the cargo metadata crate.

Thanks for cleaning this up!

PR #290, opened 2020-11-10
Fixed regression in rendered templates

UebelAndre:
This fixes a regression where a newline had been removed from outputs resulting in lines like
```python
# Generated Targets# buildifier: disable=load-on-top
```

This corrects that so the same template renders correctly:
```python
# Generated Targets
# buildifier: disable=load-on-top
```

### DIFF
--- a/impl/src/rendering/templates/crate.BUILD.template
+++ b/impl/src/rendering/templates/crate.BUILD.template
@@ -30,7 +30,7 @@ licenses([
 
 # Generated Targets
 {%- set crate_name_sanitized = crate.pkg_name | replace(from="-", to="_") -%}
-{%- if crate.build_script_target -%}
+{%- if crate.build_script_target %}
 {%      include "templates/partials/build_script.template" %}
 {%- endif -%}
 {%- for target in crate.targets -%}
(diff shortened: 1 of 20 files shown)


impl/src/rendering/templates/crate.BUILD.template
```diff
@@ -30,7 +30,7 @@ licenses([
 
 # Generated Targets
 {%- set crate_name_sanitized = crate.pkg_name | replace(from="-", to="_") -%}
-{%- if crate.build_script_target -%}
+{%- if crate.build_script_target %}
 {%      include "templates/partials/build_script.template" %}
```
PiotrSikora, line 34: Nit: I think there should be an empty line added here to match style of other targets. There is an empty line following `# Generated Targets` in all other cases.

UebelAndre: Can you suggest a change for me to apply?

PiotrSikora, line 34:
```suggestion

{%      include "templates/partials/build_script.template" %}
```

PiotrSikora: (Just the empty line, I cannot seem to find a way to add new lines without removing anything)

PiotrSikora: Thanks!

acmcarther (approved):

PR #298, opened 2020-11-22
Added setting `rust_rules_workspace_name`

UebelAndre:
A new setting `rust_rules_workspace_name` has been added which allows users to specify the name of the `rules_rust` repository they wish to generate files with.

This is blocked by https://github.com/google/cargo-raze/pull/296
This blocks https://github.com/bazelbuild/rules_rust/pull/500

PR #300, opened 2020-11-23
Tests no longer require internet

UebelAndre:
This makes them more hermetic and makes the tests less flaky and makes porting this project to build with Bazel easier.

The changes in this PR add metadata templates to `impl/src/testing/metadata_templates` which are the formatted outputs of `cargo metadata --format-version 1`. Each `.template` file has the `Cargo.toml` contents at the top of it that produced it. These templates are then rendered during the relevant tests (test similar to the file name). This avoids having to fetch any metadata from the internet during testing.

This change also adds `RazeMetadataFetcher` which is the the new class to use for gathering metadata for the planning phase of `cargo-raze`.  Note that this no longer implements `MetadataFetcher` as `MetadataFetcher` is intended to be a wrapper around a basic way to generate a `cargo_metadata::Metadata` object. `RazeMetadataFetcher` is intended to generate a `RazeMetadata` object.

This PR is blocked by https://github.com/google/cargo-raze/pull/298


UebelAndre:
@acmcarther For this one I wasn't quite sure what to do with the metadata. I opted to store it and try to keep changes to code minimal but it would make sense to change the crates used in some of these tests.

impl/src/planning.rs
```diff
@@ -291,23 +292,13 @@ mod tests {
...
-    let toml_file = indoc! { r#"
-    [package]
-    name = "build_produces_aliased_dependencies"
-    version = "0.1.0"
-
-    [lib]
-    path = "not_a_file.rs"
-
-    [dependencies]
-    actix-web = "2.0.0"
```
UebelAndre, line 303: @acmcarther what would be a smaller replacement for these crates?

UebelAndre:
Also worth noting, this only has as many new lines of code because the metadata dumps are pretty printed but I feel that's the correct thing to do as it makes it more readable and makes it easier to track the footprint of that metadata.

impl/src/metadata.rs
```diff
@@ -191,11 +255,9 @@ impl CargoMetadataFetcher {
...
 
     // First gather metadata without downloading any dependencies so we can identify any path dependencies.
-    let no_deps_metadata = MetadataCommand::new()
-      .cargo_path(&self.cargo_bin_path)
-      .current_dir(files.toml_path.parent().unwrap()) // UNWRAP: Guarded by function assertion
-      .no_deps()
-      .exec()?;
+    let no_deps_metadata = self
+      .metadata_fetcher
+      .fetch_metadata(files.toml_path.parent().unwrap(), false)?;
```
acmcarther, line 260: Nit: `/*include_deps=*/ false`

impl/src/metadata.rs
```diff
@@ -332,33 +394,18 @@ impl CargoMetadataFetcher {
   ) -> Result<Lockfile> {
     let lockfile_path = cargo_dir.join("Cargo.lock");
 
+    // Use the reusable lockfile if one si provided
```
acmcarther, line 397: s/si/is/g

UebelAndre:
@acmcarther This is rebased and ready for review now 🙏

impl/src/planning.rs
acmcarther: Are you looking for any cargo crate to use that has few dependencies?

https://crates.io/crates/cfg-if only depends on core, so should result in a small dependency tree.

acmcarther (changes requested):
I'm not sure how I feel about the templated json, but I think it's clearly better than having the tests depend on fetching stuff from the internet.

Generally LGTM, mostly nits. High level nit as well: Can you update the PR description here to describe the abstractions you're adding and the motivation for adding them? That practice makes it easier to digest medium and larger sized PRs.

impl/src/planning.rs
UebelAndre: I don't think I quite understand this test so yeah, if there's a replacement crate you can give me, I can wire it up and it'd help me understand what this is for.

UebelAndre:
@acmcarther how's the description look now? Also, still stuck on the `actix-web` test

UebelAndre:
> I'm not sure how I feel about the templated json, but I think it's clearly better than having the tests depend on fetching stuff from the internet.

@acmcarther I'm not in love with it either but I do feel it's better than what was before. If there's a simpler solution then I could take another pass. But I do feel this change would be good to get us on the path to something better.

impl/src/planning.rs
UebelAndre: I'm not sure that crate is usable for this test

UebelAndre: looks like this was added in https://github.com/google/cargo-raze/pull/162

UebelAndre:
@acmcarther Hey, are you around today? 🙏 😅

acmcarther:
You're doing that force push thing again which is breaking the diff.

To be really clear the thing that is broken is clicking file links for prior comments, such as this: https://github.com/google/cargo-raze/pull/300/files/f7c7512837473503fcab89c08d4180968c53e43b#diff-1fb1e871b8f9024bde534a1e34e1b0285f4712c79a01284605eb55cf6e7a3e36

impl/src/planning.rs
acmcarther: Is this comment obsolete?

acmcarther:
I'm ready to merge this once the open question about that test which used actix-web is answered. Sorry for the slowness on the review.

impl/src/planning.rs
UebelAndre: I don't think it is but if the format of committed metadata json dumps is an acceptable workaround for network access, then I'd rather open an issue to update the tests (which both before and after this change, there's work to do. Even literal `TODO`s from times past). If your concern is lines of code I can unformat the files to reduce them to a single line. But to me the formatted dumps are better.

UebelAndre:
> I'm ready to merge this once the open question about that test which used actix-web is answered. Sorry for the slowness on the review.
> 
> Re: description: The new contents of the description are helpful. I was being a bit more literal with my request to include the abstractions -- i meant to literally include the non-trivial new structs. I don't feel strongly about that, hence I didn't raise that. What we have here looks good to me.

@acmcarther I can do that from here on. I think this is my last large PR though since I feel all the large missing pieces of functionality are in.

> You're doing that force push thing again which is breaking the diff links.
> 
> To be really clear the thing that is broken is clicking file links for prior comments, such as this: https://github.com/google/cargo-raze/pull/300/files/f7c7512837473503fcab89c08d4180968c53e43b#diff-1fb1e871b8f9024bde534a1e34e1b0285f4712c79a01284605eb55cf6e7a3e36

Sorry, I forgot that this PR was in review when fixing up my branches. I am making an effort to not force push once a PR is in review. Since all my PRs from here on are smaller, I think this will get easier.

impl/src/planning.rs
acmcarther: Formatted dumps are fine I think.

I don't totally follow your comment about there being more work to do, but filing a bug SGTM.

UebelAndre: https://github.com/google/cargo-raze/blob/master/impl/src/planning.rs#L806-L811 are these not still relevant? That's what I was referring to as well as other clarity cleanups like this particular test. Its not clear to me what specifically it's testing and I think that could be improved. There's also some naming inconsistencies across all tests in that some are prefixed with `test_` and others are not:
```
test planning::license::tests::empty_license_is_restricted ... ok
...
test planning::tests::test_plan_build_minimum_workspace ... ok
```

Things like this. Does that clarify what I was saying a bit more?

UebelAndre: I made https://github.com/google/cargo-raze/issues/310

UebelAndre: @acmcarther with that ticket tracking additional work to do for the unit tests, would this conversation be considered resolved?

UebelAndre:
@acmcarther I've also added more to the description.

UebelAndre:
@acmcarther Hey, are you around today to take another look here? 🙏

impl/src/planning.rs
acmcarther: Those TODOs are ancient and can be disregarded.

acmcarther (approved):

UebelAndre:
Thank you so much! 🙏

PR #301, opened 2020-11-23
Adds `--manifest-path` command line option

UebelAndre:
This adds support for a `--manifest-path` argument. To do so the `crate::planning::checks` module moved to `crate::checks` since the checks are ultimately done on the metadata and files on dick, not necessarily planning. This is prevents us from having to pipe an argument through into planning to locate the directory vendored crates are expected to be found in.

Checks now run separately of the planning step

This PR also adds a new field to `context::CrateContext` called `canonical_additional_build_file`. This field is a canonicalized (relative to the root of the Cargo workspace root) path of `raze_settings.additional_build_file` when it's set.

Finally, this PR fixes some dead code where the command line pattern `cargo-raze foo --verbose` would swallow `foo` and do nothing with it. Since this argument was not being used, it's support has been removed.

blocked by https://github.com/google/cargo-raze/pull/300


UebelAndre:
This PR continues a discussion started here: https://github.com/google/cargo-raze/pull/276#issuecomment-732514620

UebelAndre:
@acmcarther This one is ready for review now 🙏

UebelAndre:
@acmcarther just FYI, this PR grew a bit in size than it originally was after I found out `planning::BuildPlanner` was being used in https://github.com/bazelbuild/rules_rust/issues/2 (since `cargo-raze` is being used as a library there). This maintains the same interface and I did feel it was a bit unnecessary to have to change the interface in the first place. As a result I opted to move checks out of planning since they were all checks on metadata (or information that could be parsed from metadata).

impl/src/settings.rs
```diff
@@ -148,6 +149,63 @@ pub struct RazeSettings {
...
+  Validated { name: String, path: PathBuf },
+}
+
+impl From<&str> for AdditionalBuildFile {
+  fn from(value: &str) -> Self {
+    AdditionalBuildFile::Raw(value.to_owned())
+  }
+}
+
+impl<'build_file> From<&'build_file AdditionalBuildFile> for &'build_file String {
```
acmcarther, line 183: This seems weird and kind of magical. Why not just "to_string" or "to_debug_string".

(the problem with automagic conversion to string is that there's generally not a great canonical string representation of some type for all use cases -- hence a choice like "to_debug_string")

(Edit: I wrote this comment before the general comment on AdditionalBuildFile. I'm hoping you can scrap most of this machinery, and then this comment will be moot.)

impl/src/settings.rs
```diff
@@ -148,6 +149,63 @@ pub struct RazeSettings {
...
+      AdditionalBuildFile::Raw(name) => name,
+      AdditionalBuildFile::Validated {
+        name,
+        path: _,
+      } => name,
+    }
+  }
+}
+
+impl<'build_file> TryFrom<&'build_file AdditionalBuildFile> for &'build_file PathBuf {
```
acmcarther, line 195: I don't think this is a good idea either. I prefer simple functions for this unless there's a really compelling argument that this approach enhances clarity (which, for sample size one, it does not).

(Edit: I wrote this comment before the general comment on AdditionalBuildFile. I'm hoping you can scrap most of this, and then this comment will be moot.)

impl/src/settings.rs
```diff
@@ -148,6 +149,63 @@ pub struct RazeSettings {
...
+ * file from a `Cargo.toml` file) but resolvable to a canonicalized path so that the additional
+ * BUILD file may be accessed at any point. Build files that have relitive paths associated with
+ * them are considered `Validated`.
+ */
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum AdditionalBuildFile {
+  Raw(String),
+  Validated { name: String, path: PathBuf },
+}
```
acmcarther, line 175: This seems really clunky. IIUC, you're using a validation function to convert an initial "raw" string value into this "validated" value. You're masking over that with a bunch of From<> implementations.

Can we just parse this as a PathBuf and validate that it exists in validate_settings, instead of jumping through all of these hoops?

acmcarther (changes requested):
Thanks for putting this together.

I think I'm _quite strongly_ against the magical way that AdditionalBuildFile is implemented both in that it enumerates over the "raw" and "processed" in a way that I think conflates parsing and validation, but also in the way it (to my eyes) relies on abuse of `From` and `Into`. You might be able to make a convincing argument for it though.

impl/src/checks.rs
```diff
@@ -146,9 +214,47 @@ pub fn warn_unused_settings(
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
```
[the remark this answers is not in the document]
UebelAndre: Ah, I wasn't thinking about the testing module at the time. I can move it there. Good catch.

impl/src/settings.rs
UebelAndre: The difficulty here is that if we have multiple packages that specify these files, we no longer know which package they came from once the settings are loaded since the `CrateSettings` are flattened into a single map. Additionally, if we simply use a `PathBuf` we end up writing canonicalized paths to files which is not ideal since it's not reproducible from one user to another when working on the same project. I then wondered if I can use the cargo workspace root as the anchor, but felt that the traceability/debugability was lacking in that users wouldn't be able to `grep` for any path rendered in a BUILD file through their `Cargo.toml` files to find problematic crate settings. Maybe that last option isn't actually so bad? My goal with this implementation was to try to keep this enum as invisible as possible.

With all that said, any new thoughts?

impl/src/settings.rs
UebelAndre: Gonna hold on addressing this until https://github.com/google/cargo-raze/pull/301#discussion_r538019610 is resolved

impl/src/settings.rs
UebelAndre: Gonna hold on addressing this until https://github.com/google/cargo-raze/pull/301#discussion_r538019610 is resolved

impl/src/checks.rs
```diff
@@ -146,9 +214,47 @@ pub fn warn_unused_settings(
...
   use super::*;
-  use crate::metadata::tests::dummy_raze_metadata;
+  use crate::{metadata::tests::dummy_raze_metadata, settings::tests::dummy_raze_settings};
+  use semver::Version;
+
+  pub fn dummy_modified_metadata() -> Metadata {
+    let mut metadata = dummy_raze_metadata().metadata.clone();
+
+    // Phase 1: Add a dummy dependency to the dependency graph.
```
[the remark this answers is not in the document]
UebelAndre: I didn't add that, it's a copy/paste from times past and just hadn't noticed it. I do try to do that for the code I write but I'd prefer to leave this for now since I plan on combing through the tests in https://github.com/google/cargo-raze/issues/310

impl/src/settings.rs
```diff
@@ -367,8 +429,90 @@ pub fn format_registry_url(registry_url: &str, name: &str, version: &str) -> Str
...
+      }
+
+      let additional_build_file: &mut &mut AdditionalBuildFile =
+        &mut crate_settings.additional_build_file.as_mut().unwrap();
+      validate_create_setting_additional_build_file(
+        additional_build_file,
+        toml_path,
+        crate_name,
+        version,
+      )?;
```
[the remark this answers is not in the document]
UebelAndre: There isn't one currently (that I know of, unless `anyhow` has some functionality for that).  I feel generally that the `error` module needs to be revisited. Each case is kinda clunky (given the common use of `anyhow` and I think we should either move more into `anyhow` or refine the Errors. I'd rather not introduce a new crate just for this since I feel like with some refactoring it won't be necessary. If you feel strongly about it I can try to crack it out quick but I have a queue of changes I'd rather get in first.

UebelAndre:
> Thanks for putting this together.
> 
> I think I'm _quite strongly_ against the magical way that AdditionalBuildFile is implemented both in that it enumerates over the "raw" and "processed" in a way that I think conflates parsing and validation, but also in the way it (to my eyes) relies on abuse of `From` and `Into`. You might be able to make a convincing argument for it though.

@acmcarther I responded on one of the comments you left about that enum. Can you elaborate on how the use of `From` and `Into` here is abuse? This is my first time ever really implementing anything of my own with them so I'm not surprised it'd be less than ideal 😄

UebelAndre:
@acmcarther Oh and just FYI, this and https://github.com/google/cargo-raze/pull/302 are my two medium PRs. I have many other small ones that I think you'll enjoy, not just because I think they add nice features, but because they're small and focused 😅

UebelAndre:
@acmcarther Following up on https://github.com/google/cargo-raze/pull/301#issuecomment-740376641 if I could have 2 hours of your time each day for the rest of this week(~10am and some time in the afternoon), I think we could get all the pending PRs through and have a release by the end of the week. This would give me the time to address your feedback and hopefully have something merged in the afternoon or the next day. This would mean a lot to me and I also feel it'd be extremely impactful for `rules_rust` as well.

impl/src/checks.rs
```diff
@@ -14,32 +14,69 @@
...
+  util::get_package_ident,
+};
 
 use cargo_metadata::{Metadata, Package, PackageId};
 
 // TODO(acmcarther): Consider including a switch to disable limiting
 const MAX_DISPLAYED_MISSING_VENDORED_CRATES: usize = 5;
 const MAX_DISPLAYED_MISSING_RESOLVE_PACKAGES: usize = 5;
 
+pub fn check_metadata(
```
acmcarther, line 38: Document this function

impl/src/checks.rs
```diff
@@ -146,9 +214,12 @@ pub fn warn_unused_settings(
 }
 
 #[cfg(test)]
-mod tests {
+pub mod tests {
```
acmcarther, line 217: Why is this pub?

impl/src/checks.rs
```diff
@@ -51,23 +88,54 @@ pub fn check_all_vendored(
...
+  bazel_workspace_root
+    // Trim the absolute label identifier from the start of the workspace path
+    .join(workspace_path.trim_start_matches('/'))
+    .join(vendor_dir)
+}
+
+/**
+ * Returns the packages expected path during current execution.
+ *
+ * Not for use except during planning as path is local to run location.
```
acmcarther, line 121: I can't tell if this is a new comment or not, but you're literally using this function outside of planning because this is called by `check_all_vendored`, which is called by `check_metadata`, which is called in cargo-raze.rs. Delete this comment?

impl/src/checks.rs
```diff
@@ -166,4 +237,19 @@ mod tests {
     // Should not panic with valid metadata.
     check_resolve_matches_packages(&metadata).unwrap();
   }
+
+  #[test]
+  fn test_verifies_vendored_state() {
```
acmcarther, line 242: This test doesn't follow the naming scheme of tests in this module, which is

test_$FUNCTION_UNDER_TEST_$HAS_BEHAVIOR.

impl/src/metadata.rs
```diff
@@ -98,8 +100,9 @@ pub struct RazeMetadata {
   // `cargo metadata` output of the current project
   pub metadata: Metadata,
 
-  // The absolute path to the current project's cargo workspace root
-  pub workspace_root: PathBuf,
+  // The absolute path to the current project's cargo workspace root. Note that the workspace
+  // root in `metadata` will be inside of a temporary directory.
+  pub cargo_workspace_root: PathBuf,
```
acmcarther, line 105: Can you add a reference to 

https://doc.rust-lang.org/cargo/reference/workspaces.html#root-package

Here? My brain was failing to parse "cargo_workspace_root" because my brain was stuck to thinking of workspace root as "bazel root".

impl/src/util.rs
```diff
@@ -330,6 +331,10 @@ pub fn get_workspace_member_path(manifest_path: &Path, workspace_root: &Path) ->
   diff_paths(manifest_path.parent().unwrap(), workspace_root)
 }
 
+pub fn get_package_ident(package_name: &str, package_version: &str) -> String {
```
acmcarther, line 334: Nit: Is just `package_ident` sufficient? `get` implies to me that we're looking it up from somewhere or something.

Also, probably good to define "package_ident" here as part of function documentation.

impl/src/testing.rs
```diff
@@ -287,3 +293,41 @@ pub fn mock_crate_index(
     None
   }
 }
+
+/** Generate some basic metadata with an injected mock dependency */
+pub fn dummy_modified_metadata() -> Metadata {
+  let mut metadata = dummy_raze_metadata().metadata.clone();
+
+  // Phase 1: Add a dummy dependency to the dependency graph.
```
acmcarther, line 301: Just delete these weird comments

impl/src/settings.rs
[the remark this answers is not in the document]
acmcarther: `check_all_vendored` does something manual that you could consider here. It aggregates the errors as crate names, then converts them to a single error.

I don't really feel strongly about this though.

impl/src/settings.rs
acmcarther: > The difficulty here is that if we have multiple packages that specify these files, we no longer know which package they came from once the settings are loaded since the CrateSettings are flattened into a single map. 

What does this mean exactly? Where are the settings "flattened into a single map"?

I looked at create_crate_context, which is responsible from plumbing the settings from the settings object into the crate context object (which is later used below for cat-ing the contents of the "additional build file" into the generated file. This relies on `crate_settings`, which seems perfectly capable of distinguishing between different versions of the same crate.

> Additionally, if we simply use a PathBuf we end up writing canonicalized paths to files which is not ideal since it's not reproducible from one user to another when working on the same project. 

I don't know what this means either. What I drew from this was that there was some possibility of users specifying a fully qualified path here. I don't really know much about this feature because I've not been doing a good job reviewing all of this new stuff. Here's an example of an existing file:

https://github.com/google/cargo-raze/blob/06eaef173bf81f9b43f088a4243a8b7bd534c662/examples/remote/complicated_cargo_library/Cargo.toml#L32

That's not fully qualified.

What's wrong with parsing this as a pathbuf exactly?

> I then wondered if I can use the cargo workspace root as the anchor, but felt that the traceability/debugability was lacking in that users wouldn't be able to grep for any path rendered in a BUILD file through their Cargo.toml files to find problematic crate settings. Maybe that last option isn't actually so bad? 

I'm guessing that "using the cargo workspace as the anchor" means use the status quo situation where users specify a non-fully qualified path. That sounds right to me. 

> My goal with this implementation was to try to keep this enum as invisible as possible.

This enum is anything but invisible [for people who have to read and understand this code]

acmcarther (changes requested):
I've spent the last several months reviewing these PRs without having looked at the code and regenerated my mental map of the code. I've done that a bit, but still, this code has really run away from me. Lots of small things that need to get cleaned up.

acmcarther:
I didn't respond anywhere on `From` or `Into`.

I'd advise against using them in general. I think people look at them as a kind of tool for polymorphism. I think they can be useful for that, but only in particularly narrow context. Using them more widely invites a kind of "conversion chaos" that allows you to patch over design issues (like what I think this PR is introducing) by making the code still "feel" convenient.

I recognize that that's not a very precise criticism. Here's an example of an Into/From usage that I think is great

https://doc.rust-lang.org/std/net/enum.SocketAddr.html#impl-FromStr

This is a convenience conversion for the type "SocketAddr" which allows you to define a function:
```rust
fn start_server<T: Into<SocketAddr>>(addr: T) {
  // Start the server.
}
```
I like this. I don't think it introduces the possibility for confusion, and I don't think it masks over API clunkiness in the same way that this PR's usage of the traits does.

I'm not sure how to convey why I think that example usage is different than this usage. I guess maybe it has a complexity:convenience ratio which isn't as good?

acmcarther:
I think above you said something really important that I should better understand

> Additionally, if we simply use a PathBuf we end up writing canonicalized paths to files 

Can you elaborate on this?

UebelAndre:
In response to https://github.com/google/cargo-raze/pull/301#discussion_r538964319

> What does this mean exactly? Where are the settings "flattened into a single map"?

Oh, that actually relates to https://github.com/google/cargo-raze/pull/302 but looking at [settings.rs#L97](https://github.com/UebelAndre/cargo-raze/blob/manifest-path/impl/src/settings.rs#L97), that map stores all crate settings and in the context of cargo workspaces, the assumption that a single crate setting's `additional_build_file` setting being relative to the cargo workspace root is going to be incorrect in almost every single case. The crate settings in that PR get collected from each crate and stored in that map. If there was no workspace support, then yeah, I think it'd be easy to parse this as a `PathBuf` but I don't think that works here.

> I don't know what this means either. What I drew from this was that there was some possibility of users specifying a fully qualified path here. I don't really know much about this feature because I've not been doing a good job reviewing all of this new stuff. Here's an example of an existing file:

I'd initially tried parsing `additional_build_file` as a `PathBuf` which kinda but again, if these paths were to be canonicalized (which they need to be especially after https://github.com/google/cargo-raze/pull/302) then the full path (including potential user information) would be written by [bazel.rs#L228-L231](https://github.com/UebelAndre/cargo-raze/blob/manifest-path/impl/src/rendering/bazel.rs#L228-L231). So the paths written there need to at least be relative to the cargo workspace root.

> I'm guessing that "using the cargo workspace as the anchor" means use the status quo situation where users specify a non-fully qualified path. That sounds right to me.

I'll see if there's a solution here. There likely is.

> This enum is anything but invisible [for people who have to read and understand this code]

I need to read your next message, but I meant to say "invisible for users, not necessarly contributors. I felt these ~40 lines of code were fairly straight forward so I don't think they're quite magic. But again, I'll read https://github.com/google/cargo-raze/pull/301#issuecomment-741495411 and reploy.

UebelAndre:
> I think above you said something really important that I should better understand
> 
> > Additionally, if we simply use a PathBuf we end up writing canonicalized paths to files
> 
> Can you elaborate on this?

I think I addressed this in https://github.com/google/cargo-raze/pull/301#issuecomment-741500690 
let me know.

acmcarther:
> Oh, that actually relates to #302

:disappointed: 

> but looking at settings.rs#L97, that map stores all crate settings and in the context of cargo workspaces, **the assumption that a single crate setting's additional_build_file setting being relative to the cargo workspace root is  going to be incorrect in almost every single case.** 

Are you saying that

> most crate settings additional_build_file will not be [placed] relative to the cargo workspace root.

Why not? Where does this assumption come from? You're going to have to explain this like I'm five, because I'm missing huge amounts of context, and I _really_ don't like this current approach, so I'm going to insist on understanding instead of just waving the change through like I usually do.

acmcarther:
Can we move the "canonicalization" out of Settings-land and into CrateContext-land? I don't think I'm against the feature, and generally, I'm not hostile to the convenience offered by allowing users to specify workspace sub-crate specific settings within that sub-crate's toml (in fact, I'm convinced this is the right decision), but I'm against the weirdness being introduced in parsing of crate settings.

This has the general shape of similar discussions we've had in the past which is basically that I see certain boundaries between the abstractions that are being crossed. In this case, I view the settings abstraction as mapping one to one to the contents of the TOML. Extension of what's known there (e.g. canonicalization, determining relationships between crates) i view as a responsibility of the "crate catalog".

UebelAndre:
> Can we move the "canonicalization" out of Settings-land and into CrateContext-land? I don't think I'm against the feature, and generally, I'm not hostile to the convenience offered by allowing users to specify workspace sub-crate specific settings within that sub-crate's toml (in fact, I'm convinced this is the right decision), but I'm against the weirdness being introduced in parsing of crate settings.
> 
> This has the general shape of similar discussions we've had in the past which is basically that I see certain boundaries between the abstractions that are being crossed. In this case, I view the settings abstraction as mapping one to one to the contents of the TOML. Extension of what's known there (e.g. canonicalization, determining relationships between crates) i view as a responsibility of the "crate catalog".

I'm super excited I have so much of your time tonight but I'm unfortunately tied up at work as well. I'll be actioning on your feedback tomorrow morning. So yeah, I think i can simplify this and make more suited to your liking but would like to continue the conversation about `From` and `Into` as well after that.

impl/src/settings.rs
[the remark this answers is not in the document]
UebelAndre: I've updated this but I'm not happy about the state of the state of it. I totally understand the value in collecting all errors vs surfacing one at a time but I think the correct solution here is to update `RazeError` to handle chained errors. Having to collect errors then write them to a string seems to invalidate the existence of `RazeError` in that e might as well be using `anyhow` everywhere if the only interesting content is the message. I'd much prefer for the errors themselves to carry more meaning and have relevant and convenient data associated that might even be able to convey the message on it's own.

Anyway, updated

impl/src/checks.rs
UebelAndre: it shouldn't be

UebelAndre:
> Can we move the "canonicalization" out of Settings-land and into CrateContext-land? I don't think I'm against the feature, and generally, I'm not hostile to the convenience offered by allowing users to specify workspace sub-crate specific settings within that sub-crate's toml (in fact, I'm convinced this is the right decision), but I'm against the weirdness being introduced in parsing of crate settings.
> 
> This has the general shape of similar discussions we've had in the past which is basically that I see certain boundaries between the abstractions that are being crossed. In this case, I view the settings abstraction as mapping one to one to the contents of the TOML. Extension of what's known there (e.g. canonicalization, determining relationships between crates) i view as a responsibility of the "crate catalog".

The PR has been updated now and since we're parsing `additional_build_file` as a `PathBuf` I feel "canonicalization" makes sense in the settings. If the metadata contains `../some/path/to/BUILD.file.bazel` then I would say it's better to canonicalize that path unless you feel it has significance as a string/relative path. `enum AdditionalBuildFile` was a bit much but canonicalizing a pathbuf seems fine. If there was a way to canonicalize paths via `serde` I'd opt to do that and simplify all this.

UebelAndre:
@acmcarther This PR is ready for another review

impl/src/settings.rs
UebelAndre: > What does this mean exactly? Where are the settings "flattened into a single map"?

Oh, that actually relates to https://github.com/google/cargo-raze/pull/302 but looking at [settings.rs#L97](https://github.com/UebelAndre/cargo-raze/blob/manifest-path/impl/src/settings.rs#L97), that map stores all crate settings and in the context of cargo workspaces, the assumption that a single crate setting's `additional_build_file` setting being relative to the cargo workspace root is going to be incorrect in almost every single case. The crate settings in that PR get collected from each crate and stored in that map. If there was no workspace support, then yeah, I think it'd be easy to parse this as a `PathBuf` but I don't think that works here.

> I don't know what this means either. What I drew from this was that there was some possibility of users specifying a fully qualified path here. I don't really know much about this feature because I've not been doing a good job reviewing all of this new stuff. Here's an example of an existing file:

I'd initially tried parsing `additional_build_file` as a `PathBuf` which kinda but again, if these paths were to be canonicalized (which they need to be especially after https://github.com/google/cargo-raze/pull/302) then the full path (including potential user information) would be written by [bazel.rs#L228-L231](https://github.com/UebelAndre/cargo-raze/blob/manifest-path/impl/src/rendering/bazel.rs#L228-L231). So the paths written there need to at least be relative to the cargo workspace root.

> I'm guessing that "using the cargo workspace as the anchor" means use the status quo situation where users specify a non-fully qualified path. That sounds right to me.

I'll see if there's a solution here. There likely is.

> This enum is anything but invisible [for people who have to read and understand this code]

I need to read your next message, but I meant to say "invisible for users, not necessarly contributors. I felt these ~40 lines of code were fairly straight forward so I don't think they're quite magic. But again, I'll read https://github.com/google/cargo-raze/pull/301#issuecomment-741495411 and reply.

impl/src/settings.rs
```diff
@@ -367,8 +380,95 @@ pub fn format_registry_url(registry_url: &str, name: &str, version: &str) -> Str
...
+  version: &VersionReq,
+) -> Result<(), RazeError> {
+  if file_path.is_absolute() {
+    return Err(RazeError::Config {
+      field_path_opt: Some(format!(
+        "raze.crates.{}.{}.additional_build_file",
+        crate_name,
+        version.to_string()
+      )),
+      message: "additional_build_file flags should be reletive paths from the manifest containing \
```
acmcarther, line 395: s/reletive/relative

impl/src/settings.rs
```diff
@@ -367,8 +380,95 @@ pub fn format_registry_url(registry_url: &str, name: &str, version: &str) -> Str
...
+        .to_owned(),
+    });
+  }
+
+  Ok(())
+}
+
+/** Ensures crate settings which specify additional build files have canonicalized paths */
+fn validate_crate_setting_additional_build_file(
+  additional_build_file: &mut &mut PathBuf,
```
acmcarther, line 406: Hey, what's going on here?

Are you using the validation function to canonicalize this path? This both syntactically is weird, and also represents a really weird and unintuitive thing for the settings code to do which I think will be confusing to future readers. Is it really not possible to do this work during catalog construction? I see you only need the toml. Are we losing the toml or something?

impl/src/settings.rs
```diff
@@ -367,8 +380,95 @@ pub fn format_registry_url(registry_url: &str, name: &str, version: &str) -> Str
...
+    version,
+  )?;
+
+  **additional_build_file = toml_path.join(&additional_build_file).canonicalize()?;
+
+  Ok(())
+}
+
+/** Ensures crate settings associatd with the parsed [RazeSettings](crate::settings::RazeSettings) have valid crate settings */
+fn validate_crate_settings(settings: &mut RazeSettings, toml_path: &Path) -> Result<(), RazeError> {
```
acmcarther, line 423: I saw this *after* seeing the above function, so I know what's going on. Just wanted to point out that readers inspecting this are going to be real suspicious that this function is doing something surreptitious seeing that it accepts the settings *mutably*. That's kind of the beauty of rust -- that it forces you to be up front when you're doing something weird like passing a mutable value to a "validation" function.

acmcarther (changes requested):
One of the things I identified while reviewing the repo's code holistically yesterday was that bin/cargo-raze.rs does too much. I'll probably start by factoring the implementation into functions which I can eventually move in the "library" code.

(this is an organic thing that dates way way back, I was just reminded of the point upon reviewing the code again)

---

I very much appreciate that you've been able to narrow this field down to a Pathbuf! Great work! I still am kind of suspicious of mutating the settings the way you're doing here though and still would champion doing this work in crate context construction.

acmcarther:
(I expect my last comments to be the last round of review unless we get into a lengthy discussion about mutating settings during validation. Anyway, everything else LGTM)

impl/src/settings.rs
UebelAndre: What should I do about this? I updated the name to hopefully make this clearer but I'm open to alternative solutions.

impl/src/settings.rs
UebelAndre: So is this being handled here: https://github.com/google/cargo-raze/pull/301#discussion_r539820850 ?

UebelAndre:
@acmcarther Some questions but I also made a few small updates. Otherwise ready for your eyes again.

impl/src/settings.rs
acmcarther: No. The opposite. Fixing this problem fixes that problem.

impl/src/settings.rs
acmcarther: Fix the other issue, and this should no longer need to be `&mut RazeSettings`

impl/src/settings.rs
UebelAndre: This is where that field gets canonicalized. How can I go about doing that?

UebelAndre: Also, this method currently mutates `RazeSettings` https://github.com/google/cargo-raze/blob/master/impl/src/settings.rs#L387

UebelAndre: Can you suggest a better pattern here?

UebelAndre:
@acmcarther More questions for you

Also

> One of the things I identified while reviewing the repo's code holistically yesterday was that bin/cargo-raze.rs does too much. I'll probably start by factoring the implementation into functions which I can eventually move in the "library" code.

I've got a PR where I've started that. I totally agree though that `bin/cargo-raze.rs` 😄

impl/src/settings.rs
UebelAndre: Actually that whole function does some sanitation next to validation. Would I be able to address the mixup there in another PR?

UebelAndre: Still happy to make some improvements here. Just don't want to caught up doing code cleanup (in this PR).

acmcarther: I wrote this comment two times already and github ate my comment both times (once because I opened the code to look at the context, and once because I accidentally ctrl-r). I am grump. This version is going to be more terse than the last two times.

So, yes, that code's existence means that you can't remove the &mut below. It's also funny, because I wrote that code (ouch).

There's a difference here. My code standardizes the path so that future code can operate on it without dealing with the presence or absence of the slash. A developer who specifies a path while debugging something is going to see approximately the same path in their toml, and in their code and will not be confused. Yours, on the other hand, is making a significant and meaningful change to the setting value of the same name. What you see in the toml, and what you see in the planning code for the value of this field is not the same. This is an nuanced argument, but I think it holds up.

In my code, I could've been consistent in this principle by either (1) telling users not to end their paths in a slash (user hostile!), or (2) updating the code which accesses this field to support *either* a trailing slash *or* no slash.

---

I'd like to provide a better suggestion, but I don't understand how this PR, and #302 are parsing the individual toml files. My initial suggestion was to retain the cargo toml path, and canonicalize using that path during render_planned_build (where you're accessing the additional_build_path). I recognized that this would only work for the root toml though, so I referenced #302 to understand how you're properly canonicalizing settings from workspace subcrates, and I don't think you are. validate_settings's toml path comes from load_settings. In #302, load_settings is only being called once, with one toml path, which is the root toml path.

Am I missing something here?

acmcarther: My working understanding is that canonicalization requires access to the workspace subcrate's toml. My lack of clarity above is basically that your next PR is not correctly handling this anyway. If your next PR *is* correct, and in fact, you *can* do canonicalization with only the workspace root toml path, then I don't see what the issue at all is. Canonicalize in render_planned_build using the path to the workspace root toml.

impl/src/settings.rs
```diff
@@ -367,8 +380,95 @@ pub fn format_registry_url(registry_url: &str, name: &str, version: &str) -> Str
...
+      message: format!("{:?}", errors),
+    });
+  }
+
+  Ok(())
+}
+
 /** Verifies that the provided settings make sense. */
-fn validate_settings(settings: &mut RazeSettings) -> Result<(), RazeError> {
+fn validate_settings(settings: &mut RazeSettings, toml_path: &Path) -> Result<(), RazeError> {
```
acmcarther, line 471: "toml_path" here means "workspace_root_toml". Actually, not even that, it means "workspace_root_toml_directory".

UebelAndre: It only means that in this context but it's incorrect to assume it will always be the root toml path

impl/src/settings.rs
acmcarther: I might be wrong on this. My initial analysis was based on some ctrl+f-ing, and it seems i didn't expand settings.rs

acmcarther: No, I think I'm right. 

You're parsing out the settings from the workspace crates in parse_raze_settings_workspace. You're validating these settings all together with the concatenated settings in validate_settings (after calling `parse_raze_settings`) only once, only with the root toml.

UebelAndre: The future I'm trying to get to is that all raze settings are parsed via cargo metadata. There are two places for metadata, packages and the workspace itself. When we're parsing package metadata, (specifically the `additional_build_file` argument), it will be a relative path to a file in the repo (or it historically has been). If we parse a package's metadata, we'll have that file path as it's written. This is not sufficient in future phases because it's a relative path to a file *from* the `Cargo.toml` file it's been defined in. Because all crate settings are stored in a HashMap `pub crates: HashMap<String, CrateSettingsPerVersion>,` (and because we don't track the path to the package a setting came from), we are unable to find the actually file on disk.

impl/src/settings.rs
acmcarther: I dug in a bit more here, and I think I'm right here. Conceptually, yes, this could be an arbitrary toml path. But in practice, this is two things. First, it's a directory which contains a toml (a minor quible). Second, this is only called one time, in the context of the root (workspace) toml, after the concatenative shenanigans of parse_raze_settings and parse workspace settings and all of that.

acmcarther:
I probably won't have another chance to revisit this tonight, but I can commit to revisiting this tomorrow morning.

impl/src/settings.rs
UebelAndre: Oh, I think I see what you're saying. This might be a bug

acmcarther:
Through the digging tonight I've gained a better appreciation for the problem here. I think the general shape of the solution I'm going to suggestion will be to extend crate settings with a reference to the toml which contained those settings, and use that to do canonicalization _later_. This will probably be a field that should not be deserialized from the toml, (so, should use a special serde annotation to prevent that if such a thing exists), and instead will be specified during settings parsing. Then during rendering, when we're looking up the crate settings, we'll have access to this.

This raises a new kind of question though, which is basically "what should the behavior be when multiple subcrates have conflicting values". I'm sure you've answered this somehow in #302, so I'll need to better understand the current appraoch

impl/src/settings.rs
UebelAndre: Yeah, this was definitely not doing what I had intended. I've updated it. The path is parsed as-is for `RazeSettings` and the path is updated during the `planning` phase where it's canonicalized. Paths must be relative paths from the *Cargo* workspace root (which should match the current behavior). It'll be a bit awkward when defining that field in multiple packages but I think it'll be fine and perhaps this can be improved in the future (or maybe it wont' really be a problem, time will tell).

impl/src/settings.rs
UebelAndre: Continuing this here: https://github.com/google/cargo-raze/pull/301#discussion_r539913565

But this is also resolved.

impl/src/settings.rs
UebelAndre: I updated this. I think it's better now

UebelAndre:
@acmcarther I think I've addressed the last big concern here about `additional_build_file`. Thanks for calling that out. Looking forward to your review 😄

impl/src/planning/subplanners.rs
```diff
@@ -335,6 +335,24 @@ impl<'planner> CrateSubplanner<'planner> {
...
+        .join(&additional_build_file)
+        .canonicalize()
+        .with_context(|| {
+          format!(
+            "Failed to find additional_build_file: {}",
+            &additional_build_file.display()
+          )
+        })?;
+
+      raze_settings.additional_build_file = Some(canonicalized_path);
```
acmcarther, line 353: What do you think about adding a field to CrateContext "canonical_additional_build_file" instead of mutating the settings?

This seems like a minor quibble, but for some reason I'm still peeved about this. I think I view the settings as sacrosanct, and the mutation of the value here in the middle of the logic seems like heresy.

(at a meta level, I'm still trying to understand why this bothers me....)

acmcarther (commented):
You've been more than sufficiently accomodating for my (i guess) inexplicable aversion to mutating settings beyond the particular way I did years ago. Sorry for the inconvenience here. I think I have a final alternative to that which can work for the way we're currently dealing with additional_build_file (noted in the other comment)

impl/src/planning/subplanners.rs
acmcarther: (then we can pull the canonical_additional_build_file right out and use it here: https://github.com/google/cargo-raze/blob/4d1721ed32e19dfea8794f868a4884bdffdc4014/impl/src/rendering/bazel.rs#L217)

UebelAndre: Sorry, just saw this suggestion but I think I updated this accordingly?

UebelAndre:
@acmcarther Ready for another review 🙏

acmcarther (approved):
LGTM

PR #277, opened 2020-11-01
Workspace support examples

UebelAndre:
This is a followup of https://github.com/google/cargo-raze/pull/276 which adds examples demonstrating the use of cargo workspaces. Only the final two commits in this PR are relevant. Please see https://github.com/google/cargo-raze/pull/276 for details about anything before that.

Issue #48, opened 2018-06-04
Make #18 more noticeable


mfarrugi:
It's difficult to notice given the flood of output, and can happen often when versions are changing. 

Might be better to hard fail and alert the human, or hide some of the other output.

acmcarther:
Perhaps we could suppress most of the other output or quietly write it into somewhere temp-like?

Thoughts?

UebelAndre:
@mfarrugi Do you feel things have improved since posting this?

mfarrugi:
I have not been a user for a long time. Feel free to close this issue.

dfreese:
I think this still could be improved.  Having semver support probably helps, but it might be worth having an option to have unused overrides being a hard error, or similar.

nikclayton-dfinity:
+1 to hard errors (speaking as someone who spent some time tracking down why overrides weren't working only to realize I'd written `[raze.crate...]` and not `[raze.crates...]` for one entry...)

PR #302, opened 2020-11-23
Parse settings from Cargo metadata

UebelAndre:
This change enables settings to be parsed from Cargo metadata to take full advantage of the changes introduced in https://github.com/google/cargo-raze/pull/274. This also improves upon the changes in https://github.com/google/cargo-raze/pull/276 to allow users to specify `CrateSettings` to be specified in the relevant packages when dealing with [workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) projects

This also contains updates to the top level README. A rendering can be found [here](https://github.com/UebelAndre/cargo-raze/blob/settings/README.md)

Blocked by https://github.com/google/cargo-raze/pull/306


acmcarther:
Why does this CL have a 57k LOC added diff?

UebelAndre:
@acmcarther This PR contains changes from others I'm hoping to get through before it. So it'll appear large until then. I'll be marking the PRs ready for review in the order I'd planned on rolling them out but you can also follow the chain in the tickets noted in the description.

UebelAndre:
> Why does this CL have a 57k LOC added diff?

But to directly answer that. It's due to the changes in https://github.com/google/cargo-raze/pull/300 where cargo metadata dumps were added for testing. Unfortunately, some tests use very large packages resulting in large json files.

UebelAndre:
> Why does this CL have a 57k LOC added diff?

Much smaller now that #300 is not in this CL. This will also continue to shrink.

UebelAndre:
Documentation has been updated in this PR as per a discussion here: https://github.com/google/cargo-raze/pull/277#issuecomment-732357034

README.md
```diff
@@ -90,40 +78,93 @@ path = "fake_lib.rs"
...
+#### Cargo workspace projects
 
-```bash
-$ cargo raze
+In projects that use [cargo workspaces](cargo_workspaces) uses should organize
+all of their `raze` settings into the `[workspace.metadata.raze]` field in the
+top level `Cargo.toml` file which contains the `[workspace]` definition. These
+settings should be identical to the ones seen in `[package.metadata.raze]` in
+[the previous section](#Using existing Cargo.toml). However, crate settings may still
+be placed in the `Cargo.toml` files of the workspace memebers:
```
acmcarther, line 144: Is this the reason for this "canonicalization" discussion from https://github.com/google/cargo-raze/pull/301?

Can we require all raze metadata live in the workspace root toml?

examples/remote/binary_dependencies/Cargo.toml
```diff
@@ -15,6 +15,7 @@ workspace_path = "//remote/binary_dependencies/cargo"
 gen_workspace_prefix = "remote_binary_dependencies"
 genmode = "Remote"
 default_gen_buildrs = true
+package_aliases_dir = "cargo"
```
acmcarther, line 18: ugh

acmcarther: translation:

what is this field

why is it here

why isn't this some default setting

why is this not mentioned in the PR description.

README.md
UebelAndre: I feel quite strongly that settings relevant to to dependencies for a specific package should be able to live in that package's `Cargo.toml`

examples/remote/binary_dependencies/Cargo.toml
UebelAndre: Ultimately doesn't really impact the tests but in the grand scheme of things I do feel that it's more correct to have a named directory and not `"."` for this setting. But that can be discussed later (since this PR is still in draft phase and another PR more directly relates to this)

UebelAndre:
@acmcarther This is a draft PR. I open draft PRs mainly for reference and high level conversation. The state of most of my changes right now are that they're based on other changes that I plan to open PRs for (if they're not open already). This change is 2 PRs ahead of master.

README.md
acmcarther: IIUC package in this context doesn't mean "free floating crate in the sea of crates", but rather "crate within a cargo workspace". What's the value add here that is worth the additional complexity imposed on the design?

examples/remote/binary_dependencies/Cargo.toml
acmcarther: I don't understand how you envision this code review process working...

acmcarther:
I came here because you referenced this PR in the other PR as an explanation of a design decision.

README.md
UebelAndre: When I author the `Cargo.toml` file for my package (which your definition is correct in this context), and I add `serde-json` as a dependency, I shouldn't have to go over to my root `Cargo.toml` file to update it to be compatible with cargo raze. When working on a project, users should only need to deal with 1 manifest.

UebelAndre:
> I came here because you referenced this PR in the other PR as an explanation of a design decision.

Yeah, my message was in response to https://github.com/google/cargo-raze/pull/302#discussion_r538991018

Which I'm happy to talk about but not quite ready to do so since the reasoning for that is in another PR. But I thought it'd be valuable for you to see the state of things here at a high level.

examples/remote/binary_dependencies/Cargo.toml
UebelAndre: I think the real issue here is that these drafts are stacked on other changes. So while I'm thinking, "oh, that's relevant to another PR", that won't be clear to anyone else.

README.md
acmcarther: I think you've convinced me of this, so for now, I'm approaching this in a different way in #301 where I"m basically hoping you can move the "nasty shenanigans"_(tm)_ out of settings processing and into CrateContext

UebelAndre:
Just a heads up, I'm going to be rebasing this branch. I will stop rebasing once this is marked "Ready for review".

README.md
UebelAndre: I think I was able to accomplish that. Right now these paths are relative to the cargo workspace root, not the current toml file (which might be bad? Certainly not ideal). I'm fine with it for now since the pattern seems to be that additional build files are generally put in `workspace_path` anyway. I think once the dust is settled there can be a small PR to improve that usability.

impl/src/settings.rs
```diff
@@ -395,78 +463,304 @@ fn validate_settings(settings: &mut RazeSettings) -> Result<(), RazeError> {
...
+    .collect();
+
+  let mut duplicate_binary_deps = Vec::new();
+  let mut duplicate_crate_settings = Vec::new();
+
+  for package in workspace_packages.iter() {
+    if let Some(pkg_value) = package.metadata.get("raze") {
+      let pkg_settings = RawRazeSettings::deserialize(pkg_value)?;
+      if pkg_settings.contains_primary_options() {
+        return Err(anyhow!(
```
UebelAndre, line 571: (hopefully this comment doesn't move) 
Here is where error messages for duplicate crate settings (2 package metadata sets contain a definition for the same crate setting) are generated. My though is if you use the same dependency in more than one package, it's a workspace dependency and should be updated to reflect that in raze settings.

UebelAndre:
@acmcarther  This PR is now ready for review.

examples/remote/binary_dependencies/Cargo.toml
UebelAndre: Does this make more sense after reviewing #306 ?

README.md
acmcarther: Fully agreed with your assessment here. Marking resolved.

impl/src/metadata.rs
```diff
@@ -31,7 +31,7 @@ use url::Url;
 
 use crate::util::package_ident;
 
-const SYSTEM_CARGO_BIN_PATH: &str = "cargo";
+pub(crate) const SYSTEM_CARGO_BIN_PATH: &str = "cargo";
```
acmcarther, line 34: Why pub(crate)?

impl/src/settings.rs
```diff
@@ -483,77 +468,299 @@ fn validate_settings(
...
+    .iter()
+    .filter(|pkg| metadata.workspace_members.contains(&pkg.id))
+    .collect();
+
+  let mut duplicate_binary_deps = Vec::new();
+  let mut duplicate_crate_settings = Vec::new();
+
+  for package in workspace_packages.iter() {
+    if let Some(pkg_value) = package.metadata.get("raze") {
+      let pkg_settings = RawRazeSettings::deserialize(pkg_value)?;
```
acmcarther, line 569: Did you consider parsing workspace package settings using a different type? That seems like the more natural approach, I'm guessing you have a reason.

(I argue that's the more natural approach because you can then disallow "primary" fields by omitting them in the workspace package raze settings. You could argue that this results in two settings objects needing to be maintained in order, but the current approach requires additions to "primary" settings to be included in the "is_primary" fn, which I think is comparable.)

impl/src/settings.rs
```diff
@@ -483,77 +468,299 @@ fn validate_settings(
...
-  {
-    if toml_contents.raze.is_some() {
-      eprintln!(
-        "WARNING: Both [raze] and [package.metadata.raze] are set. Using [package.metadata.raze] \
-         and ignoring [raze], which is deprecated."
-      );
+  // Check for duplication errors
+  if duplicate_binary_deps.len() > 0 {
+    return Err(anyhow!(
+      "Duplicate `raze.binary_deps` values detected accross various crates: {:?}",
```
acmcarther, line 603: Hmm..

I think we should permit duplicate binary deps provided they match. This serves as a record for humans as well, and they might be confused having to decide which toml to put the binary dependency in.

(i dont feel strongly about this)

impl/src/settings.rs
```diff
@@ -483,77 +468,299 @@ fn validate_settings(
...
-pub fn load_settings<T: AsRef<Path>>(cargo_toml_path: T) -> Result<RazeSettings, RazeError> {
-  let path = cargo_toml_path.as_ref();
-  let mut toml = match File::open(path) {
-    Ok(handle) => handle,
-    Err(err) => {
-      return Err(RazeError::Generic(err.to_string()));
-    },
+impl MetadataFetcher for SettingsMetadataFetcher {
+  fn fetch_metadata(&self, working_dir: &Path, _include_deps: bool) -> Result<Metadata> {
+    // This fetch should not require network access.
```
acmcarther, line 713: What does "should not" mean here? I feel like this should be more definitive (e.g. "does not").

Any reason to be unsure?

impl/src/settings.rs
```diff
@@ -483,77 +468,299 @@ fn validate_settings(
...
+      .unwrap_or(SYSTEM_CARGO_BIN_PATH.to_string())
+      .into(),
   };
 
-  let toml_contents = {
-    let mut contents = String::new();
-    let result = toml.read_to_string(&mut contents);
-    if let Some(err) = result.err() {
-      return Err(RazeError::Generic(err.to_string()));
+  // UNWRAP: safe due to earlier assert
```
acmcarther, line 743: Which assertion / where assertion?

impl/src/settings.rs
```diff
@@ -618,10 +859,9 @@ pub mod tests {
       .ok()
       .expect("Failed to set up temporary directory");
     let cargo_toml_path = temp_workspace_dir.path().join("Cargo.toml");
-    let mut toml = File::create(&cargo_toml_path).unwrap();
-    toml.write_all(toml_contents.as_bytes()).unwrap();
+    std::fs::write(&cargo_toml_path, &toml_contents).unwrap();
 
-    let settings = load_settings(cargo_toml_path).unwrap();
+    let settings = load_settings(cargo_toml_path, None).unwrap();
```
acmcarther, line 864: Nit: /*name_of_argument=*/None?

acmcarther (changes requested):
LGTM, just a few comments

impl/src/metadata.rs
acmcarther: Oops, I found why later, marking resolved.

impl/src/settings.rs
UebelAndre: I wasn't sure how best to do that so opted for this approach since I thought it was more straight forward. I feel there should only be 1 struct that contains all relevant settings/information that users interact with/care about. Having this be 1 struct also simplifies users ability to locate and understand settings available to them. In another change, I'd also like to add [deny_unknown_fields](https://serde.rs/container-attrs.html#deny_unknown_fields) to these structs to eliminate a sizable point of confusion which is called out in https://github.com/google/cargo-raze/issues/48. I agree having to manage two places for the same information is not ideal. Could you recommend something more structured here? Happy to make improvements 😄

impl/src/settings.rs
UebelAndre: I think if something is used by multiple packages, then it should go in workspaces' metadata. I think allowing users to specify it in multiple packages promotes some sloppy copy/pasting of configurations. I'd like for this to be well defined. I'd opt to leave it as is but if you feel stronger about this I can make the change since the benefit of your suggestion is that users can see more relevant information about their package.

PS: Ultimately, my hope is that the need for this functionality will some day go away (if Cargo ever figures out what it wants to do about these things).

impl/src/settings.rs
UebelAndre: Sorry, I know you call this out regularly, it's harder for me to catch these things given that vscode renders the names in there: 
Screen Shot 2020-12-14 at 1 16 25 PM

I'll try to be more conscious of this.

impl/src/settings.rs
UebelAndre: I must have removed it and not noticed this. I updated the error handling there.

impl/src/settings.rs
UebelAndre: I've tested this from a clean cache and it works as expected but was unsure if there were any edge cases to test as well. Updated the wording here.

UebelAndre:
@acmcarther two open questions for you but otherwise ready for your eyes again.

impl/src/settings.rs
UebelAndre: One other thing to consider, with this implementation I found it quite nice to be able to define deprecations: https://github.com/google/cargo-raze/pull/303/files#r542809410 whatever design this ends up being, I think there should be a clear, single place where deprecation warnings can be found.

impl/src/settings.rs
acmcarther: I think the right answer here depends on the expected mental model. If users associate the binary dependency with an individual sub-crate, then it might make more sense for this setting permit duplicates, since you might have multiple crates using the given binary dependency. If they view it as boilerplate for raze which is mostly independent of the subcrate in question then it might make more sense to prohibit duplicates (though, in that case I'd argue it ought to be *required* to be in the workspace root toml).

I still don't feel strongly about this though, so final decision up to you (though I'd recommend trying to find a place to document this restriction in the README if it isn't already documented).

impl/src/settings.rs
acmcarther: Hmm, alright.

impl/src/settings.rs
acmcarther: woah, that's fancy

impl/src/settings.rs
UebelAndre: I've updated the documentation. I do think you're instincts are toward the better overall outcome here, but I would rather keep the functionality of binary dependencies very minimal and find a way to remove it. Maybe I'll try to drive some RFCs forward in cargo to try and get this resolved. 

Anyway, I think it's a much smaller change to make interactions here more elegant and I'd love to find out that people feel strongly enough about it to make a PR for it 😄

examples/remote/binary_dependencies/Cargo.toml
UebelAndre: Resolving for now

UebelAndre:
@acmcarther I *think* I've addressed all the open conversations. Please let me know if I missed anything or if you have any other feedback 🙏

acmcarther:
LGTM

PR #303, opened 2020-11-23
Added deprecation warnings

UebelAndre:
This flags certain settings as deprecated. Namely:
- `incompatible_relative_workspace_path` is planned to be removed
- `target` can be replaced by `targets`
- `default_gen_buildrs` is expected to change it's default settings per https://github.com/bazelbuild/rules_rust/issues/490

Blocked by https://github.com/google/cargo-raze/pull/302

### DIFF
--- a/impl/src/settings.rs
+++ b/impl/src/settings.rs
@@ -542,7 +542,7 @@ impl RawRazeSettings {
 
     if self.incompatible_relative_workspace_path.unwrap_or(false) {
       eprintln!(
-        "WARNING: `[*.raze.incompatible_relative_workspace_path]` will is deprecated. Please set \
+        "WARNING: `[*.raze.incompatible_relative_workspace_path]` is deprecated. Please set \
          this flag to true and make sure `workspace_path` is the label where the Cargo-raze is \
          expected to write it's output."
       );
@@ -550,7 +550,7 @@ impl RawRazeSettings {
 
     if self.target.is_some() {
       eprintln!(
-        "WARNING: `[*.raze.target]` will soon be deprecated. Please update your project to use \
+        "WARNING: `[*.raze.target]` is deprecated. Please update your project to use \
          `[*.raze.targets]`."
       );
     }
--- a/impl/src/settings.rs
+++ b/impl/src/settings.rs
@@ -532,11 +532,19 @@ impl RawRazeSettings {
       || self.vendor_dir.is_some()
   }
 
-  fn print_deprecation_warnings(&self) {
-    if self.incompatible_relative_workspace_path.is_some() {
+  fn print_notices_and_warnings(&self) {
+    if self.default_gen_buildrs.is_none() {
       eprintln!(
-        "WARNING: `[*.raze.incompatible_relative_workspace_path]` will soon be deprecated. Please \
-         update your project to use the `true` behavior."
+        "NOTICE: The default of `[*.raze.default_gen_buildrs]` will soon be set to `true`. Please \
+         explicitly set this flag to prevent a change in behavior."
+      );
+    }
+
+    if self.incompatible_relative_workspace_path.unwrap_or(false) {
+      eprintln!(
+        "WARNING: `[*.raze.incompatible_relative_workspace_path]` will is deprecated. Please set \
+         this flag to true and make sure `workspace_path` is the label where the Cargo-raze is \
+         expected to write it's output."
       );
     }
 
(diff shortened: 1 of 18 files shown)


faith (approved):
Seems good to me 👍

impl/src/settings.rs
```diff
@@ -483,77 +468,341 @@ fn validate_settings(
...
+      || self.output_buildfile_suffix.is_some()
+      || self.default_gen_buildrs.is_some()
+      || self.registry.is_some()
+      || self.index_url.is_some()
+      || self.incompatible_relative_workspace_path.is_some()
+      || self.rust_rules_workspace_name.is_some()
+      || self.vendor_dir.is_some()
+  }
+
+  fn print_deprecation_warnings(&self) {
```
UebelAndre, line 528: Deprecation warnings have been added here

UebelAndre:
@acmcarther This PR is now ready for review

impl/src/settings.rs
```diff
@@ -531,6 +531,45 @@ impl RawRazeSettings {
       || self.rust_rules_workspace_name.is_some()
       || self.vendor_dir.is_some()
   }
+
+  fn print_deprecation_warnings(&self) {
+    if self.incompatible_relative_workspace_path.is_some() {
+      eprintln!(
+        "WARNING: `[*.raze.incompatible_relative_workspace_path]` will soon be deprecated. Please \
+         update your project to use the `true` behavior."
```
acmcarther, line 539: This warning should probably be more explicit about what the user should do. I looked up the field. Seems like the non-deprecated behavior is to have `workspace_path` be a valid bazel label, not a path relative to cargo.toml. Update the error to convey this?

impl/src/settings.rs
```diff
@@ -531,6 +531,45 @@ impl RawRazeSettings {
...
+    if self.target.is_some() {
+      eprintln!(
+        "WARNING: `[*.raze.target]` will soon be deprecated. Please update your project to use \
+         `[*.raze.targets]`."
+      );
+    }
+
+    if self.default_gen_buildrs.is_none() {
+      eprintln!(
+        "WARNING: The default of `[*.raze.default_gen_buildrs]` will soon be set to `true`. \
```
acmcarther, line 552: Hmm. I don't know about this one. What you're doing here is basically making the flag mandatory.

Perhaps this should be a "notice", instead of a "warning".

impl/src/settings.rs
```diff
@@ -531,6 +531,45 @@ impl RawRazeSettings {
...
+         Please explicitly set this flag to prevent a change in behavior."
+      );
+    }
+
+    if self.rust_rules_workspace_name.is_none() {
+      eprintln!(
+        "WARNING: The default of `[*.raze.rust_rules_workspace_name]` will soon be set to \
```
acmcarther, line 559: Should we include the old value of the workspace name here? Would help folks search their code to know whether or not they're affected.

acmcarther (changes requested):
Bunch of nits I don't feel strongly about.

impl/src/settings.rs
UebelAndre: The non-deprecated behavior will use `workspace_path` to determine where the output is. Previously, users would put their `Cargo.toml` files in some path and then that path would be rendered into the files. However, if that path was wrong or you for some reason didn't want to move your `Cargo.toml` file to some other directory (ie. you want to maintain compatibility with Cargo), then you had to fumble around with `--output` and write a wrapper for running raze. 

The non-deprecated behavior will always place your files in the location of that label, making it the source of truth and easier to use in that you can specify this behavior in your toml files without needing to write a wrapper or change your project structure.

I can definitely update the wording

impl/src/settings.rs
UebelAndre: I can change this to `NOTICE:`. The inspiration for adding this is in https://github.com/bazelbuild/rules_rust/issues/490

UebelAndre: And It's not making it mandatory, I'm just trying to inform users that behavior will change

UebelAndre: Updated

impl/src/settings.rs
UebelAndre: Updated

impl/src/settings.rs
```diff
@@ -531,6 +531,45 @@ impl RawRazeSettings {
       || self.rust_rules_workspace_name.is_some()
       || self.vendor_dir.is_some()
   }
+
+  fn print_deprecation_warnings(&self) {
+    if self.incompatible_relative_workspace_path.is_some() {
+      eprintln!(
+        "WARNING: `[*.raze.incompatible_relative_workspace_path]` will soon be deprecated. Please \
```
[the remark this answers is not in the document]
UebelAndre: Updated

impl/src/settings.rs
UebelAndre: Updated

UebelAndre:
@acmcarther Ready for another review

acmcarther (approved):
LGTM!

PR #306, opened 2020-11-24
Addressed user feedback for cargo workspace support

UebelAndre:
# Changes

## Renamed `workspace_member_dir` to `package_aliases_dir`
The name `workspace_member_dir`  was extremely vague and doesn't quite communicate the intent that this path is where the [aliases](https://docs.bazel.build/versions/master/be/general.html#alias) will be placed relative to where the `Cargo.toml` file is for the current package.

## Notes
This PR continues a discussion started here: https://github.com/google/cargo-raze/pull/276#issuecomment-732514620

blocked by https://github.com/google/cargo-raze/pull/301


UebelAndre:
@dae Can you reiterate your thoughts on the default setting for `workspace_member_dir`? Again, I only care about the common case, whether or not users can update it, and if users can reason about how to get the behavior they want.

UebelAndre:
@dae The changes currently have `workspace_member_dir` renamed to `package_aliases_dir` but I saw you suggested `aliases_build_file_dir` and `top_build_file_dir`, can you explain your thought process behind those names?

dae:
I was using a Cargo.toml file in $wksp/cargo, and running 'cargo raze' from that folder. Previously it created a build file in that folder; after updating it was creating it in $wksp/cargo/cargo. I was just bringing it up as it is a breaking change in my case, and potentially for others - and thus it may be worth either changing the default to match the old behaviour, or documenting it in the changelog.

I don't have strong feelings about the exact naming of the setting as long as it's documented - I was just trying to point out that after updating I found the build file was empty (because it had been relocated), so I went hunting for settings related to build files

UebelAndre:
> I was using a Cargo.toml file in $wksp/cargo, and running 'cargo raze' from that folder. Previously it created a build file in that folder; after updating it was creating it in $wksp/cargo/cargo. I was just bringing it up as it is a breaking change in my case, and potentially for others - and thus it may be worth either changing the default to match the old behaviour, or documenting it in the changelog.
> 
> I don't have strong feelings about the exact naming of the setting as long as it's documented - I was just trying to point out that after updating I found the build file was empty (because it had been relocated), so I went hunting for settings related to build files

Can you explain a bit about why your cargo manifest file is in `$wksp/cargo`? Could all that information live in https://github.com/ankitects/anki/blob/master/rslib/Cargo.toml ? And again, I'm not trying to defend the current state of things, just trying to understand existing use cases.

Though, I do think the answer to my question here (https://github.com/google/cargo-raze/pull/276#issuecomment-732581070) seems to "yes".

dae:
The README in this repo talks about generating a Cargo.toml file instead of referencing an existing one, so I assumed standard practice was to gather all deps used in a monorepo into a single Cargo.toml file, so that a single aliases file is generated. Then for each Rust package in the monorepo, you can refer to deps with //cargo:dep_name.

I'm just trying out the new workspace code at the moment, and it instead seems to take the approach of putting an aliases file into each package of the workspace. I can see this having an advantage of ensuring you don't reference crates not mentioned in that package's Cargo.toml, but I did find a single //cargo location convenient.

Please note I'm not frustrated, I was mentioning the issue I ran into, as if other people will hit it as well it may be worth mentioning in the changelog. If I was doing something non-standard, fair enough. :-)

UebelAndre:
> The README in this repo talks about generating a Cargo.toml file instead of referencing an existing one, so I assumed standard practice was to gather all deps used in a monorepo into a single Cargo.toml file, so that a single aliases file is generated. Then for each Rust package in the monorepo, you can refer to deps with //cargo:dep_name.
> 
> I'm just trying out the new workspace code at the moment, and it instead seems to take the approach of putting an aliases file into each package of the workspace. I can see this having an advantage of ensuring you don't reference crates not mentioned in that package's Cargo.toml, but I did find a single //cargo location convenient.

The whole purpose of `package_aliases_dir` was to support exactly what you want for mono repos with multiple single package crates. However, when you're using workspaces, a single location for aliases doesn't quite work. For example, you might have two crates that depend on the same external crate but at different versions. Currently, only one alias is made which means if you were to refer to that target as `//cargo:external_crate` you'd be using the incorrect version for one of your cases. You can see this in https://github.com/google/cargo-raze/pull/277. Do you have any suggestions on how we could use a single alias location and still have managing dependencies be easy and clear for users? I do feel the way things are now is a bit clunky.

> Please note I'm not frustrated, I was mentioning the issue I ran into, as if other people will hit it as well it may be worth mentioning in the changelog. If I was doing something non-standard, fair enough. :-)

🙏 😄 Thanks for checking out these changes! I've updated the documentation and made a tweak to the README. Do you feel this is enough of a breadcrumb to lead people toward the desired behavior?

UebelAndre:
Oh, worth noting, only the last two commits in this draft are relevant to the conversation here. This change was built on top of another draft PR.

dae:
Good point about aliases to different versions. I don't see an easy way that could be handled in a single location without some sort of renaming magic, and using separate alias build files is not particularly difficult now that I've tried it (though I do wish Bazel would allow us to refer to them relatively, instead of having to list out the absolute package path)

Doc updates look good, but since my original report I've switched over to using a workspace (which seems to be working well - thank you for your work on that!)

acmcarther:
@UebelAndre 

I was somewhat concerned about BUILD files not being generated into the working directory in the relevant PR because that was my workflow as well.

Is it possible to restore that behavior. I realize that dae@ is OK with switching to the new behavior (via adding a workspace), but I'd really prefer to avoid this type of breaking change which requires users to dig through documentation.

UebelAndre:
> @UebelAndre
> 
> I was somewhat concerned about BUILD files not being generated into the working directory in the relevant PR because that was my workflow as well.
> 
> Is it possible to restore that behavior? I realize that dae@ is OK with switching to the new behavior (via adding a workspace), but I'd really prefer to avoid this type of breaking change which requires users to dig through documentation.

@acmcarther That behavior is still possible by setting `package_aliases_dir = "."` in your raze settings. The open question is if that's the common case. I would understand if it's the common case right now, but I deal with more people using raze to convert existing cargo projects over to bazel and it's been a point of frustration that raze will overwrite existing files without warning. This is why it defaults to something that's not the same directory as `Cargo.toml` file. What're your thoughts there?

acmcarther:
I think the default should therefore be to make `package_aliases_dir = "."`  so that existing behavior is preserved, even if it's strictly inferior. It's important to me that we avoid disrupting existing workflows.

For me the most important thing is to avoid the experience described in https://github.com/google/cargo-raze/pull/306#issuecomment-732647630 where dae had something mysterious happen and then had to do a bunch of digging.

I don't think updating the documentation is sufficient to mitigate this concern.

UebelAndre:
> I think the default should therefore be to make `package_aliases_dir = "."` so that existing behavior is preserved, even if it's strictly inferior. It's important to me that we avoid disrupting existing workflows.
> 
> For me the most important thing is to avoid the experience described in [#306 (comment)](https://github.com/google/cargo-raze/pull/306#issuecomment-732647630) where dae had something mysterious happen and then had to do a bunch of digging.
> 
> I don't think updating the documentation is sufficient to mitigate this concern.

Updated. Though that then raises the question, do you think the examples are a correct representation of how projects should be set up?

UebelAndre:
After some reflection, I've come up with something I think is quite nice. Instead of rendering aliases at all, we can can add a macro to the `crates.bzl` file that maps a package to it's dependencies. So instead of rendering an `alias`, a macro is generated and called to get a list of dependencies. Once some of the open PRs go through, I'll open a PR for this functionality.

Side note: I feel like these draft PRs have just added confusion but my goal was to create insight into what I was working on. Maybe they're still helpful but maybe opening too many is bad? Interested in some thoughts here.

acmcarther:
I haven't spent much time (except for this CL for some reason... I guess because dae@ chimed in) reviewing the drafts. The diff is not friendly to read because it contains the other changes.

It might be helpful if I was more proactive in the reviews, but I'm operating strictly in an "emergency reviewer" capacity, meaning I review things slowly, only when they're literally blocked on me.

UebelAndre:
How many other maintainers are there for this project? It seems to me that you're the only one doing reviews. Which I greatly appreciate but I think the less others do reviews, the less likely they are to ever do them or to be able to understand the impact of a change. Are all maintainers reflected in the code owners file? And maybe you could ping a few of them to see if they have time to do reviews as well? Just a thought.

Again, I'm making a focused effort to keep my PRs small so hopefully it gets easier to do reviews from here on (since it doesn't seem like others are opening large PRs)

UebelAndre:
@acmcarther This PR is ready for review now

acmcarther:
LGTM

PR #311, opened 2020-12-03
Add data deps attrs to CrateDependencyContext

sitaktif:
Specifying data dependencies is needed for some projects.

This PR only adds the corresponding attributes to the data structures, allowing dependencies of cargo-raze to use them.

This is needed for an upcoming contribution that @illicitonion is working on.

### DIFF
--- a/impl/src/rendering/templates/partials/common_attrs.template
+++ b/impl/src/rendering/templates/partials/common_attrs.template
@@ -20,9 +20,16 @@
     {%- if target.kind != "bin" %}
     crate_type = "{{ target.kind }}",
     {%- endif %}
-    {%- if crate.raze_settings.data_attr %}
-    data = {{crate.raze_settings.data_attr}},
-    {%- endif %}
+    data = []
+    {%- if crate.raze_settings.data_attr %} + {{crate.raze_settings.data_attr}}
+    {%- endif -%}
+    {# Note: data_dependencies can only be set when using cargo-raze as a library at the moment #}
+    {%- if crate.default_deps.data_dependencies %} + [
+    {%- for dependency in crate.default_deps.data_dependencies %}
+        "{{dependency.buildable_target}}",
+    {%- endfor %}
+    ]
+    {%- endif %},
     edition = "{{ target.edition }}",
     {%- if crate.default_deps.proc_macro_dependencies %}
     proc_macro_deps = [
--- a/examples/remote/binary_dependencies/cargo/remote/BUILD.addr2line-0.14.0.bazel
+++ b/examples/remote/binary_dependencies/cargo/remote/BUILD.addr2line-0.14.0.bazel
@@ -39,6 +39,7 @@ rust_library(
     ],
     crate_root = "src/lib.rs",
     crate_type = "lib",
+    data = [],
     edition = "2015",
     rustc_flags = [
         "--cap-lints=allow",
--- a/examples/remote/binary_dependencies/cargo/remote/BUILD.adler-0.2.3.bazel
+++ b/examples/remote/binary_dependencies/cargo/remote/BUILD.adler-0.2.3.bazel
@@ -39,6 +39,7 @@ rust_library(
     ],
     crate_root = "src/lib.rs",
     crate_type = "lib",
+    data = [],
     edition = "2015",
     rustc_flags = [
         "--cap-lints=allow",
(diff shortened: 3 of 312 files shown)


UebelAndre:
Can this not also be solved with the awkward setting, `data_attr`?
https://github.com/google/cargo-raze/blob/master/impl/src/settings.rs#L203-L209
```rust
  /**
   * The verbatim `data` clause to be included for the generated build targets.
   *
   * N.B. Build scripts are always provided all crate files for their `data` attr.
   */
  #[serde(default = "default_crate_settings_field_data_attr")]
  pub data_attr: Option<String>,
```

Also, I think you'd need to update the templates for this to take effect, no?

UebelAndre:
Oh, https://github.com/bazelbuild/rules_rust/issues/2#issuecomment-738803545 gives some great context for this.

sitaktif:
> Oh, [bazelbuild/rules_rust#2 (comment)](https://github.com/bazelbuild/rules_rust/issues/2#issuecomment-738803545) gives some great context for this.

Yes sorry I was a bit terse on the context, I just wanted to wait for the info to be official before being too specific.
This change allows rules_rust_external to store data (and build script data) attributes when generating rules for external crates. It has its own set of templates for this. There isn't any feature today to set these values in cargo raze but if this gets implemented one day, it should certainly store the values in the `CrateDependencyContext` struct as suggested here.

> Can this not also be solved with the awkward setting, data_attr?

From what I can tell, `CrateDependencyContext` is the right layer to put the information for these two properties (next to regular deps, proc macro deps, build deps, etc). The attribute is not meant to represent attributes directly parsed by serde but rather a safe, decoupled value passed to the constructor, so I don't think it quite fits `CrateSettings`. Also the fact that it's a raw string is a bit awkward as you mentioned earlier :-)

UebelAndre:
Is there anything currently in `cargo-raze` that would use these variables? It feels kinda odd to be adding unused fields maybe they could be plugged into something in this PR? I at least think they should be added to the templates so if they were populated they'd get rendered correctly. I imagine this would be a small, maybe 6 line change?

sitaktif:
@UebelAndre that sounds very reasonable :-)

Sorry I didn't get to work on that today, I'll update the PR hopefully tomorrow.

sitaktif:
Here we go, PTAL :-)

This changes the BUILD files by adding a `data = []` when no data is set anywhere because the templates would be quite a bit less readable otherwise (having two optional lists makes things somewhat messy on the template side).

I have manually tested the case where the vec is set to some values, things seem to be looking ok, I put them in https://gist.github.com/sitaktif/b0d5c37e5c710e9ab5705674f816fbda so you can have a look.

The case where no data is set can already be checked in this PR (I put it as a separate commit for clarity but I'm also happy to merge the two commits if you prefer).

acmcarther:
I think you might need to push the changes (or, alternatively, I might be blind). All I see are the new fields.

As is, this LGTM, but yeah, probably best to include the BUILD stuff too.

sitaktif:
Ha sorry, I did that just before going to bed and didn't check that properly 🙄 Changes are in now.

impl/src/rendering/templates/partials/common_attrs.template
```diff
@@ -20,9 +20,16 @@
     {%- if target.kind != "bin" %}
     crate_type = "{{ target.kind }}",
     {%- endif %}
-    {%- if crate.raze_settings.data_attr %}
-    data = {{crate.raze_settings.data_attr}},
-    {%- endif %}
+    data = []
+    {%- if crate.raze_settings.data_attr %} + {{crate.raze_settings.data_attr}}
+    {%- endif -%}
+    {# Note: data_dependencies can only be set when using cargo-raze as a library at the moment #}
```
UebelAndre, line 26: I feel like this comment would also be nice to have in `Context.rs` or something

sitaktif: Sounds good, I added comments as suggested

acmcarther:
LGTM. I'm very much looking forward to see the followup.

sitaktif:
> I'm very much looking forward to see the followup.

Just to be sure, are you talking about https://github.com/bazelbuild/rules_rust/issues/2#issuecomment-738803545 or something else?

acmcarther:
Yeah, I'm referring to https://github.com/bazelbuild/rules_rust/issues/2#issuecomment-738803545

PR #317, opened 2020-12-10
Disabled tools so rust-analyzer stops trying to parse them.

UebelAndre:
The existence of this crate makes rust analyzer very upset. Can we prevent it from being recognized as a crate until someone decides to update it? (This PR prevents it from being recognized as a crate).


acmcarther:
This SGTM

Issue #313, opened 2020-12-08
Need to migrate from Travis CI to another provider


acmcarther:
Due to some touch-and-go permissions-related stuff[1] with Travis CI migrating travis-ci.org usages from the org domain to the com domain, this project needs to prepare a migration to an alternative CI provider.

BuildKite is probably the best contender, since we already have it for rules_rust. That's running under organization `bazelbuild` though, so there might be some work involved. Some other alternatives are GitHub Actions and CircleCI.

(heads up for you @UebelAndre)

---

Deadline on this is Dec 31st 2020[1]

[1] Travis folks are currently requiring `repo` scope in the Oauth token grant for users migrating to the `.com` instance. This is prohibited by a Google organization policy.
[2] https://docs.travis-ci.com/user/migrate/open-source-repository-migration#q-when-will-the-migration-from-travis-ciorg-to-travis-cicom-be-completed

acmcarther:
CI might get flaky in the lead up to the 31st both because I'll be messing with it and also because the Travis CI folks might be doing migration stuff.

UebelAndre:
I feel like the repo is in pretty decent shape to migrate to Bazel (especially since https://github.com/google/cargo-raze/pull/300 was merged which eliminates the need for external dependencies). The smoke tests are the only mystery for me but I think I have a path forward there. I've never setup anything with BuildKite but I'd expect it to be pretty straight forward if the project is building with Bazel. But let me know if you think that's a good idea.

@acmcarther Thanks for the heads up 🙏 😄

UebelAndre:
Though that said, I have a couple of small changes and 2 medium changes I'd like to get through. If you could review them for me I can work on a Bazel migration this weekend. I've als been really hoping for a release with workspace support 🙏

UebelAndre:
I think i finally got openssl to build correctly in Bazel over the weekend. I'm working on a few more crates that I might be able to get working before the end of the week. If anyone is interested in the progress. Check out https://github.com/UebelAndre/cargo-raze/tree/bazel

UebelAndre:
@acmcarther Building with Bazel now works! There's still some issues compiling the tests (namely [basic-cookie](https://crates.io/crates/basic-cookies)) but if I can get that working then I just need to write a test rule that checks to see `examples` are only the result of running `cargo-raze` (no user modifications directly committed to that directory) and then I think it's ready. 

[...]

PR #319, opened 2020-12-16
Add experimental api `crate_deps` and `all_crate_deps` macro for gathering package dependencies

UebelAndre:
A new experimental API has been added for accessing crates. If `experimental_api = true` is set in `RazeSettings`, then the following macros will be rendered into the `crates.bzl` file for both Remote and Vendored genmodes.
```python
def crate_deps(deps, package_name = None):
    """EXPERIMENTAL -- MAY CHANGE AT ANY TIME: Finds the fully qualified label of the requested crates for the package where this macro is called.

    WARNING: This macro is part of an expeirmental API and is subject to change.

    Args:
        deps (list): The desired list of crate targets.
        package_name (str, optional): The package name of the set of dependencies to look up.
            Defaults to `native.package_name()`.
    Returns:
        list: A list of labels to cargo-raze generated targets (str)
    """
    # ...

def all_crate_deps(normal = False, normal_dev = False, proc_macro = False, proc_macro_dev = False, package_name = None):
    """EXPERIMENTAL -- MAY CHANGE AT ANY TIME: Finds the fully qualified label of all requested direct crate dependencies \
    for the package where this macro is called.

    If no parameters are set, all normal dependencies are returned. Setting any one flag will
    otherwise impact the contents of the returned list.

    Args:
        normal (bool, optional): If True, normal dependencies are included in the
            output list. Defaults to False.
        normal_dev (bool, optional): If True, normla dev dependencies will be
            included in the output list. Defaults to False.
        proc_macro (bool, optional): If True, proc_macro dependencies are included
            in the output list. Defaults to False.
        proc_macro_dev (bool, optional): If True, dev proc_macro dependencies are
            included in the output list. Defaults to False.
        package_name (str, optional): The package name of the set of dependencies to look up.
            Defaults to `native.package_name()`.

    Returns:
        list: A list of labels to cargo-raze generated targets (str)
    """
    # ...
```

Examples of the macros can be seen below:
```python
load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary", "rust_library", "rust_test")
load("//impl/cargo:crates.bzl", "all_crates", "crates")

package(default_visibility = ["//visibility:public"])

rust_library(
    name = "cargo_raze_lib",
    srcs = glob(["src/**/*.rs"], exclude = ["src/bin/cargo-raze.rs"]),
    # Explicitly gather specific crate targets relevant to the current package with the given names
    proc_macro_deps = crates([
        "indoc",
        "serde_derive",
    ]),
    # Gather all crate targets relevant to the current package
    deps = all_crates(),
)

rust_binary(
    name = "cargo_raze_bin",
    srcs = [
        "src/bin/cargo-raze.rs",
    ],
    deps = [":cargo_raze_lib"] + all_crates(),
    proc_macro_deps = all_crates(proc_macro = True),
)

rust_test(
    name = "cargo_raze_lib_test",
    crate = ":cargo_raze_lib",
    deps = all_crates(normal_dev = True),
    proc_macro_deps = all_crates(proc_macro_dev = True),
)

rust_test(
    name = "cargo_raze_bin_test",
    crate = ":cargo_raze_bin",
    deps = all_crates(normal_dev = True),
    proc_macro_deps = all_crates(proc_macro_dev = True),
)

```

This should make it easier for users to manage their dependencies in situations where they have multiple `Cargo.toml` files representing in their workspace (ie: they're using [cargo workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)).


impl/src/testing.rs
```diff
@@ -295,35 +293,14 @@ pub fn mock_crate_index(
 }
 
 /** Generate some basic metadata with an injected mock dependency */
-pub fn dummy_modified_metadata() -> Metadata {
-  let mut metadata = dummy_raze_metadata().metadata.clone();
+pub fn dummy_modified_metadata() -> RazeMetadata {
```
UebelAndre, line 296: For this change I needed to update the tests using it and opted to remove this function and add `dummy_modified_metadata.json.template`  which is a direct dump (with template's being added for things like system paths) of the metadata generated by this file before.

impl/src/context.rs
```diff
@@ -128,6 +128,7 @@ pub struct CrateContext {
   pub features: Vec<String>,
   pub workspace_path_to_crate: String,
   pub workspace_member_dependents: Vec<PathBuf>,
+  pub workspace_member_dev_dependents: Vec<PathBuf>,
```
UebelAndre, line 131: This field is the cause of code changes outside of `rendering::bazel.rs`. I needed to be able to differentiate development and normal dependencies when rendering.

impl/src/testing.rs
UebelAndre: I've decided to spin this off into https://github.com/google/cargo-raze/pull/321

When that's merged I'll be rebasing these changes on top of that @acmcarther just FYI

impl/src/rendering/templates/partials/crates_macro.template
```diff
@@ -0,0 +1,183 @@
+# A mapping of package names to a set of normal dependencies for the Rust targets of that package.
```
acmcarther, line 1: For each of these maps, does it make sense to include a description of the key and value?

acmcarther, line 1: What kind of stability guarantee are we making for these maps? What about the functions? Their return types? Their argument sets?

---

I don't have strong opinions about any of these (except a general bias toward "stable" and "thought-out").

impl/src/rendering/templates/partials/crates_macro.template
```diff
@@ -0,0 +1,183 @@
...
+{%- if crate.is_proc_macro and workspace_member in crate.workspace_member_dev_dependents %}
+{%- set crate_name_sanitized = crate.pkg_name | replace(from="-", to="_") %}
+        "{{ crate.pkg_name }}": "{{ crate.workspace_path_to_crate }}:{{ crate_name_sanitized }}",
+{%- endif %}
+{%- endfor %}
+    },
+{%- endfor %}
+}
+
+def crates(deps):
```
acmcarther, line 73: What do you think about splitting this out into a separate `bzl` file which is not a template, and is just copied during generation? That way it can be edited and linted with regular tools.

I recognize that it depends on the definitions of the maps. These could be brought in via `load` I think, with a hard coded path local to the file itself, like this:

```bzl
load(":crates.bzl",
     "DEPENDENCIES", 
     "PROC_MACRO_DEPENDENCIES", 
     "DEV_DEPENDENCIES", 
     "DEV_PROC_MACRO_DEPENDENCIES")
```

impl/src/context.rs
```diff
@@ -136,6 +137,7 @@ pub struct CrateContext {
   pub source_details: SourceDetails,
   pub sha256: Option<String>,
   pub registry_url: String,
+  pub is_proc_macro: bool,
```
acmcarther, line 140: Is this the right place for this? IIUC, "proc_macro"-ness is a property of a target within the crate, not  the crate itself. I expect that a proc macro should contain a target with the corresponding kind.

acmcarther (changes requested):
I haven't conducted a real serious review yet. Couple questions/suggestions about the generation of the bzl file.

impl/src/rendering/templates/partials/crates_macro.template
UebelAndre: I don't think I quite understand. The functions have a full docstring explaining what they're expected to produce with explicit failure modes (some one or two which might be too aggressive). These maps should always be present in this file and I think bucketing the cargo dependencies into bazel packages works quite nicely since the definition of both a cargo and bazel pacakge are directory based (more or less in the case of cargo but definitely works for this case).

impl/src/rendering/templates/partials/crates_macro.template
UebelAndre: `crates.bzl` is buildifier compliant (proven by the passing CI) or maybe I misunderstand what you mean.

Can you also elaborate on what you're trying to accomplish by loading each map individually? I almost made them private but didn't want to be restrictive. Users should only ever interact with the `crates` and `all_crates` macros.

impl/src/context.rs
UebelAndre: This is meant to go along with `pub lib_target_name: Option<String>,` which is currently set by both `lib` and `proc-macro` targets but the distinction is important to be able to identify later. Though the placement of that variable doesn't quite convey that so I can update that if you think this is something that could stay.

impl/src/rendering/templates/partials/crates_macro.template
UebelAndre: The maps should all be the same but contain targets of differing significant. So explaining one should be sufficient. Also, I touched on this in another comment. I felt these maps should be private and users should be made to go through the interface macros `crates` and `all_crates` but I didn't want to be too restrictive.

UebelAndre:
@acmcarther some questions for you.

examples/remote/cargo_workspace/cargo/crates.bzl
```diff
@@ -9,6 +9,161 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+
+    for dep_map in all_maps:
+        for package_name in dep_map:
+            if package_name in dependencies:
+                dependencies[package_name].extend(dep_map[package_name])
+            else:
+                dependencies[package_name] = dep_map[package_name]
+
+    if not dependencies:
+        fail("No crates have been rendered. Use of this macro should be removed")
```
UebelAndre, line 160: Playing around with this some more. I've decided not to have this be an error. It's useful to allow users to stub out their dependencies such that if a dependency were to be added in the future, this would automatically add the dependency to the crate.

Once https://github.com/google/cargo-raze/pull/321 is closed I'll rebase and have a commit that addresses this.

impl/src/rendering/templates/partials/crates_macro.template
acmcarther: You're asserting here that we will have these values in this file indefinitely, but that's clearly not true. In the other comment you weren't sure whether or not users would even need the map you're providing.

Values that are being provided here are essentially part of our public api and will basically need to be preserved indefinitely (or, a migration approach provided if we find the need to change this).

The question remains -- what type of stability guarantee are we providing about the build macros here, about their arguments, and about their return types. If you're not sure, consider annotating these "experimental", so that we can break them later with impunity.

impl/src/rendering/templates/partials/crates_macro.template
acmcarther: My request here is for you to separate the non-dynamic parts of this file from the generated parts of this file. I think this is easier to understand for future maintainers because the crate maps are simple generated maps, and the functions are just functions in a bzl file. Developers will be able to easily and automatically run buildifier on the bzl file. This approach also eliminates the temptation of doing code generation inside of the bzl functions, which I think is a perilous temptation which would inevitably result in a large increase in complexity.

acmcarther (changes requested):
Perhaps I shouldn't say this, but I feel based on this and prior discussions, that you take the concept of maintaining API compatibility too lightly. On prior occasions I've had to make the point against making sudden breaking changes, and I think here you're underestimating the new type of commitment-to-stability being made by generating this bzl file.

Perhaps I am too paranoid about this type of thing (or perhaps I've been traumatized too often by maintainers of upstream tools with this same attitude).

acmcarther:
Part of what might be accomplished by splitting the map generation out from the functions, is that it makes clear that there's another option here -- move the functions here into a bzl library which can be included and versioned separately. Despite my pushback about the map public/private-ness, I think we can much more easily make guarantees about stability about the map than we can about stability of the generated functions. Moving those functions into a separately versioned bzl library allows end users to deal with breakages there independent of the raze updates.

UebelAndre:
@acmcarther I think you've lost me here. I don't understand the motivation for splitting the maps into their own files and I don't understand what you're referring to with the "experimental" annotation. To me, the giant banner in the README claiming this is not a google product communicates to me that there's some volatility here. I don't believe the maps are significant to users and I think it's better to make them private (just wasn't originally confident enough to assert that). I think my confusion primarily stems from my assumption that the information being put here is the exact same that's being rendered into the neighboring BUILD file in form of aliases. Perhaps you could elaborate on your concerns about stability?

acmcarther:
There are two questions here:

1. What is the value in splitting the generation of the CRATE maps out from the rest of the bzl file

and 

2. What is the concern about "api stability" that I keep harping on about.

---

### What is the value in splitting the generation of the CRATE maps out from the rest of the bzl file

Described above:
- Separate the dynamic component of the generated output from the static component. I think may be easier for readers to understand.
- Put a "soft" guarantee in place that we will not generate "non-data" starlark in the future. I think that's a terrible design rabbit hole.
- Make it easier for future developers to maintain this and keep it formatted (because the files will be simple bzl files, not templates)
- Open the door to (potentially a future change) where cargo-raze consists of both a generation component and a bzl library component.

### What is the concern about "api stability" that I keep harping on about.
This PR introduces new API surface area. As a general practice, when introducing API surface area, you have to be careful because users may begin depending on it immediately, and it can limit future evolution of the software. A guideline I follow is to add the minimal api surface area that works for the user requirements only when users actually need the new behavior.

Though this contains the same "data" as what is in the generated build files it represents a new mechanism (by more than one definition of mechanism I think) by which end users can access and depend upon this data.

---

I would prefer if you distinguish in your reply between the arguments above not being convincing, versus them just not being clear. I think repeatedly in this PR you've mean the former, but you've expressed it as the latter.

UebelAndre:
@acmcarther Heads up, just rebased. That will be the only time I do this for this PR.

UebelAndre:
re: https://github.com/google/cargo-raze/pull/319#issuecomment-749304056

Everything up until that comment was not clear to me. I replied based on my understanding of what you wrote but wasn't sure if we were on the same page.

I think related to both questions, I find it incredibly difficult to understand the impact of any change to this tool. From what I've gathered, other contributors are the only other users who use the latest releases. Releasing a new version with some change in behavior shouldn't break companies using this since they can easily install the tool with `--version`. I would love to be able to roll this out with some experimental framing so that the API can change if someone thinks of something smarter but I do feel this add quite a nice quality of life improvement to the tool in that users only need to update dependencies in one place. Because I want these changes to be viewed as experimental, I'm hesitant to do a bunch of work to formalize this without thoroughly understanding the pros and cons of the current API (which should really only be the `crates` and `all_crates` macro, going back to the public/private conversation about the maps). 

My question from here is how can I frame this API as experimental such that there's still flexibility to change it and I don't have to add new functionality to copy static files vs rendering semi static templated ones until there's more confidence in this API?

acmcarther:
Ah I think I see what you're saying, and why we differ on our assumption of impact.

#### On the philosophy
I think generally I have begun (since the ancient times of yore whence the first third party contributors appeared) to assume that some (ill advised) organizations have built nightmare scaffolding[1] upon what this tool generates.  In practice, I've heard mixed signals from folks (some, such as Marco, expressing that they've been able to deal with incompatibilities easily, some from internal teams expressing appreciation that I keep an eye out for breaking changes)

#### On unblocking this change
I don't know if this is a standard practice, but my suggestion above (re "marking this experimental"), can mean one of two things. Either extend the bzl macro documentation to say "EXPERIMENTAL -- MAY CHANGE AT ANY TIME", or prefix the function with "experimental". There may be a third option which is more conventional... I'm not experienced enough with bzl library maintenance to know for sure.

---

How do you feel about going with the "mark this experimental for now" approach. I like this because I feel less bad if users depend upon it and then we break them.

[1] meaning, incredibly complex bzl functions, code generators, and other shenanigans, likely not maintained by the original author. My day job is feature development for a c++ product, but I actually spend a lot of time dealing with 10+ year old "nightmare scaffolding" -- where we're just the middle layer between our dependencies and external users. It's terrible!

UebelAndre:
re: https://github.com/google/cargo-raze/pull/319#issuecomment-749310756

I'm happy to update the documentation. I don't like changing the name of the names of the macros only because I think `experimental_*` looks bad 😆 but can be convinced. Are we in agreement on adding some "EXPERIMENTAL" banners on these macros and making the maps private?

UebelAndre:
re: https://github.com/google/cargo-raze/pull/319#issuecomment-749312311

@acmcarther Updated the API and added some comments. I hope that's what you wanted, if not, looking forward to your reply 😄

impl/src/context.rs
UebelAndre: I've updated the placement of this variable and made a comment. Hopefully this makes it more clear.

impl/src/rendering/templates/partials/crates_macro.template
UebelAndre: I think the design has changed to not require this. Gonna resolve it but definitely reopen it if you feel otherwise.

impl/src/rendering/templates/partials/crates_macro.template
UebelAndre: They've been made private

examples/remote/binary_dependencies/BUILD.bazel
```diff
@@ -8,7 +9,5 @@ rust_binary(
     data = [
         "//remote/binary_dependencies/cargo:cargo_bin_texture_synthesis",
     ],
-    deps = [
-        "//remote/binary_dependencies/cargo:ferris_says",
-    ],
+    deps = all_crates(),
```
acmcarther, line 12: Why is this change being made? Do you think end users should just include "all crates"?

I'm against this because I'm generally against making dependencies opaque anywhere. libraries and binaries should include what they need and only what they need.

examples/remote/cargo_workspace/cargo/crates.bzl
```diff
@@ -9,6 +9,162 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+
+# EXPERIMENTAL: A mapping of package names to a set of proc_macro dev dependencies for the Rust targets of that package.
+_DEV_PROC_MACRO_DEPENDENCIES = {
+    "remote/cargo_workspace/num_printer": {
+    },
+    "remote/cargo_workspace/printer": {
+    },
+    "remote/cargo_workspace/rng": {
+    },
+}
```
acmcarther, line 53: I can't unresolve the other comment because I cannot access it after a rebase.

I view the privacy of these values as separate from the bzl file generation issue, and so I don't think that issue is resolved.

examples/remote/cargo_workspace/cargo/crates.bzl
```diff
@@ -9,6 +9,162 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+
+def crates(deps):
+    """EXPERIMENTAL: Finds the fully qualified label of the requested crates for the package where this macro is called.
```
acmcarther, line 56: I prefer my verbiage here: EXPERIMENTAL -- MAY CHANGE AT ANY TIME.

acmcarther (changes requested):

examples/remote/binary_dependencies/BUILD.bazel
UebelAndre: > Why is this change being made? Do you think end users should just include "all crates"?

Here I'm demonstrating what is possible with the new macros, but yes, I think it's nicer to be able to simply refer to your dependencies via `all_crates` rather than having to track your dependencies twice, once in the `Cargo.toml` file and again in your BUILD files. I also don't think it's unclear what dependencies are being used at build time since the way `cargo-raze` functions is everything is written out before Bazel interracts with it. Any confusion can be resolved by reading the `crates.bzl` file and looking for your package name in the maps. I generally agree with the value of being explicit about dependencies, but I don't think that philosophy plays nicely with a tool that generates dependencies.

acmcarther:
My 'why are you using "all_crates()" in deps' comment may seem out of the blue given that I previously "reviewed" this PR, but in practice my prior review was mostly syntatic. I'm now actually examining what this PR is doing. I think when I previously reviewed it, I assumed that the "all_crates" rule would be used differently than it is currently being used.

Can you motivate the inclusion of this feature with some example or examples? Under what circumstances would you expect users to want to include "all_crates()" as a dependency?

My context here is that of a large organization where the Cargo.toml doesn't represent dependencies for one binary, but rather represents the collective dependencies for a whole set of microservices and libraries, where none would be expected to include every dependency. Perhaps this is geared toward a different type of use case?

examples/remote/cargo_workspace/cargo/crates.bzl
UebelAndre: That seems unnecessary. I feel that these maps should live in `crates.bzl` until there's a good reason to move them out. I don't see this as any different than the giant macro that gets rendered into the same file for defining each dependency in `genmode="Remote"`. Given that the maps are private and there are experimental notes on each component of the new API, I think it's best to keep everything in one spot.

examples/remote/cargo_workspace/cargo/crates.bzl
UebelAndre: Updated

examples/remote/cargo_workspace/cargo/crates.bzl
acmcarther: I believe I've made my case very clearly several times, and you have not made an effort to address any of the individual points.

acmcarther:
I'm somewhat inclined to argue against submission of this PR in general. I don't think the case for this new functionality has been made clearly (in contrast with PRs that, for example, add support for cargo workspaces), and it's not really clear (1) what the alternatives are, and (2) what the "cost of doing nothing" is.

I think I would be much more open to a more limited version of this which exports the dependency lists (publically).

UebelAndre:
> My context here is that of a large organization where the Cargo.toml doesn't represent dependencies for one binary, but rather represents the collective dependencies for a whole set of microservices and libraries, where none would be expected to include every dependency. Perhaps this is geared toward a different type of use case?

If the `Cargo.toml` represents more than just the dependencies for a single package, then this API should probably gain an additional macro to specifically handle this case or the feature should not be used. I feel with the introduction of workspace support here that the need to lump all your dependencies into a single `Cargo.toml` file is a matter of personal choice vs necessity now.

> Can you motivate the inclusion of this feature with some example or examples? Under what circumstances would you expect users to want to include "all_crates()" as a dependency?

I don't think it's a secret that interacting with external crates is clunky and leaves much to be desired. A month or so ago I was working on a project that I converted to use Rust. This was great at first until I realized I wanted to add a new crate for a new feature I was trying to add. I then had to add it to my `Cargo.toml` file, then regenerate raze, then go plug it into my target. This was fine for the occasional new feature but when I tried starting something new from scratch, it was awful. I was constantly writing things twice and running Raze over and over. The goal if of this macro is to solve for the pain I felt there. I should not have to add a dependency twice when `cargo-raze` is able to do that for me. Sure, it's not as fancy as some gazelle generated BUILD file that has the correct dependencies explicitly written for me. But it is enough to know that if I want to know what my dependencies are, I should only have to look in one place, the `Cargo.toml` file. If there are cases that this is not true and users are confused as to what's being included, they can always look in `crates.bzl` and file a new issue explaining the usecase. 

> My context here is that of a large organization where the Cargo.toml doesn't represent dependencies for one binary, but rather represents the collective dependencies for a whole set of microservices and libraries, where none would be expected to include every dependency. Perhaps this is geared toward a different type of use case?

This is another reason I felt workspace support was critical for this tool. I would assert it's bad practice to dump all your dependencies in one place like that when there's tooling like what's in cargo where you can have multiple manifests but generate a single lockfile. This is far more maintainable IMO.

UebelAndre:
@acmcarther This feature is for people who are actively growing their Rust ecosystem in Bazel. The cost of doing nothing is zero for large companies that have a "one-and-done" integration effort in that they were able to generate something that works and the need for new crates is rare or will basically not be necessary. At places where there's growing amount of Rust, this feature solves for a large point of frustration.

acmcarther:
I think we have different mental models for how bazel integrates with cargo (with mine being far more rigid and probably outdated). I think this is the root cause of my negative attitude here. I'd like to better understand how you or your organization think about rust crates within the bazel workspace.

How do applications in your organization map into the main repository? Is there one cargo.toml representing the main cargo-workspace, and then one workspace crate for each microservice or binary? I think I can better understand the need for this feature if your workflow works like that, because adding a new application (and workspace subcrate) would require some extra dependency-related toil that this alleviates.

examples/remote/cargo_workspace/cargo/crates.bzl
UebelAndre: > Separate the dynamic component of the generated output from the static component. I think may be easier for readers to understand.

> Put a "soft" guarantee in place that we will not generate "non-data" starlark in the future. I think that's a terrible design rabbit hole.

> Make it easier for future developers to maintain this and keep it formatted (because the files will be simple bzl files, not templates)

My response to all of these is that the change is intended to be a small change that adds an experimental API to try and ease the pain for users who frequently add new crates to their projects. I don't think the state it's in screams to other contributors that they should be following a bad pattern.

> Open the door to (potentially a future change) where cargo-raze consists of both a generation component and a bzl library component.

I'm very interested in working on something like this and before jumping into building out new functionality for two categories of outputs from this tool, I'd like to know that the API works. This is the same thing that `rules_python` does for requirements files. `all_requirements`: https://github.com/bazelbuild/rules_python/blob/master/docs/pip.md#pip_import

acmcarther:
Opened https://github.com/google/cargo-raze/discussions/327 for this

UebelAndre:
Use of Rust in my organization is extremely experimental. We have a few small projects that we're building out and am trying to understand the ecosystem with Bazel and measure all the pros and cons before adding it to our larger repositories. My months of work on `rules_rust` and `cargo-raze` have all been in an effort to address pain points that make people decide to simply not use Rust because our main build system is Bazel. We have a few mono-ish repos that would immediately make use of cargo workspaces. We want to be able to take advantage of the lockfiles and avoid having a single volatile/hand authored place for all dependencies. And we don't want to have to be spending all our time writing toml and BUILD files. In an ideal world, we would near seamlessly replace the use of Cargo with Bazel and be able to do all the same things without a bunch of additional hassle.

acmcarther:
In general, I think I'm open to approving this change, with the rationale of enabling a particular kind of more decentralized cargo-raze usecase. Can we discuss a bit in #327 and hammer out what this looks like?

UebelAndre:
I'm happy to discuss this but with the holidays basically here, I'm concerned I won't be able to complete the Bazel rollout for https://github.com/google/cargo-raze/issues/313 by the end of the month if you're going to be unavailable then. The changes I've made for the Bazel integration are built on this.

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+    },
+}
+
+def crates(deps):
+    """EXPERIMENTAL -- MAY CHANGE AT ANY TIME: Finds the fully qualified label of the requested crates for the package where this macro is called.
+
+    WARNING: This macro is part of an expeirmental API and is subject to change.
+
+    Args:
+        deps (list or str): Either a list of dependencies or a string of one which will
```
acmcarther, line 43: Can this be more specific? Is this a crate name? Does it contain a version?

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+    },
+}
+
+# EXPERIMENTAL -- MAY CHANGE AT ANY TIME: A mapping of package names to a set of proc_macro dev dependencies for the Rust targets of that package.
+_DEV_PROC_MACRO_DEPENDENCIES = {
+    "remote/binary_dependencies": {
+    },
+}
+
+def crates(deps):
```
acmcarther, line 37: Should this be called "dependencies"? "crates" communicates almost nothing about the semantics.

acmcarther, line 37: How is this intended to be used? I don't see any usages in this pr.

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther, line 43: Can this just take a list? I don't think the flexibility here makes this easier to use.

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+    dependencies = dict()
+    for dep_map in [_DEPENDENCIES, _PROC_MACRO_DEPENDENCIES, _DEV_DEPENDENCIES, _DEV_PROC_MACRO_DEPENDENCIES]:
+        for package_name in _DEPENDENCIES:
+            if package_name in dependencies:
+                dependencies[package_name].extend(dep_map[package_name])
+            else:
+                dependencies[package_name].update(dep_map[package_name])
+
+    if not deps:
+        fail("An invalid argument has been provided. Please pass a crate name or a list of crate names")
```
acmcarther, line 59: Does this need to be an error? I'd expect empty list to return empty list.

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+
+    if not dependencies:
```
acmcarther, line 61: I don't think this needs to be special cased. Below code should handle it.

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+        return []
+
+    if type(deps) == "string":
```
acmcarther, line 64: (per above, I don't think this flexibility is buying us anything.)

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+        deps = [deps]
+
+    errors = []
```
acmcarther, line 67: Nit: "missing_crates" communicates more than "errors"

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+    crates = []
```
acmcarther, line 68: Optional Nit: "crate_targets" or "crate_labels" is a little clearer.

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+    for crate in deps:
+        if crate not in dependencies[native.package_name()]:
+            errors.append(crate)
+        else:
+            crates.append(dependencies[native.package_name()][crate])
+
+    if errors:
+        fail("Missing crates `{}` for package `{}`. Available crates `{}".format(
```
acmcarther, line 76: Nit: I have a slight preference for tweaked wording here:

"Could not find crates {} among dependencies of {}. Available dependencies were {}"

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther, line 37: Can we take native.package_name() as an argument instead of inferring it from the call site?

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+        dev_only (bool, optional): If True, only development dependencies will be
+            returned by this list. This paramter otherwise follows the same rules
+            as `dev`. Defaults to False.
+
+    Returns:
+        list: A list of labels to cargo-raze generated targets (str)
+    """
+
+    # Determine the relevant maps to use
+    dependencies = dict()
```
acmcarther, line 109: Move closer to usage site.

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+    all_maps = []
```
acmcarther, line 110: I think this needs a better name. 

all_deps or all_dependencies?

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+    Setting any one flag will otherwise impact the contents of the returned list
+
+    WARNING: This macro is part of an expeirmental API and is subject to change.
+
+    Args:
+        normal (bool, optional): If True, normal dependencies are included in the
+            output list. Defaults to False.
+        proc_macro (bool, optional): If True, proc_macro dependencies will be
+            included in the output list. Defaults to False.
+        dev (bool, optional): If True, dev dependencies are included when the
```
acmcarther, line 98: The semantics of this argument is weird. Can this just return dev dependencies?

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+            all_maps.append(_DEV_DEPENDENCIES)
+
+    if not all_maps:
+        fail("Failed to add at least 1 map to the `all_maps` list with parameters: " +
+             "normal = {normal}, proc_macro = {proc_macro}, dev = {dev}, dev_only = {dev_only}".format(
+                 normal = normal,
+                 proc_macro = proc_macro,
+                 dev = dev,
+                 dev_only = dev_only,
+             ))
```
acmcarther, line 136: This map setup stuff is jank. I'm hoping that we can clean this up by updating the semantics of the `dev` argument.

examples/remote/binary_dependencies/cargo/crates.bzl
```diff
@@ -9,6 +9,144 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")  # bui
...
+        for package_name in dep_map:
+            if package_name in dependencies:
+                dependencies[package_name].extend(dep_map[package_name])
+            else:
+                dependencies[package_name] = dep_map[package_name]
+
+    if not dependencies:
+        return []
+
+    return dependencies[native.package_name()].values()
```
acmcarther, line 148: I feel less strongly about this one, but here too can we take package_name as an argument?

acmcarther (changes requested):

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: I kinda liked the vernacular of `crates` and I think contextually it's clear that the crates are dependencies. I can change it if you feel strongly but I would ask for more thoughts on this first.

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: It's a crate name. Are you asking why `deps` is either a list or a string? I thought it would be more convenient as I know I and others have run into situations where something expects a list but a string was given instead. 

The equivalent thing for this field is something like:
```python
load("//third_party/cargo:crates.bzl", "crates")
rust_library(
    name = "example",
    deps = [
        "//third_party/cargo:dependency1",
        crates("dependency2"),
    ],
    #...
)
```

The two dependencies are pulled from the same `cargo-raze` output, just using different methods of doing so.

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: I posted this on another comment but:
```python
load("//third_party/cargo:crates.bzl", "crates")
rust_library(
    name = "example1",
    deps = [
        "//third_party/cargo:dependency1",
    ] + crates("dependency2"),,
    #...
)

rust_library(
    name = "example2",
    deps = crates([
        "dependency1"
        "dependency2"
    ]),
    #...
)
```

`example1` and `example2` are the same crate, just with different names for the sake of the example

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: Yeah, would changing this also address https://github.com/google/cargo-raze/pull/319#discussion_r548356293 and https://github.com/google/cargo-raze/pull/319#discussion_r548356156 ?

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: What's the use case of passing an empty list? I think that's a better assumption but I was trying to be more defensive here and not allow unknown behavior before users decide they needed for less than ideal reasons.

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: This represents a grouping of all dependency maps, which are then iterated over. So it's not exactly `all_deps` but it does technically contain that data. Still change?

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: The use cases this needs to be able to handle is,
1. Grab all normal dependencies
2. Grab all proc_macro dependencies
3. Grap all normal dev dependencies
4. Grab all proc_macro dev dependencies
5. Grab dev and default normal dependencies
6. Grab dev and default proc_macro dependencies

What would clean this up best and solve for those? Or do you think that list is not quite accurate?

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: It can be an optional argument. But I don't want to force users to have to write that.

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: If you can operate with the input (even if the return value is not useful) it's best to just deal with it, rather than break.

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: No. I mean, as a reader, "list of dependencies" is insufficiently specific. Is it a crate identifier?

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: The crates aren't even actually dependencies. They're build targets. What this is doing is resolving a list of crate names to the build targets. 

What do you think about "crate_targets"

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: No. The weird type thing here is separate from "dependencies" being insufficiently specific.

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: yes, or alternative, a more specific and accurate name if you have one. "all_maps" isn't a good name.

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: Why not?

UebelAndre: This goes back to my desire to not have to write things unnecessarily. Why *would* we want to force users to always specify a package name? This to me is the purpose of `native.package_name()`. Macros can understand what package they're being invoked in. The only use case for providing an explicit package name is if you were trying to use dependencies from another package.

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: `all_dependency_maps`?

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: Even with this example, the semantics here are not clear. 

Pseudocode of something I'd find clearer

```
resolve_crate_dependences("crate-name", [
  "serde"
])
```

If you insist upon the magical version, that can be derived from the non-magical version

```
resolve_crate_dependencies_here([
  "serde"
])
```

where

```
def resolve_crate_dependencies_here(crate_names):
  return resolve_crate_dependencies(native.package_name(), crate_names)
```

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: I wrote some pseudo code below where I called this argument "crate_names". How do you feel about that option.

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: I think `crate_targets` is implied by `crates`. The macro is returning a list of targets already. It feels unnecessary to make the name more verbose. Is there another short one word name we can use?

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: I am anti magic, and this is magic.

I suggest one compromise above to break this into two macros: one "non magical" version, and one "magical" version. The magical version communciates that the output is dependent on the current package ("here").

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: I prefer verbosity especially when we're dealing with magical stuff, so I'm going to insist upon a verbose name. I suggest below the name "resolve_crate_dependences" and "resolve_crate_dependences_here" for the non-magic and the magic version

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: sgtm

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: How about

`normal, normal_dev, proc_macro, proc_macro_dev`?

It's one more argument, but I think the semantics are better.

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: That works for me. What should these macros be called then? Anything better than `*_here`? Not the worst, just wondering if there's anything better.

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: I think that's good, I'll update that

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: What do you think about 

`resolved_crate_deps(package_name = native.package_name(), crate_names)

(overload the same function). You get the magic for free, but the explicit version is also available. If mandatory args can't follow optional args, then 

`resolved_crate_deps(crate_names, package_name = native.package_name())` is an an option.

acmcarther: I became more open to this overload because I did a github wide search for native.package_name and realized what this is actually doing -- it's just returning the name of the build rule itself. That's less serious than what I thought it was, which was a fully qualified build target.

UebelAndre: Can we go with `crate_deps`? Can `resolved` be implied?

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: Updated

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: Updated

UebelAndre: "crate_names" is Ok. The names I chose were inspired by `rules_python'`s `requirements()` and `all_requirements`. I wanted `crates` and `all_crates()` to represent the same thing. Granted there is a bit more "magic" here that in `rules_python` but I don't think it's that much more (especially now that `package_name` is an argument). I generally prefer verbosity as well but I thought that the current names matched an existing and common pattern and would be more intuitive. What are your thoughts on the difference between this and what `rules_python` is doing?

UebelAndre:
@acmcarther Some updates and some questions for you.

UebelAndre:
@acmcarther Just a heads up, I gated this behind a new raze setting `experimental_api`. The new API introduced in this PR will only be available to users who opt in.

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: I prefer the verbose name, but not strongly enough to effort an argument or continue pushing back.

UebelAndre: I care a lot about this name being intuitive and not obnoxious but I value the correctness of it as well. If I can get your thoughts on https://github.com/google/cargo-raze/pull/319#discussion_r548377332 I can change it.

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: I better understand where this design came from having familiarized myself with rules_python.

I still prefer crate_names though. "deps" could either be expected to be a build target, a bare crate name, or a crate-crate_version.

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: Discussion on this is split into multiple comment sections, but I think we can both live with "crate_deps".

examples/remote/cargo_workspace/cargo/crates.bzl
acmcarther: Marking resolved (given the comments and the explicit opt-in)

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: Marking resolved per other discussion

UebelAndre: So `crate_deps` confirmed. Leaving now... `package_crate_deps`?

acmcarther: That sounds fine to me

UebelAndre: No, `all_crate_deps` I think is more correct since `package_name` is a parameter now.

UebelAndre: Ah dang, we both replied at the same time 😅

Your call. I'll make the change for `all_crate_deps` but can still change.

UebelAndre: Updated the names

examples/remote/binary_dependencies/cargo/crates.bzl
UebelAndre: Thanks for responding there 🙏

examples/remote/binary_dependencies/cargo/crates.bzl
acmcarther: iiuc, "all_crate_deps" isn't right for either the parameterized or unparameterized version. package_name is used to specify which package's dependencies are looked up, whereas the unparameterized version assumes that the current build target is a package in the resolved dependencies.

is that understanding incorrect?

UebelAndre: I don't think that's quite right, they both assume the current package unless the `package_name` is set to something otherwise. `crate_deps` is for looking up individual dependencies where `all_crate_deps` is for returning all the dependencies associated with `package_name` (again, defaulting to the current package).

acmcarther: Oh, my bad, I've gotten confused between the discussions.

`all_crate_deps` is fine.

acmcarther: Marking resolved.

acmcarther: Marking resolved.

acmcarther (commented):
I think the examples need to be regenerated (i see an instance of "all_crates".

else lgtm

UebelAndre:
> I think the examples need to be regenerated (i see an instance of "all_crates".
> 
> else lgtm

@acmcarther I thought I updated the examples already. I'm not able to find an outdated name.

acmcarther (approved):
Looks updated now.

Issue #310, opened 2020-12-02
Finish adding unit tests


UebelAndre:
There are a few `TODO`s scattered through the `tests` submodules as well some things to clarify/improve about the state of the current tests.

The following should be done to improve unit tests:
- [ ] Finish adding all tests (see `TODO`s like the following: https://github.com/google/cargo-raze/blob/master/impl/src/planning.rs#L806-L811)
- [x] Use smaller crates for testing: (see conversation here: https://github.com/google/cargo-raze/pull/300#discussion_r529642642)
- [ ] Double check that every test is still relevant (The code base has changed quite a bit since some of them were written)
- [x] Ensure the tests can be run without network connectivity

This list may grow but I feel these are the biggest action items to improve testing overall.

UebelAndre:
One thing I wanna clearly call out here. I'm not sure what `planning::tests::test_plan_build_produces_aliased_dependencies` is testing. This is relevant to the bullet

> Use smaller crates for testing

UebelAndre:
> One thing I wanna clearly call out here. I'm not sure what `planning::tests::test_plan_build_produces_aliased_dependencies` is testing. This is relevant to the bullet
> 
> > Use smaller crates for testing

I feel this is satisifed by https://github.com/google/cargo-raze/pull/346 which was a huge reduction to the amount of metadata used for testing.

PR #321, opened 2020-12-17
Removed `dummy_injecting_metadata` in favor of a templated metadata dump

UebelAndre:
This is a spinoff of https://github.com/google/cargo-raze/pull/319#discussion_r543835396 where I needed to update tests using the `dummy_injecting_metadata` and opted to replace that function with a templated dump of the metadata it's generating to cleanup how many ways metadata is created for tests. Ultimately, I think this metadata can be deleted since it's quite similar to one of the pacakges included in `dummy_workspace_members_metadata.json.template`. This PR should also be looked at when actioning #310

### DIFF
--- a/impl/src/checks.rs
+++ b/impl/src/checks.rs
@@ -241,7 +241,7 @@ mod tests {
     settings.genmode = GenMode::Vendored;
 
     let result = check_all_vendored(
-      &dummy_modified_metadata(),
+      &dummy_modified_metadata().metadata,
       &settings,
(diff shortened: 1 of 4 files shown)


UebelAndre:
@acmcarther If you have a moment, this would reduce the size of #319 and make it more focused. Again, this updates a helper function to conform to the (now) common way of generating metadata. I think the value here is it makes the code more consistent. Not trying to make a statement that the metadata dumps are how everything should be.

acmcarther:
LGTM

PR #324, opened 2020-12-18
`package_aliases_dir` now defaults to `cargo`

UebelAndre:
This is the safer option. As sated by the previous warning. Users can set this to `"."` to prevent any change in behavior from previous versions of `cargo-raze`.

### DIFF
--- a/impl/src/settings.rs
+++ b/impl/src/settings.rs
@@ -314,7 +314,7 @@ fn default_crate_settings_field_data_attr() -> Option<String> {
 }
 
 fn default_package_aliases_dir() -> String {
-  ".".to_owned()
+  "cargo".to_owned()
 }
 
 fn default_render_package_aliases() -> bool {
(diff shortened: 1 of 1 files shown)


acmcarther:
This should probably be bundled with the other pr.

UebelAndre:
I have another PR I'll open if https://github.com/google/cargo-raze/pull/319 is merged which I think will address your concern here. This change should be fine on it's own since ideally, users would have locked their `Cargo.toml` files down by the time the next release is out. Ultimately, I think this field will be deprecated since `cargo-raze` currently knows enough about what's being rendered to do something more automated like https://github.com/google/cargo-raze/pull/319. I do feel the change here is more correct that what's currently commited (hence the warning) but am happy to hold until the other PR is through.

acmcarther:
I still think this needs to wait longer than one release.

UebelAndre:
Any particular set of time you're thinking here?

acmcarther:
I don't have a good mental model for how often users are updating. The goal is for users to see the warning before the field changes. In general, I'm a slow adopter, so I assume folks will take a while (~ month or two) to see a warning and then make the change to be ready.

I don't feel strongly enough to push back if you insist upon making the change (as is with almost every single review discussion we have).

UebelAndre:
I'm fine with letting this sit for a month (two seems long). Again, the reason I opened a PR per "deprecated" setting is so each could be considered differently and rolled out with appropriate timing.

acmcarther:
It's been about a month and I'm ready to merge this. I think we need to regenerate examples though.

UebelAndre:
I've rebased and there's no change to examples

UebelAndre:
@acmcarther If this gets merged, can I get you to do a new release? 🙏

acmcarther:
I'll cut a release later today after the other PRs are submitted as well (setting up a calendar reminder for this).

PR #325, opened 2020-12-18
`default_gen_buildrs` now defaults to `true`

UebelAndre:
Closes https://github.com/bazelbuild/rules_rust/issues/490

In the common case, running `build.rs` is fine and doesn't break determinism or violate any sandboxing. For the cases that do, users can set `gen_buildrs = false` in the `CrateSetting` of the crate that has a complex `build.rs` file to avoid undesirable behavior. There are enough settings to work around most cases users would run into.

### DIFF
--- a/impl/src/settings.rs
+++ b/impl/src/settings.rs
@@ -334,7 +334,7 @@ fn default_raze_settings_field_output_buildfile_suffix() -> String {
 }
 
 fn default_raze_settings_field_gen_buildrs() -> bool {
-  false
+  true
 }
 
 fn default_raze_settings_registry() -> String {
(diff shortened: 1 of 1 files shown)


acmcarther:
bundle with the other pr?

UebelAndre:
Wait, why?

acmcarther:
No, I'm saying to remove all the deprecated behavior in one swoop. It's clear to me (though perhaps this wasn't the intent) that you want to update the defaults for the options introduced in the prior workspace PR. That sounds fine to me, but I think all of them should wait "a reasonable amount of time" before being updated. 

Else, what exactly was the point of warning users in the first place.

Reasonable amount of time here I think is longer than the regular definition, because we're in the holiday season where things tend to slow down at organizations.

UebelAndre:
The reasoning behind having a single PR for each notice/warning is to be able to selectively consider the impact of the change. If you feel the change should wait, I'm happy to have it wait, so long as there's a reasonable time when the changes should be merged. I feel a month is quite long for this change but I do understand where you're coming from. Do you still want me to group these changes?

UebelAndre:
Give that https://github.com/bazelbuild/rules_rust/issues/490 is over a month old. Could this be merged to be included in the next release?

acmcarther:
:shrug:

jfirebaugh:
README.md still implies that `default_gen_buildrs` defaults to `false`.

PR #336, opened 2020-12-29
Updated dependencies

UebelAndre:
The primary goal of this PR is to bring in the new version of `lalrpop` so we can make use of the fix introduced in https://github.com/lalrpop/lalrpop/pull/572

Otherwise, this brings in general improvements for a few dependencies. This should otherwise be no functional change for users.


UebelAndre:
This would mean I can remove the `lalrpop` directory in my Bazel branch for third part dependencies the README can be seen below:

# [lalrpop](https://github.com/lalrpop/lalrpop)

This dependency is only tracked because it requires a patch to work with Bazel.

## symlink_support.patch

This patch enables lalrpop to resolve symlinks which is required for it to be usable in Bazel. More details can be found on [lalrpop/lalrpop#572](https://github.com/lalrpop/lalrpop/pull/572). When this issue is resolved and is usable in [httpmock](https://github.com/alexliesenfeld/httpmock) (which is currently the only thing pulling in this dependency) then this patch should be removed.

UebelAndre:
Also, the `symlink_support.patch` was the diff you see in https://github.com/lalrpop/lalrpop/pull/572

UebelAndre:
@acmcarther Hey, just FIY, I force pushed here but the only change was updating the lockfile again. Figured I might as well since the PR was still here.

acmcarther:
Ah no worries. Sorry for the delay on the review. LGTM!

acmcarther (approved):

UebelAndre:
Thanks! 🙏

PR #337, opened 2020-12-30
Fixed a bug where export_files was being rendered into BUILD files it shouldn't have been

UebelAndre:
This would only impact projects using cargo workspace functionality since they're the the only projects that would have multiple workspace members. 

Aliases are rendered for each workspace member in the Bazel package where they're located, but only the root package will have a `crates.bzl` file. This PR limits where the `exports_files` macro is declared to the correct places.


acmcarther:
LGTM

Issue #335, opened 2020-12-29
Bazel workspace == Cargo workspace => invalid label (?)


cceckman:
Minimal (maybe) demo in https://github.com/cceckman/raze-bug.

*TL;DR* It seems that "Bazel workspace root" == "Cargo workspace root" doesn't work; `cargo raze` generates a label that begins `///`, which is not valid in Bazel.

Is this:
a) working as intended;
b) a bug;
c) not actually the case, and instead I've done something wrong? (Entirely possible! :-) )

---
This setup in https://github.com/cceckman/raze-bug is:

- `//` is a Bazel workspace and a Cargo workspace (and, incidentally, the Git root)
  - The workspace `Cargo.toml` has a `workspace.metadata.raze` section,
  - with `workspace_path = "//"`
- `//alib` is a Cargo crate
  - with a single external dependency (`num_enum`)
  - and `rust_library` and `rust_test` rules

The error:

```
∵ bazel test //alib:alib_test
ERROR: Traceback (most recent call last):
        File "/home/cceckman/r/github.com/cceckman/raze/WORKSPACE.bazel", line 22, column 25, in <toplevel>
                raze_fetch_remote_crates()
        File "/home/cceckman/r/github.com/cceckman/raze/crates.bzl", line 21, column 27, in raze_fetch_remote_crates
                build_file = Label("///remote:BUILD.derivative-2.1.1.bazel"),
Error in Label: Illegal absolute label syntax: ///remote:BUILD.derivative-2.1.1.bazel
ERROR: error loading package 'external': Package 'external' contains errors
INFO: Elapsed time: 0.114s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
∴ 1
```

(Same error from `bazel build //alib:alib`.)

Versions used in `VERSIONS` file there - it's `HEAD` of `rules_rust` and `0.8.0` of `cargo raze`.

If I'm reading right- and I may not be - this line comes from remote_crates.bzl.template; [here][v8] at the version of `cargo raze` used in my test, but still [here][head] at HEAD. The formatting there seems to assume that the relative path is not "the root of the repository".

The "smoke tests" don't seem to exercise this condition:

```
∵ pwd; ls
/home/cceckman/r/github.com/google/cargo-raze/smoke_test
README.md  remote  tests  vendored
∴ 0 cargo-raze:master…smoke_test
[...]

PR #338, opened 2021-01-03
Update comments on workspace_path

cceckman:
Per [issue 335], I didn't find these docstrings very helpful. The
important bit of this property appears to be "this is where outputs go".

Try to make that clearer in the property docstring and the README.

Also do a little cleanup:

- README:
  - Deduplicate the `Cargo.toml` contents between two headers. We only need to list the specifics once.
  - Use `package.metadata.raze` when used in a package-metadata context (rather than workspace-metadata)
  - Fix typos 
- Run `rustfmt` on `settings.rs` (my precommits complain otherwise).

[issue 335]: https://github.com/google/cargo-raze/issues/335

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,28 +78,10 @@ path = "fake_lib.rs"
 [dependencies]
 log = "=0.3.6"
 
-[workspace.metadata.raze]
-# The path relative path to the Bazel workspace root (location of
-# WORKSPACE.bazel/WORKSPACE file). If no workspace file is found,
-# the current working directory is used.
-workspace_path = "//cargo"
-
-# This causes aliases for dependencies to be rendered in the BUILD
-# file located next to this `Cargo.toml` file.
-package_aliases_dir = "."
-
-# The set of targets to generate BUILD rules for.
-targets = [
-    "x86_64-apple-darwin",
-    "x86_64-pc-windows-msvc",
-    "x86_64-unknown-linux-gnu",
-]
-
-# The two acceptable options are "Remote" and "Vendored" which
-# is used to idnicate whether the user is using a non-vendored or
-# vendored set of dependencies.
-genmode = "Remote"
 ```
+
+Once the standard Cargo.toml is in place, add the `[package.metadata.raze]`
+directives per the next section.
 
 ### Using existing Cargo.toml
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -78,29 +78,11 @@ path = "fake_lib.rs"
 [dependencies]
 log = "=0.3.6"
 
-[workspace.metadata.raze]
```
UebelAndre, line 81: While the documentation here is extremely similar, the difference is `[workspace.metadata` vs `[package.metadata`. I think it's worth adding a comment explaining that there are two settings

UebelAndre (approved):
One comment but otherwise this looks good to me. @acmcarther would be the person to merge this though so ultimately up to him

README.md
UebelAndre: but don't feel super strongly about it 😄 Overall still an improvement in my eyes

cceckman: This example is a package-level `Cargo.toml` (line 70 is `[package]`) rather than a workspace-level one, right? Do we use `workspace.metadata` even for a single-package workspace (i.e. when it's just a `[package]`)?

There's a later section on "Cargo workspace projects" that says "use `workspace.metadata.raze`"; I think that covers the workspace case.

UebelAndre: Good call. You **can** use workspace metadata for single packages. Cargo metadata always contains some notion of a workspace even if you only have a single root package. That said though, it'd be more consistent to use the package's metadata for single package crates (or at least that's what I'd expect to see as a reader). So this looks good to me 😄

acmcarther: @UebelAndre Do you think there's value in having these `[workspace]` instructions elsewhere?

acmcarther:
This LGTM. It's not clear to me whether or not there's value in preserving the `[workspace]` example though. I'm looking for @UebelAndre's input there.

acmcarther (approved):

README.md
UebelAndre: That should be covered by https://github.com/google/cargo-raze#cargo-workspace-projects (which I forgot existed). That's what @cceckman was mentioning earlier in this thread. 

Though, in reading the docs, I see the rendering is bad (my fault 😅 if there's bandwidth to fix that here it'd be greatly appreciated but I can otherwise open another PR)
Screen Shot 2021-01-05 at 7 15 48 PM

I think this is otherwise a good improvement.

UebelAndre:
@cceckman Hey, another PR that changed the documentation style caused a conflict with your PR here and it'll need a rebase. Should be a straight forward merge though 😄

cceckman:
Ack, thanks. I'll move this forward tonight / this weekend.

README.md
cceckman: Went ahead and fixed the rendering:

Screenshot 2021-01-10 at 10 47 15

cceckman:
Updated on top of existing formatting; should be good to go.

dfreese (approved):

PR #340, opened 2021-01-06
Restructured project to be buildable with Bazel

UebelAndre:
This adds Bazel support for the repository.

## Bootstrapping
In cases where the Bazel build doesn't work for whatever reason, a `./tools/bootstrap.sh` script can be run to build cargo-raze with cargo and generate the Bazel dependencies again

## Examples
The `smoke_tests` directory has been deleted and there is now only the `examples` directory. Examples are now managed by a repository rule (defined in `tools/examples.bzl` from the root level workspace). This rule will vendor the sources needed to compile the "vendored" examples but otherwise acts like a `local_repository`.

## Cargo compatibility
Because there's still a use case for this crate as a library on `crates.io` it's still important to make sure the cargo build works. This is handled by a `cargo` rule defined in `./tools/cargo.bzl` which is a simple wrapper around a cargo command. The rule is used as a `build_test` to ensure cargo builds continue to work.

## Publishing
The publishing script simply got moved. If maintainers are looking to publish new versions of the crate, they can do so via `bazel run //:publish_new_version`

## Third party
Some of the `-sys` dependencies need special libraries to compile. These have been added to the `third_pary` directory. Most of these are built using [rules_foreign_cc](https://github.com/bazelbuild/rules_foreign_cc) in an attempt to make pulling in new versions easier (than having to update a hand authored BUILD file).


UebelAndre:
I've opened as a draft PR for now to test CI and probably take some time to split up the commit to make it easier to review. Just wanted to get this opened though.

UebelAndre:
I think the travis build plan is correct but I need to make some changes for linux support. I've only tested on a Mac so far

UebelAndre:
@acmcarther FYI, this PR is blocked by https://github.com/google/cargo-raze/pull/343

impl/.rustfmt.toml
```diff
@@ -1 +0,0 @@
-tab_spaces = 2
```
acmcarther, line 1: We don't need this anymore?

impl/Cargo.toml
```diff
@@ -55,3 +55,87 @@ httpmock = "0.5.2"
 indoc = "1.0.3"
 lazy_static = "1.4.0"
 tar = "0.4.30"
+
+[package.metadata.raze]
```
acmcarther, line 59: This (raze-settings-in-raze) is really surreal for me. Great work!

acmcarther (approved):
This is amazing! Fantastic work!

Generally LG, though I definitely want to wait until those other CLs are merged.

impl/.rustfmt.toml
UebelAndre: There were two `rustfmt.toml` files, The one in the root of the repo seemed to have more in it so I kept that over this one.

UebelAndre:
@acmcarther Ok, this is ready for review now. I'd recommend looking at each commit since the generated BUILD files and removed `./smoke_test` directory are bloating the diff.

I'd say take a close look at `./third_party` and `./tools`. I made most of the design decisions for the tooling from the perspective of CI but might have a sorta clunky UX. Example being, the `examples_repository` macro which vendors crate source into the examples directory is really only useful when you're running bazel from the top level. If you're working in the `./examples` directory, you might see some behavior where the source code doesn't get vendored or updated when expected. If there's a way to improve that I'd love to learn about it. I otherwise will probably open PRs in the future as ideas come to me 😄

impl/.rustfmt.toml
acmcarther: SGTM

UebelAndre:
@acmcarther just a heads up, I'm gonna rebase this one more time just to be absolutely sure there are no regressions.

UebelAndre:
> @acmcarther just a heads up, I'm gonna rebase this one more time just to be absolutely sure there are no regressions.

Rebase complete. I will not be rebasing anymore until the review is complete.

UebelAndre:
It's also worth noting that without any caching, the bazel builds take longer than the current CI solution. When this is merged an new CI solution should probably be quickly added after the fact that has support for Bazel caching.

examples/.bazelversion
```diff
@@ -0,0 +1 @@
+3.7.1
```
acmcarther, line 1: Should this be `4.0.0` to match the other file?

acmcarther (approved):
LG.

I need to read up on the official way to set up buildkit.

examples/.bazelversion
UebelAndre: Ah yeah, good catch

UebelAndre: Updated

PR #341, opened 2021-01-06
Added `--generate-lockfile` for generating a new lockfile

UebelAndre:
This PR adds a new argument `--generate-lockfile` which will force a new lockfile to be generated.

## `CargoWorkspaceFiles` was deleted
`Cargo.toml` and `Cargo.lock` were basically never used for anything other than finding the cargo workspace. It's simpler to pass around a `Path` and allow the underlying cargo command to handle finding the correct metadata.

## `--generate-lockfile`
This will always generate a `Cargo.raze.lock` file which will take precedence over an existing `Cargo.lock` file. As a reminder, this file exists because to allow projects using `binary_deps` to have a lockfile that doesn't conflict with the standard lockfile. I'm hoping to one day delete the need for this file and standardize on the canonical `Cargo.lock` file, but for now, the use of `Cargo.raze.lock` makes the outputs of `cargo-raze` more consistent.

Running this command against the `examples/remote/cargo_workspace` example yields the following changes to the workspace
```command
cargo_workspace % git status
```
```output
On branch checksum
Your branch is up to date with 'origin/checksum'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   cargo/crates.bzl
	deleted:    cargo/remote/BUILD.addr2line-0.13.0.bazel
	modified:   cargo/remote/BUILD.atty-0.2.14.bazel
	deleted:    cargo/remote/BUILD.backtrace-0.3.53.bazel
	deleted:    cargo/remote/BUILD.cfg-if-0.1.10.bazel
	modified:   cargo/remote/BUILD.error-chain-0.10.0.bazel
	deleted:    cargo/remote/BUILD.getrandom-0.1.15.bazel
	deleted:    cargo/remote/BUILD.gimli-0.22.0.bazel
	modified:   cargo/remote/BUILD.hermit-abi-0.1.17.bazel
	deleted:    cargo/remote/BUILD.libc-0.2.79.bazel
	deleted:    cargo/remote/BUILD.object-0.21.1.bazel
	deleted:    cargo/remote/BUILD.ppv-lite86-0.2.9.bazel
	modified:   cargo/remote/BUILD.rand-0.7.3.bazel
	modified:   cargo/remote/BUILD.rand_chacha-0.2.2.bazel
	modified:   cargo/remote/BUILD.rand_core-0.5.1.bazel
	deleted:    cargo/remote/BUILD.rustc-demangle-0.1.17.bazel

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	cargo/Cargo.raze.lock
	cargo/remote/BUILD.addr2line-0.14.1.bazel
	cargo/remote/BUILD.backtrace-0.3.55.bazel
	cargo/remote/BUILD.getrandom-0.1.16.bazel
	cargo/remote/BUILD.gimli-0.23.0.bazel
	cargo/remote/BUILD.libc-0.2.81.bazel
	cargo/remote/BUILD.object-0.22.0.bazel
	cargo/remote/BUILD.ppv-lite86-0.2.10.bazel
	cargo/remote/BUILD.rustc-demangle-0.1.18.bazel

no changes added to commit (use "git add" and/or "git commit -a")
```

Note the addition of `cargo/Cargo.raze.lock` here. `cargo-raze` will use this for subsequent runs. It's also worth noting that if no `Cargo.lock` or `Cargo.raze.lock` files exist, a `Cargo.raze.lock` file will be generated.

## New warning for missing checksum
There's a new warning in the `checks` module that will print a warning for when a crates.io crate is missing a checksum:

With the following diff
```diff
diff --git a/examples/remote/cargo_workspace/Cargo.lock b/examples/remote/cargo_workspace/Cargo.lock
index 44494886..54cc7fa1 100644
--- a/examples/remote/cargo_workspace/Cargo.lock
+++ b/examples/remote/cargo_workspace/Cargo.lock
@@ -13,7 +13,7 @@ dependencies = [
 name = "adler"
 version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee2a4ec343196209d6594e19543ae87a39f96d5534d7174822a3ad825dd6ed7e"
+#checksum = "ee2a4ec343196209d6594e19543ae87a39f96d5534d7174822a3ad825dd6ed7e"

 [[package]]
 name = "ansi_term"
``` 

The following error is produced
```output
Running Cargo Raze for cargo_workspace
WARNING: Packages are missing checksums, perhaps `cargo generate-lockfile` needs to be run in the current cargo workspace. Missing checksums: ["adler-0.2.3"]
```

### DIFF
--- a/impl/src/checks.rs
+++ b/impl/src/checks.rs
@@ -326,7 +326,10 @@ mod tests {
 
     // Ensure all packages are workspace members for this test
     for package in raze_metadata.metadata.packages.iter() {
-      assert!(raze_metadata.metadata.workspace_members.contains(&package.id))
+      assert!(raze_metadata
+        .metadata
+        .workspace_members
+        .contains(&package.id))
     }
 
     // Ensure no checks fail
--- a/impl/src/checks.rs
+++ b/impl/src/checks.rs
@@ -20,10 +20,11 @@ use std::{
   path::PathBuf,
 };
 
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 
 use crate::{
   error::RazeError,
+  metadata::RazeMetadata,
   settings::{CrateSettingsPerVersion, GenMode, RazeSettings},
   util::collect_up_to,
   util::package_ident,
@@ -37,19 +38,67 @@ const MAX_DISPLAYED_MISSING_RESOLVE_PACKAGES: usize = 5;
 
 /// Ensure that the given Metadata is valid and ready to use for planning.
 pub fn check_metadata(
-  metadata: &Metadata,
+  raze_metadata: &RazeMetadata,
   settings: &RazeSettings,
   bazel_workspace_root: &Path,
 ) -> Result<()> {
   // Check for errors
-  check_resolve_matches_packages(metadata)?;
+  check_resolve_matches_packages(&raze_metadata.metadata)?;
 
   if settings.genmode == GenMode::Vendored {
-    check_all_vendored(metadata, settings, bazel_workspace_root)?;
+    check_all_vendored(&raze_metadata.metadata, settings, bazel_workspace_root)?;
+  }
+
+  // Check for incomplete lockfiles
+  if let Err(err) =
+    check_lockfile_for_missing_checksums(&raze_metadata.metadata, &raze_metadata.checksums)
+  {
+    eprintln!("WARNING: {}", err);
   }
 
   // Check for unused crate settings
-  warn_unused_settings(&settings.crates, &metadata.packages);
+  warn_unused_settings(&settings.crates, &raze_metadata.metadata.packages);
+
+  Ok(())
+}
+
+fn check_lockfile_for_missing_checksums(
+  metadata: &Metadata,
+  checksums: &HashMap<String, String>,
+) -> Result<()> {
+  let missing_checksums: Vec<String> = metadata
+    .packages
+    .iter()
+    // Filter out workspace members
+    .filter(|pkg| !metadata.workspace_members.contains(&pkg.id))
+    // Filter out non crates.io sources
+    .filter(|pkg| pkg.source.as_ref().map_or(false, |src| src.is_crates_io()))
+    // Filter for missing checksums and save the package identifier
+    .filter_map(|pkg| {
+      let package_ident = package_ident(&pkg.name, &pkg.version.to_string());
+      if !checksums.contains_key(&package_ident) {
+        Some(package_ident)
+      } else {
+        None
+      }
+    })
+    .collect();
+
+  if !missing_checksums.is_empty() {
+    let lockfile_path = metadata.workspace_root.join("Cargo.lock");
+    if !lockfile_path.exists() {
+      return Err(anyhow!(
+        "Packages are missing checksums, perhaps `cargo generate-lockfile` needs to be run in the \
+         current cargo workspace. Missing checksums: {:?}",
+        missing_checksums
+      ));
+    }
+    return Err(anyhow!(
+      "Packages are missing checksums, perhaps `cargo update` needs to be run in the current \
+       cargo workspace. Missing checksums: {:?}",
+      missing_checksums
+    ));
+  }
 
   Ok(())
 }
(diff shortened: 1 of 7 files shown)


UebelAndre:
@acmcarther Hey, do you have time to take a look at this today? Would hugely appreciate it 🙏 😅

impl/src/checks.rs
```diff
@@ -37,19 +38,67 @@ const MAX_DISPLAYED_MISSING_RESOLVE_PACKAGES: usize = 5;
...
   if settings.genmode == GenMode::Vendored {
-    check_all_vendored(metadata, settings, bazel_workspace_root)?;
+    check_all_vendored(&raze_metadata.metadata, settings, bazel_workspace_root)?;
+  }
+
+  // Check for incomplete lockfiles
+  if let Err(err) =
+    check_lockfile_for_missing_checksums(&raze_metadata.metadata, &raze_metadata.checksums)
+  {
+    eprintln!("WARNING: {}", err);
```
acmcarther, line 56: This is a non-fatal warning?

Comment in code why?

impl/src/metadata.rs
```diff
@@ -491,6 +485,26 @@ impl Default for RazeMetadataFetcher {
   }
 }
 
+/// Locates a lockfile for the associated crate. A `Cargo.raze.lock` file in the
+/// [RazeSettings::workspace_path](crate::settings::RazeSettings::workspace_path)
+/// direcotry will take precidence over a standard `Cargo.lock` file.
+pub fn find_lockfile(cargo_workspace_root: &Path, raze_output_dir: &Path) -> Option<PathBuf> {
```
acmcarther, line 491: I feel like this should be in the util package, not the metadata package, because it's not directly related to loading metadata.

acmcarther (approved):
LG, but I'm still not completely clear why just using Cargo.lock doesn't work. Can you explain that again?

You said above

> This will always generate a Cargo.raze.lock file which will take precedence over an existing Cargo.lock file. As a reminder, this file exists because to allow projects using binary_deps to have a lockfile that doesn't conflict with the standard lockfile. I'm hoping to one day delete the need for this file and standardize on the canonical Cargo.lock file, but for now, the use of Cargo.raze.lock makes the outputs of cargo-raze more consistent.

But I'm not clear why this "[allows] projects using binary_deps to have a lockfile that doesn't conflict with the standard lockfile". What happens if you just used `Cargo.lock`?

impl/src/checks.rs
UebelAndre: It wasn't originally fatal so I'm maintaining that behavior. We've just moved from silently writing repository definitions with missing `sha256` fields to informing users about it. Do you think it should be an error?

impl/src/metadata.rs
UebelAndre: Updated

UebelAndre:
> LG, but I'm still not completely clear why just using Cargo.lock doesn't work. Can you explain that again?

Currently, binary dependencies are being added to manifests during the `fetch` phase of `cargo-raze`
https://github.com/google/cargo-raze/blob/0cb6178cb5c5bde511779ef647a093fafc0a8d5d/impl/src/metadata.rs#L328-L364

This will cause the lockfile to contain content not reproducible via canonical Cargo commands. To avoid any conflicts with things like rust-analyzer that might update the `Cargo.lock` file under certain circumstances, `cargo-raze` will take a `Cargo.raze.lock` file over the canonical `Cargo.lock` file. This keeps "cargo workspaces" clean for projects that follow the 'many-tomls' model.

Users can use `Cargo.lock` files without any problems, but until `inject_binaries_into_workspace` can be removed, I think it's better to maintain a separate lockfile. I think with the progress in https://github.com/rust-lang/rfcs/pull/3028 we're getting closer to Cargo being able to handle binary dependencies first class and we won't need this tool to do anything unique anymore. 

However, Maybe I can be smarter about detecting which lockfile is relevant, the custom `Cargo.raze.lock` file or the canonical `Cargo.lock` file based on the existence of `binary_deps` in the current `RazeSettings`? Your thoughts?

impl/src/checks.rs
acmcarther: Hmm. I don't feel strongly either way. I just thought this looked weird. 

Current state sgtm.

acmcarther:
Given what you've said above, I prefer always using our weird cargo.raze.lock, rather than conditionally using it when we need to.

PR #343, opened 2021-01-11
Added settings for adding additional content to `cargo_build_script` targets

UebelAndre:
This PR adds some new CrateSettings
1. `data_dependencies` - A list of targets for the `data` attribute of a Rust target
2. `build_data_dependencies` - The `data` attribute for buildrs targets
3. `buildrs_additional_deps` - Additional dependencies for buildrs targets. See `additional_deps`

Also notably, this PR removes the `data_dependencies` and `build_data_dependencies` fields from `CrateDependencyContext` (added in https://github.com/google/cargo-raze/pull/311)so that they may be used by cargo raze (since they were previously only usable by `cargo-raze` as a library). These changes are used to get builds in https://github.com/google/cargo-raze/pull/340 to successfully build.

### DIFF
--- a/impl/src/context.rs
+++ b/impl/src/context.rs
@@ -85,8 +85,12 @@ pub struct SourceDetails {
 pub struct CrateDependencyContext {
   pub dependencies: Vec<BuildableDependency>,
   pub proc_macro_dependencies: Vec<BuildableDependency>,
+  // data_dependencies can only be set when using cargo-raze as a library at the moment.
+  pub data_dependencies: Vec<BuildableDependency>,
   pub build_dependencies: Vec<BuildableDependency>,
   pub build_proc_macro_dependencies: Vec<BuildableDependency>,
+  // build_data_dependencies can only be set when using cargo-raze as a library at the moment.
+  pub build_data_dependencies: Vec<BuildableDependency>,
   pub dev_dependencies: Vec<BuildableDependency>,
   pub aliased_dependencies: Vec<DependencyAlias>,
 }
--- a/impl/src/context.rs
+++ b/impl/src/context.rs
@@ -85,12 +85,8 @@ pub struct SourceDetails {
 pub struct CrateDependencyContext {
   pub dependencies: Vec<BuildableDependency>,
   pub proc_macro_dependencies: Vec<BuildableDependency>,
-  // data_dependencies can only be set when using cargo-raze as a library at the moment.
-  pub data_dependencies: Vec<BuildableDependency>,
   pub build_dependencies: Vec<BuildableDependency>,
   pub build_proc_macro_dependencies: Vec<BuildableDependency>,
-  // build_data_dependencies can only be set when using cargo-raze as a library at the moment.
-  pub build_data_dependencies: Vec<BuildableDependency>,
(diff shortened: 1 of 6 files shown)


UebelAndre:
@sitaktif Hey, I'd love to pick your brain on this. Can you explain why you added `data_dependencies` and `build_data_dependencies` specifically to `CrateDependencyContext`? I wanted to hook these up to use them in another PR I'm working on but felt they were too limiting on that struct and there wasn't a great way to access them. I moved them to `CrateSettings` so users can set these from a `Cargo.toml` but I'm not entirely sure if it's correct to replace what you added. Can you explain the use case or rationale a bit more behind the original change?

sitaktif:
Hey @UebelAndre! So the main issue here from an API usage perspective is related to targeted deps: it is sometimes useful to have data deps depend on the target triple (and it's currently how we use it). If you look at the `CrateTargetedDepContext` class:

```rust
pub struct CrateTargetedDepContext {
  pub target: String,
  pub deps: CrateDependencyContext,
  pub conditions: Vec<String>,
}
```

That's the main reason behind putting these attributes in `CrateDependencyContext`.

Do you think we could instead add the attributes to `CrateSettings` (as you did) but also keep them in `CrateDependencyContext` and have them added to the `default_deps` CrateDependencyContext of the CrateContext for the appropriate crate? Then we can use the latter in the template as it is currently.

Doing that will allow cargo-raze to handle targeted data deps once there is user-side support for it in the Cargo.toml raze metadata. WDYT?

UebelAndre:
@sitaktif  re: https://github.com/google/cargo-raze/pull/343#issuecomment-758110065
I think that sounds good. My biggest concern was whether or not I was missing some context that might make this change unnecessary but it seems like it's a good one with the exception of removing the fields from `CrateDependencyContext`. Do you concur?

Also Is there an RFC for the Cargo side of this that describes targeted data dependencies?

UebelAndre:
@sitaktif How's it look now?

sitaktif:
I don't have too strong an opinion about it but I feel it might be better to add the CrateSettings data attributes to the default_deps of the crates outside of the template itself.

Other than that, this is the idea!

UebelAndre:
I've been thinking about the same thing. I feel like there shouldn't be a need to add `raze_settings` to a `CrateContext` and the "planning" step should instead extract the necessary data. But I think this is in alignment with how things are currently done. That can be an optimization made in the future.

UebelAndre:
@acmcarther This is ready for review now 😄

acmcarther (approved):
This LGTM.

Can we come up with any examples which exercise this behavior? I don't feel strongly about this, so if absolutely none come to mind, then I'm happy to just merge this as is. I'm just a bit nervous about merging new un-exercised Tera templates.

UebelAndre:
I was lazy in not adding examples because the functionality here is demonstrated in https://github.com/google/cargo-raze/pull/340 I can take some time to add an example but that just means the Bazel migration will take longer for https://github.com/google/cargo-raze/issues/313 just FYI (not sure what the state of the time pressure is right now).

acmcarther:
No I think that's OK, we can wait until the Bazel build CL is submitted.

I'm not sure what the travis CI situation is right now.

PR #347, opened 2021-01-19
Added support for a `CARGO` environment variable

UebelAndre:
Similar to [cargo_metadata::MetadataCommand::cargo_path](https://docs.rs/cargo_metadata/0.12.3/cargo_metadata/struct.MetadataCommand.html#method.cargo_path) (which is the underlying crate used for the primary cargo commands in `cargo-raze`), Cargo raze will now accept an optional `CARGO` environment variable for specifying the path to the Cargo binary. This makes it realistic to use `cargo-raze` in Bazel rules since `cargo-raze` requires a Cargo binary.

### DIFF
--- a/impl/src/metadata.rs
+++ b/impl/src/metadata.rs
@@ -14,7 +14,7 @@
 
 use std::{
   collections::HashMap,
-  fs,
+  env, fs,
   path::{Path, PathBuf},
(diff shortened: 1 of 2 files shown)


acmcarther (approved):
LGTM

PR #352, opened 2021-01-25
Add initial bazelci/presubmit.yml

acmcarther:
This PR adds an initial .bazelci/presubmit.yml per onboarding instructions of github.com/bazelbuild/continuous-integration.

We won't be able to sanity check this because we haven't actually been onboarded yet (although a CI run _might_ trigger because I created a pipeline already).


acmcarther:
CI is super bork. buildkite seems to have seen the webhook OK but can't clone. Travis is dead.

Yolosubmit it is.

Issue #355, opened 2021-02-02
v0.9.0 no longer emits dev dependency targets


wchargin:
In `cargo-raze` v0.8.0 and prior, the user-facing `BUILD` package
included aliases for dev dependencies, just like normal dependencies. As
of v0.9.0, these seem to be removed entirely. This breaks tests and dev
tools.

Repro repo:
<https://github.com/wchargin/tensorboard/tree/wchargin-raze-v0.8.0>
in the `tensorboard/data/server/` directory:

Note first that at v0.8.0, the project builds fine:

```
$ cd tensorboard/data/server/
$ ~/tmp/cargo-raze-v0.8.0
$ git status
HEAD detached at wc/wchargin-raze-v0.8.0
nothing to commit, working tree clean
$ bazel test ...
INFO: Analyzed 13 targets (4 packages loaded, 50 targets configured).
INFO: Found 10 targets and 3 test targets...
INFO: Elapsed time: 1.506s, Critical Path: 0.04s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
//tensorboard/data/server:rustboard_core_doc_test               (cached) PASSED in 3.5s
//tensorboard/data/server:rustboard_core_test                   (cached) PASSED in 0.5s
//tensorboard/data/server:update_protos_test                    (cached) PASSED in 0.1s

Executed 0 out of 3 tests: 3 tests pass.
INFO: Build completed successfully, 1 total action
```

But when re-generating at v0.9.0, the configs are broken:

```
$ ~/tmp/cargo-raze-v0.9.0
$ git diff -U0
diff --git a/tensorboard/data/server/cargo/BUILD.bazel b/tensorboard/data/server/cargo/BUILD.bazel
index 1877cee4c..968f90e0f 100644
--- a/tensorboard/data/server/cargo/BUILD.bazel
+++ b/tensorboard/data/server/cargo/BUILD.bazel
@@ -114,9 +113,0 @@ alias(
@@ -168,9 +158,0 @@ alias(
@@ -214,2 +196,2 @@ alias(
@@ -222,6 +204,4 @@ alias(
@@ -228,0 +209 @@ alias(
-alias(
-    name = "prost_build",
-    actual = "@raze__prost_build__0_7_0//:prost_build",
-    tags = [
-        "cargo-raze",
-        "manual",
-    ],
-)
-
-alias(
-    name = "tempfile",
-    actual = "@raze__tempfile__3_1_0//:tempfile",
[...]

Issue #339, opened 2021-01-05
Feature Request: Argument for setting up a Bazel workspace on top of a Cargo workspacee


UebelAndre:
I'd like a `--init` argument (name WIP) that I can pass to `cargo-raze` and have it not just output build files, but generate a Bazel workspace in place on top of a Cargo workspace with all the appropriate dependencies plugged in. This would help tremendously with migrating various repos over to using Bazel.

It's worth noting that most, if not all, repositories are for binaries and will not be used for sharing dependencies. They're expected to be built, more or less, in isolation and used as tools for other rules or data for other targets. But even then, I would find it incredibly useful and in alignment with other Rust tools to have some kind of `--init` flag ([cargo init](https://doc.rust-lang.org/cargo/commands/cargo-init.html), [wasm-pack new](https://rustwasm.github.io/docs/wasm-pack/commands/new.html)).

UebelAndre:
My expectation is that by implementing this, the need for `` and `` will be removed. 
https://github.com/google/cargo-raze/blob/3b0a7cbc7ecc048c41ab6644f6108de650e3d1e7/impl/src/planning/subplanners.rs#L389-L390

These fields are used for the experimental API to render the dependency maps into `crates.bzl`. If a `PlannedBuild` can be made aware of the workspace members, then we can iterrate over those when rendering instead of iterating over each crate to find out if it's a dependency of the current package.
https://github.com/google/cargo-raze/blob/3b0a7cbc7ecc048c41ab6644f6108de650e3d1e7/impl/src/rendering/templates/partials/crates_macro.template#L3-L17

UebelAndre:
@GregBowyer Just wanted to share https://github.com/google/cargo-raze/issues/339#issuecomment-760405817 with you in case you find the existence of `workspace_member_dependents` and `workspace_member_dev_dependents` odd.

PR #358, opened 2021-02-03
Fixed dev dependency aliases

UebelAndre:
This fixes a regression brought up in #355


UebelAndre:
@dfreese @acmcarther do you have a moment to review this and potentially do a patch release?

dfreese (approved):

dfreese:
To the point brought up in #355, having the examples technically catches this, but it would be preferable if this was covered by a unit test some how.

UebelAndre:
> To the point brought up in #355, having the examples technically catches this, but it would be preferable if this was covered by a unit test some how.

Agreed. I've been meaning to do some slight refactoring as part of https://github.com/google/cargo-raze/issues/339 which I believe will require similar testing. When I have some free time I'll try to do this.

UebelAndre:
@acmcarther @dfreese Can either of you do a release to actually resolve #355 ? 🙏

PR #359, opened 2021-02-03
Add github actions for cargo-based builds

dfreese:
I don't have a lot of familiarity with bazelci, so I'm not sure how much I can help on that front.  My team uses github actions so I'm a little more familiar with that.

This PR puts together a basic version of a cargo-only version of the CI.  It's based off of the [MRSV example](https://github.com/actions-rs/meta/blob/master/recipes/msrv.md) from the actions-rs project.  I might have a chance to look at bazel-based builds later on, but perhaps this will unblock some things.

Perhaps there was something I didn't completely understand on the rustfmt options, but it seems like there were some formatting regressions that I fixed in a separate commit in the stack.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ jobs:
       matrix:
         rust:
           - stable
-          - 1.46.0
+          - 1.45.0
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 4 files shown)


dfreese:
cc: @UebelAndre 

I have been testing this at dfreese/cargo-raze#1

UebelAndre:
I think this is a good thing to have if BuildKite has no way to test Cargo. I was trying to write a Bazel rule for running Cargo but haven't gotten it working. I would really appreciate some help 😅 

https://github.com/UebelAndre/cargo-raze/tree/cargo

acmcarther:
I really don't have the cycles this week to get build kite working, so this is great.

PR #365, opened 2021-02-10
Fixed windows bug in examples repository rule

UebelAndre:
This fixes vendoring examples source on windows machines. I found out that just because `bash` exists on a Windows host (namely on a BuildKite Windows worker), doesn't mean the programs are the same. The issue I ran into was `FIND: Parameter format not correct` and this is because the `find` program on a unix host is not the same on windows. Thus the bash script would not have worked as is. I decided to write my first ever batch script.

### DIFF
--- a/tools/BUILD.bazel
+++ b/tools/BUILD.bazel
@@ -2,6 +2,7 @@ package(default_visibility = ["//:__subpackages__"])
 
 exports_files([
     "examples_repository_tools.sh",
+    "examples_repository_tools.bat",
 ])
 
 sh_binary(
--- a/tools/examples_repository.bzl
+++ b/tools/examples_repository.bzl
@@ -12,8 +12,8 @@ def _examples_dir(repository_ctx):
     Returns:
         path: A path to the cargo-raze workspace root
     """
-    workspace_root = repository_ctx.path(repository_ctx.attr._script).dirname.dirname
-    return repository_ctx.path(str(workspace_root) + "/" + repository_ctx.attr._examples_dir.lstrip("/"))
+    examples_cargo_raze_root = repository_ctx.path(repository_ctx.attr._script).dirname.dirname
+    return repository_ctx.path(str(examples_cargo_raze_root) + "/examples")
 
 EXECUTE_FAIL_MESSAGE = """\
 Failed to setup examples repository with exit code ({}).
@@ -31,7 +31,6 @@ def _examples_repository_impl(repository_ctx):
     """
 
     examples_dir = _examples_dir(repository_ctx)
-    vendor_script = repository_ctx.attr._script
 
     if repository_ctx.attr.target_triple:
         target_triple = repository_ctx.attr.target_triple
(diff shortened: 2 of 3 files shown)


UebelAndre:
@dfreese This is a good one for getting windows closer to working

UebelAndre:
@dfreese Can you take a look at this one as well?

dfreese:
I don't believe that I have enough windows expertise to review this.

UebelAndre:
@dfreese I don't feel I have enough windows experience to have written this 😅. I have confirmed in https://github.com/bazelbuild/rules_rust/pull/578 that the changes here correctly vendor the crates on windows. There may be a better way to write this code but I think for the time being it's worth merging since it addresses some broken functionality.

UebelAndre:
@dfreese One more comment on this. I don't feel there's a risk in changing something that already doesn't work. I *have* confirmed in that closed PR that these changes work and means I can move on to updating the examples to being windows compatible (which they currently are not).

UebelAndre:
Additionally, this PR helped me make https://github.com/google/cargo-raze/pull/370 

There's more to sort out but I'd prefer to merge them individually since they're still have an impact on the other platforms.

UebelAndre:
@dfreese Also, I'm happy to answer any questions I can, but I'm not windows expert. This was pretty tough to get working

UebelAndre:
@dfreese I'm not sure anyone else has a large amount of windows experience either. The batch script is not particularly complex, could I please get you to take a look here again 🙏

UebelAndre:
@acmcarther I think if this PR gets merged, then there should be more understandable issues with BazelCI. Though, I think BazelCI should be supported currently so long as you disable Windows. If you can somehow enable BazelCI for PRs, I can solve the rest. But without this I'm stuck having coupled PRs or super slow testing in a windows VM 😅

If you could take a look here I'd greatly appreciate it.

acmcarther:
We can ascertain whether or not this fixes the issue in windows by syncing this PR to current master, then un-commenting out the .bazelci/presubmit.yml change that disabled windows.

For now it will be pretty hard to tell what the root cause of the failure is. I'm working on getting the pipeline made public visible though.

UebelAndre:
It doesn't fix all issues but does fix issues in the repository rule. I don't have the bandwidth to fix all windows issues at once so I'd rather get in why I can

acmcarther:
Let me see if I can trigger CI and verify that this fixes part of the issue.

acmcarther:
Woah, I did it: https://buildkite.com/bazel/cargo-raze/builds/61

acmcarther:
Oh, nope, this won't run the windows stuff.

acmcarther:
This failure is partially a red herring. It's using your currently synced presubmit, which is broken for OSX. Sync should make this green (and also prevent windows builds).

The logs from the windows build: https://hastebin.cheapassbox.com/?054dbce6b96aec07#BHpGfDUy5JhCQvpW17JSphhq2kMPY7hESesE5BnqMz2G

It looks better than before. Breakage is in building `winapi`, which I think requires a build.rs

PR #366, opened 2021-02-14
Added a helper function for locating the cargo binary

UebelAndre:
This is a slight cleanup but means we don't have to add the same assumption on the `CARGO` environment variable everywhere.


UebelAndre:
Actually, gonna solve this another way.

PR #368, opened 2021-02-19
Updated buildifier annotations 

UebelAndre:
This makes it so users running `buildifier -lint=fix -mode=fix -warnings=all -r ./` don't generate diffs in the generated build files. The output build files should now stay the same no matter what buildifier flags are used.

### DIFF
--- a/impl/src/rendering/templates/crate.BUILD.template
+++ b/impl/src/rendering/templates/crate.BUILD.template
@@ -5,7 +5,9 @@ cargo-raze crate build file.
 DO NOT EDIT! Replaced on runs of cargo-raze
 """
 
-# buildifier: disable=out-of-order-load
+# buildifier: disable=load
+load("@bazel_skylib//lib:selects.bzl", "selects")
+
 # buildifier: disable=load
 load(
     "@{{ rust_rules_workspace_name }}//rust:rust.bzl",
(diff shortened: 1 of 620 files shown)


UebelAndre:
@dfreese This should be a quick one if you could take a look 🙏

impl/src/rendering/templates/crate.BUILD.template
```diff
@@ -5,6 +5,7 @@ cargo-raze crate build file.
 DO NOT EDIT! Replaced on runs of cargo-raze
 """
 
+# buildifier: disable=out-of-order-load
```
dfreese, line 8: for both `io_bazel_rules_rust` and `rules_rust`, it looks like this could be solved by moving the `bazel_skylib` load before `rules_rust`.

Is there a reason to go this way, rather than just rearranging them?

UebelAndre: Good call! Sorry about that, I should have spent more time on this. I've had many headaches solving buildifier defects in these templates so was just looking for the quickest solution 😅

UebelAndre: Just pushed some updates

dfreese (approved):

PR #369, opened 2021-02-20
Set `CARGO` and `RUSTC` environment variables to make Bazel tests more hermetic

UebelAndre:
Using the changes introduced in https://github.com/bazelbuild/rules_rust/pull/588, cargo raze can now correctly specify the `cargo` and `rustc` binaries to use during tests and no longer require the host to have either installed. For users running the target, however, the binaries are still required to be installed but this should be a simple change that can be made in the future

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,11 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "rules_rust",
-    sha256 = "42e60f81e2b269d28334b73b70d02fb516c8de0c16242f5d376bfe6d94a3509f",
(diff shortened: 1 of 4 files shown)


UebelAndre:
@dfreese This one solves for one of issues @acmcarther ran into in his quest for setting up BazelCI. This contains the most impactful fixes of getting that working.

dfreese:
This looks like it deals with three separate issues.  Can you break those out into individual PRs?

UebelAndre:
> This looks like it deals with three separate issues. Can you break those out into individual PRs?

This has been split into https://github.com/google/cargo-raze/pull/372 and https://github.com/google/cargo-raze/pull/371

dfreese (approved):

PR #371, opened 2021-02-20
Fixed missing "CARGO" environment variable uses

UebelAndre:
A bug was fixed where the `"CARGO"` environment variable wasn't being when looking for a cargo binary. This has been corrected and is now in use.

### DIFF
--- a/impl/src/metadata.rs
+++ b/impl/src/metadata.rs
@@ -14,7 +14,7 @@
 
 use std::{
   collections::HashMap,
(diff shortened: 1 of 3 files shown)


dfreese (approved):

dfreese:
Thanks for splitting that out.

acmcarther:
Thanks for fixing this!

(i independently found this in my experimentation, before realizing that you already had this fixed and checked in)

PR #372, opened 2021-02-20
Prefixed 3rd party dependencies with `cargo_raze__` to match existing…

UebelAndre:
The 3rd party dependencies had their names updated to prevent any kinds of conflicts with other repository names. This should further improve the reproducibility of cargo-raze builds and tests.


third_party/iconv/BUILD.iconv.bazel
```diff
@@ -10,7 +10,7 @@ filegroup(
 configure_make(
     name = "iconv",
     configure_env_vars = select({
-        "@platforms//os:macos": {"AR": ""},
+        "@rules_rust//rust/platform:darwin": {"AR": ""},
```
dfreese, line 13: Is this change auto-generated, or manual?  If manual, why go with `rules_rust`'s version, rather than just `platforms`?

dfreese (approved):
generally looks good to me.  Minor question though.

third_party/iconv/BUILD.iconv.bazel
UebelAndre: Use of `@platform` here isn't supported in versions before... 3.6 if I recall correctly. Using the rust config settings guarantees the same level of compatibility the rules offers.

UebelAndre: Specifically because `@platforms` defines constraint_values and not config_settings.

dfreese: Looks like 3.7
>select() directly supports constraint_value with no need for an intermediate config_setting.

https://blog.bazel.build/2020/10/20/bazel-3-7.html

dfreese: Opened #373

PR #375, opened 2021-02-20
Ensure examples are up to date in CI

UebelAndre:
`//tools:examples_check` will run `cargo-raze` (built from Bazel) on all the examples directory and then check for any diffs to ensure users don't forget to update them when changes are to the library.


acmcarther:
Can you sync? That will do double duty both to unblock merging, and also to trigger CI so we can verify that this works.

acmcarther:
Apparently I have the power to sync and I have done so. Hopefully me doing that didn't cause a problem.

acmcarther:
👍

acmcarther:
Reverting this. I haven't had a chance to check why this is broken. It seems like it might be some kind of infinite loop.

UebelAndre:
@acmcarther I can take a look at this. If you can get CI builds publicly available, I'll fix my own bug 😅

acmcarther:
The best I can do right now is dump the failing logs into my hastebin:

I know you can't iterate with that, so feel free to punt on this until permissions are fixed. 

Just kidding. Log file is 160 MB. Nope.

UebelAndre:
Once CI is made public I can take a look.

Issue #377, opened 2021-02-21
raze panics when Cargo.toml doesn't contain raze metadata


gkelly:
Pretty easy to replicate. `cargo new` somewhere and then `cargo raze` it:

```
gkelly@praxis> cargo new --bin razor
     Created binary (application) `razor` package
gkelly@praxis> cd razor
gkelly@praxis> cargo raze
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', src/settings.rs:640:24
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

gkelly:
With #378 merged this now looks like:

```
gkelly@praxis> cargo raze
Error: Raze failed with cause: "No raze settings were specified in the Cargo.toml file, see README.md for details on expected fields"
```

PR #378, opened 2021-02-21
Fix panic when invoked on crate missing metadata (#377)

gkelly:
Also add a test that ensures the error is reported and a panic doesn't
occur.

### DIFF
--- a/impl/src/settings.rs
+++ b/impl/src/settings.rs
@@ -17,7 +17,7 @@ use crate::{
   metadata::{MetadataFetcher, DEFAULT_CRATE_INDEX_URL, DEFAULT_CRATE_REGISTRY_URL},
   util,
 };
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use cargo_metadata::{Metadata, MetadataCommand, Package};
 use semver::VersionReq;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


gkelly:
With this patch instead of panicking the result is now:

```
Error: Raze failed with cause: "Exactly one package must contain primary raze settings: []"
```

impl/src/settings.rs
```diff
@@ -626,9 +626,9 @@ fn parse_raze_settings_any_package(metadata: &Metadata) -> Result<RazeSettings>
   }
 
   // There should only be one package with raze
-  if settings_packages.len() > 1 {
+  if settings_packages.len() != 1 {
```
dfreese, line 629: It's probably worth specifying the zero case separately.  Perhaps something like:
>No raze settings were specified in the Cargo.toml file, see README.md for details on expected fields

would be more helpful?

also, not your doing, but I'm noticing that
```rust
return Err(anyhow!("error message"));
```
is used.  Anyhow provides anyhow::bail
```rust
bail!("error message");
```
that does the same thing, more concisely.

dfreese (commented):
Thanks for the help providing clearer errors.

impl/src/settings.rs
gkelly: > It's probably worth specifying the zero case separately. Perhaps something like:
> 
> > No raze settings were specified in the Cargo.toml file, see README.md for details on expected fields
> 
> would be more helpful?

Done. I do worry that it should perhaps be a URL, as if someone is just running cargo-raze as part of CI, etc, that it won't be clear which `README.md` is intended in context.

> also, not your doing, but I'm noticing that
> 
> ```rust
> return Err(anyhow!("error message"));
> ```
> 
> is used. Anyhow provides anyhow::bail
> 
> ```rust
> bail!("error message");
> ```
> 
> that does the same thing, more concisely.

Done.

dfreese:
great, thanks!

gkelly:
> great, thanks!

I appreciate the reviews and merge, thanks!

Issue #354, opened 2021-02-02
Package names start with a leading slash


nobu-k:
I encountered an issue that package names in `crates.bzl` start with a leading slash when:

1. I use workspace
2. The root `Cargo.toml` that contains `[workspace]` is in the same directory as `WORKSPACE` (i.e. in the Bazel's root directory)

As a result, experimental APIs including `crate_deps` fails with the default `package_name` (i.e. `None`) to locate dependencies because the package name in `crates.bzl` and the one returned from `native.package_name()` are different.

I've set up a repository to reproduce this issue: https://github.com/nobu-k/cargo-raze-debug

With this layout, package names look like as follows:

```
# EXPERIMENTAL -- MAY CHANGE AT ANY TIME: A mapping of package names to a set of normal dependencies for the Rust targets of that package.
_DEPENDENCIES = {
    "/test-lib": {
        "rand": "@raze__rand__0_8_3//:rand",
    },
    "/test-bin": {
    },
    "/nested/lib": {
    },
}
```

Package names in `crates.bzl` shouldn't have leading slashes to be the same as the return value of `native.package_name()`.

If I put the root `Cargo.toml` and other Rust packages under a subdirectory, the leading slash disappears. I created a branch to demonstrate it. You can see generated `crates.bzl` [here](https://github.com/nobu-k/cargo-raze-debug/blob/nested/cargo/crates.bzl). Package names in this branch don't start with a `/`.

I confirmed this behavior with Bazel 4.0.0 and 3.7.2.

There's fortunately an obvious workaround that I can explicitly pass a package name with a leading slash to the functions like `crate_deps([...], "/test-lib")`, so this issue isn't critical.

UebelAndre:
Thanks for the heads up! I'll take a look at this once CI is fixed 😄

PR #385, opened 2021-02-24
Fixed leading slashes in experimental API dependencies maps

UebelAndre:
In the way the dependency maps were being rendered, it always assumed the bazel package containing the cargo workspace is not an empty file. This is the case for the `cargo_workspace` examples in this repo but is an incorrect assumption. This PR fixes that.

Closes https://github.com/google/cargo-raze/issues/354

### DIFF
--- a/impl/src/rendering/bazel.rs
+++ b/impl/src/rendering/bazel.rs
@@ -882,4 +882,67 @@ mod tests {
             "# }
     }))
   }
+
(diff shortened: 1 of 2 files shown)


acmcarther (approved):
LGTM

UebelAndre:
@acmcarther A patch release for this would be super helpful! 🙏

PR #386, opened 2021-02-24
Add clippy check in Bazel

UebelAndre:
Maybe it can be disabled from the Github Actions? I'd prefer to exercise the Bazel rules wherever possible.

### DIFF
--- a/impl/BUILD.bazel
+++ b/impl/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@rules_rust//rust:rust.bzl", "rust_binary", "rust_library", "rust_test")
+load("@rules_rust//rust:rust.bzl", "rust_binary", "rust_clippy", "rust_library", "rust_test")
 load("//third_party/cargo:crates.bzl", "all_crate_deps")
 
 package(default_visibility = ["//visibility:public"])
@@ -81,4 +81,13 @@ rust_test(
     env = _TEST_ENV,
     proc_macro_deps = all_crate_deps(proc_macro_dev = True),
     deps = all_crate_deps(normal_dev = True),
+)
+
+rust_clippy(
+    name = "cargo_raze_clippy",
+    testonly = True,
+    deps = [
+        ":cargo_raze",
+        ":cargo_raze_bin",
+    ],
 )


acmcarther:
Yeah I'd like to get rid of these github checks ASAP

This [doesn't look like it's being run in buildkite](https://buildkite.com/bazel/cargo-raze/builds/89#0ea26658-719f-4ce1-aaa2-32b1e32c2c7c). Can this be added to the presubmit.yml so it's run?

UebelAndre:
Clippy is a build target, so because we're building `...` it will be picked up.

acmcarther:
Does clippy fail to build if the checks fail? Why is it a regular compilation target instead of a test target?

UebelAndre:
> Does clippy fail to build if the checks fail? Why is it a regular compilation target instead of a test target?

I'm not sure. That's just how it's defined in `rules_rust` (and has been for some time). It will fail to build if there is a clippy defect

acmcarther:
I found the PR where this was implemented in rules rust: https://github.com/bazelbuild/rules_rust/pull/339

It seems like this can be exercised via the implementation as an aspect via

```
$ bazel build --aspects=@rules_rust//rust:rust.bzl%rust_clippy_aspect \
              --output_groups=clippy_checks //...
```

What do you think about just including this in presubmit.yml? I think that might be marginally easier to understand if it fails.

acmcarther:
(i don't really feel strongly about this, so if that's a nuisance to implement, we can go with what's already here)

UebelAndre:
> I found the PR where this was implemented in rules rust: [bazelbuild/rules_rust#339](https://github.com/bazelbuild/rules_rust/pull/339)
> 
> It seems like this can be exercised via the implementation as an aspect via
> 
> ```
> $ bazel build --aspects=@rules_rust//rust:rust.bzl%rust_clippy_aspect \
>               --output_groups=clippy_checks //...
> ```
> 
> What do you think about just including this in presubmit.yml? I think that might be marginally easier to understand if it fails.

I can give it a whirl but don't know if the examples are currently compatible. I'll test this when I have some free time.

dfreese:
> Yeah I'd like to get rid of these github checks ASAP

Can you comment on why?  From my perspective they provided a very clear, fast, and thorough indication of things.

acmcarther:
https://github.com/google/cargo-raze/pull/386#issuecomment-785353979

They seem redundant with BuildKite (and so, seemed more confusing than helpful superficially). I don't feel strongly about this though.

PR #384, opened 2021-02-23
Added documentation for the Bazel-only workflow

UebelAndre:
This PR updates documentation for what users would need to do to be able to add `cargo-raze` to their WORKSPACE file and run it as a tool built purely in Bazel for environments that do not have Cargo installed on the host.

PR #388, opened 2021-02-24
Fixed the need for cargo on the host when running `//:raze`

UebelAndre:
I forgot that I still have cargo installed on my machine when testing this scenario. The `//:raze` target should be a wrapper script that explicitly sets the needed environment variables for running `cargo-raze` without installing cargo components on the host

With the following diff:
```diff
diff --git a/examples/WORKSPACE b/examples/WORKSPACE
index 327a21ef..bf138d0c 100644
--- a/examples/WORKSPACE
+++ b/examples/WORKSPACE
@@ -19,3 +19,12 @@ rust_repositories()
 load("//:repositories.bzl", "repositories")

 repositories()
+
+local_repository(
+    name = "cargo_raze",
+    path = "..",
+)
+load("@cargo_raze//:repositories.bzl", "cargo_raze_repositories")
+cargo_raze_repositories()
+load("@cargo_raze//:transitive_deps.bzl", "cargo_raze_transitive_deps")
```

I'm still able to run raze in another workspace
```
examples % bazel run @cargo_raze//:raze -- --help
```
```
INFO: Invocation ID: d2c294cd-d33d-4df6-b66e-1fbb7a6b49e5
INFO: Analyzed target @cargo_raze//:raze (3 packages loaded, 372 targets configured).
INFO: Found 1 target...
Target @cargo_raze//tools:raze up-to-date:
  bazel-bin/external/cargo_raze/tools/raze
INFO: Elapsed time: 4.432s, Critical Path: 0.04s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
Generate BUILD files for your pre-vendored Cargo dependencies.

Usage:
    cargo-raze (-h | --help)
    cargo-raze (-V | --version)
    cargo-raze [--verbose] [--quiet] [--color=<WHEN>] [--dryrun] [--cargo-bin-path=<PATH>]
               [--manifest-path=<PATH>] [--output=<PATH>] [--generate-lockfile]

Options:
    -h, --help                          Print this message
    -V, --version                       Print version info and exit
    -v, --verbose                       Use verbose output
    -q, --quiet                         No output printed to stdout
    --color=<WHEN>                      Coloring: auto, always, never
    -d, --dryrun                        Do not emit any files
    --cargo-bin-path=<PATH>             Path to the cargo binary to be used for loading workspace metadata
    --manifest-path=<PATH>              Path to the Cargo.toml file to generate BUILD files for
    --output=<PATH>                     Path to output the generated into.
    --generate-lockfile                 Force a new `Cargo.raze.lock` file to be generated
```

### DIFF
--- a/tools/BUILD.bazel
+++ b/tools/BUILD.bazel
@@ -40,3 +40,40 @@ sh_binary(
     ],
     data = ["//impl:cargo_raze_bin"],
 )
+
+alias(
+    name = "cargo",
+    actual = select({
+        "@rules_rust//rust/platform:aarch64-apple-darwin": "@rust_darwin_aarch64//:cargo",
+        "@rules_rust//rust/platform:aarch64-unknown-linux-gnu": "@rust_linux_aarch64//:cargo",
+        "@rules_rust//rust/platform:x86_64-apple-darwin": "@rust_darwin_x86_64//:cargo",
+        "@rules_rust//rust/platform:x86_64-pc-windows-msvc": "@rust_windows_x86_64//:cargo",
+        "@rules_rust//rust/platform:x86_64-unknown-linux-gnu": "@rust_linux_x86_64//:cargo",
+    }),
+)
+
+alias(
+    name = "rustc",
+    actual = select({
+        "@rules_rust//rust/platform:aarch64-apple-darwin": "@rust_darwin_aarch64//:rustc",
+        "@rules_rust//rust/platform:aarch64-unknown-linux-gnu": "@rust_linux_aarch64//:rustc",
+        "@rules_rust//rust/platform:x86_64-apple-darwin": "@rust_darwin_x86_64//:rustc",
+        "@rules_rust//rust/platform:x86_64-pc-windows-msvc": "@rust_windows_x86_64//:rustc",
+        "@rules_rust//rust/platform:x86_64-unknown-linux-gnu": "@rust_linux_x86_64//:rustc",
+    }),
+)
+
+sh_binary(
+    name = "raze",
+    srcs = ["wrapper.sh"],
+    data = [
+        ":cargo",
+        ":rustc",
+        "//impl:cargo_raze_bin",
+    ],
+    env = {
+        "CARGO": "$(execpath :cargo)",
+        "CARGO_RAZE": "$(execpath //impl:cargo_raze_bin)",
+        "RUSTC": "$(execpath :rustc)",
+    },
+)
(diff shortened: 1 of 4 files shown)


UebelAndre:
@acmcarther If you could take a quick peak at this, it makes https://github.com/google/cargo-raze/pull/384 correct 😅

tools/BUILD.bazel
```diff
@@ -40,3 +40,40 @@ sh_binary(
...
+        "@rules_rust//rust/platform:aarch64-apple-darwin": "@rust_darwin_aarch64//:rustc",
+        "@rules_rust//rust/platform:aarch64-unknown-linux-gnu": "@rust_linux_aarch64//:rustc",
+        "@rules_rust//rust/platform:x86_64-apple-darwin": "@rust_darwin_x86_64//:rustc",
+        "@rules_rust//rust/platform:x86_64-pc-windows-msvc": "@rust_windows_x86_64//:rustc",
+        "@rules_rust//rust/platform:x86_64-unknown-linux-gnu": "@rust_linux_x86_64//:rustc",
+    }),
+)
+
+sh_binary(
+    name = "raze",
```
acmcarther, line 67: This is not compatible with windows and probably shouldn't be the default "cargo-raze" target.

acmcarther:
I don't really feel strongly about this, but I suspect this is a step backward for window support.

I'm not sure what doing this properly would look like.

tools/BUILD.bazel
UebelAndre: What makes it incompatible? I thought Bazel expected windows users to have `bash` installed.

UebelAndre:
I unfortunately don't have a windows machine to try this out on but am happy to address any concerns if this doesn't work. The correct solution is likely a custom rule that outputs a batch file but I'm not sure

tools/BUILD.bazel
acmcarther: I guess sh_binary rules work on windows if bash is installed:

https://docs.bazel.build/versions/master/windows.html#sh_binary-and-sh_-rules-and-ctxactionsrun_shell-without-bash

acmcarther:
I guess we'll be able to iterate on this once the windows presubmit targets are restored.

(FYI: you should be able to see the build kite builds now)

UebelAndre:
> (FYI: you should be able to see the build kite builds now)

I can, thank you! Windows dev is just insanely time consuming since I don't know what I'm doing and can only test in the CI builds 😅

PR #393, opened 2021-03-05
Use defs.bzl and use rust_proc_macro rule

hlopko:
This PR updates cargo-raze for https://github.com/bazelbuild/rules_rust/issues/591.

Feedback very welcomed :)


acmcarther:
Usually when a new type of rule is added, an example is included as a coarse grained check that the build rule works and continues to work.

OK to add an example into examples/remote?

UebelAndre:
I think simply running `bazel run //tools:examples_raze` should generate enough uses of this change. It may be worth setting up a [run_targets](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#running-commands-shell-scripts-or-binary-targets) command for `//tools:examples_check` at some point to ensure examples aren't outdated.

PR #396, opened 2021-03-06
Pull @rules_rust in cargo_raze_repositories().

PiotrSikora:

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,9 +232,6 @@ on the host machine. To do so, simply add the following to the WORKSPACE file in
 your project:
 
 ```python
-# Note: The `cargo_raze` repository expects `rules_foreign_cc` and `rules_rust` to
-# have already been specified in the WORKSPACE per the Usage section of the README.
-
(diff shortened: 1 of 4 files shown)


PiotrSikora:
cc @UebelAndre

UebelAndre (approved):
Nice! Thanks for fixing that 😀

UebelAndre:
Oh, actually, I'm not sure if the call to `rust_repositories` should be in that file. Users may not want to call that and instead setup toolchains themselves using `rust_repository_set` directly or something.

PiotrSikora:
> Oh, actually, I'm not sure if the call to `rust_repositories` should be in that file. Users may not want to call that and instead setup toolchains themselves using `rust_repository_set` directly or something.

`rules_repositories()` configures a few default `rust_repository_set()`, but it doesn't prevent people from adding more on their own, so I don't believe it breaks any real use cases.

UebelAndre:
That sounds right to me. Then I think this change is good. Thanks again!

acmcarther:
LGTM

PR #405, opened 2021-03-27
fix aliases for build dependencies not being generated in 0.9.0+

dae:
Applies the same changes made in https://github.com/google/cargo-raze/pull/358
to build dependencies as well. Without this change, dependencies that
are only required at build time have no alias generated, and can only
be accessed via the direct @raze__package string.


dae:
I've pushed some minor fixes that get rustfmt nightly to pass on this change, and created a separate issue about that: #409

dfreese:
Thanks. Lgtm. I merged some updates to formatting.

Is there a clear unit test that can cover this?

dfreese (approved):

Issue #305, opened 2020-11-23
Remove special logic for binary dependencies


UebelAndre:
This is a followup of https://github.com/google/cargo-raze/issues/218 where special support was added to account for binary dependencies in Remote genmode. It's unclear as to whether what the future of binary dependencies will look like for cargo but for this project, this issue should stand as a reminder that we're doing something unique and should at some point conform to whatever `cargo` decides to do. 

https://github.com/rust-lang/cargo/issues/8888 tracks a request to support a `cargo clone` command which can be used to clean up the implementation to support binary dependencies and reduce the amount of dependencies on `cargo-raze`. 

Ultimately, once https://github.com/rust-lang/cargo/issues/8708 is totally resolved then there will likely be an action item here. This issue tracks that progress.

UebelAndre:
https://github.com/rust-lang/rfcs/pull/3028 is also relevant here (and very exciting should it get approved and go through).

UebelAndre:
I think there might be an easy path forward here. It would require updating the `CrateContext` list generated in the planning step to include an entry for the local package. Then some Bazel repository_rule could be written to take a url to some crate and download it and use. The rule would invoke cargo raze with some set of flags that cause it to render a build file for the local newly downloaded crate. This way we can clean up all the cruft created by the `binary_deps` functionality and the need for `Cargo.raze.lock` and be able to more directly rely/reuse existing cargo patterns. This might be the next thing I try to tackle. Interested in your thoughts here @acmcarther

UebelAndre:
https://github.com/rust-lang/cargo/issues/9096 can potentially be a resolution to this issue.

UebelAndre:
For documentation's sake, this feature was added in https://github.com/google/cargo-raze/pull/227

PR #407, opened 2021-03-29
Implemented `binary_deps` behind a feature

UebelAndre:
The `binary_deps` flag brings in some pretty large dependencies. This PR creates the feature `binary_deps` and wraps all uses of that functionality behind it. 

Additionally, the feature `cli` was added to have additional control over the dependencies required by `cargo-raze` when used as a library.


UebelAndre:
The version of rust in github actions was updated to solve for
```
 error[E0658]: `match` is not allowed in a `const fn`
Error:    --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/socket2-0.4.0/src/lib.rs:156:9
    |
156 | /         match address {
157 | |             SocketAddr::V4(_) => Domain::IPV4,
158 | |             SocketAddr::V6(_) => Domain::IPV6,
159 | |         }
    | |_________^
    |
    = note: see issue #49146 <https://github.com/rust-lang/rust/issues/49146> for more information
```

UebelAndre:
@acmcarther Hey, just pinging this one again 🙏

UebelAndre:
@acmcarther Can you please take a look at this 🙏

dfreese:
I'm quite hesitant about feature flags as it makes testing all combinations more difficult.  The cli and binary-deps features here seem to bifurcate the codebase.

Is the intention is to use cargo-raze as a library, rather than a binary?

UebelAndre:
Yes, the intent is to use cargo-raze as a library

illicitonion:
Wider context here is that we're using `cargo-raze` as a library for the `crate_universe` rule in `rules_rust`, and it's much easier for us to bootstrap binaries for `rules_rust` without needing to pull in and build `openssl-sys`, `libgit2-sys`, and `libssh2-sys`, which are only needed for `binary_deps`.

dfreese:
Got it.  So the typical approach here would be to pull things out into a `-core` or `-base` crate.  As the PR stands, a large chunk of the planning code ends up functioning in two different ways, which is really hard to maintain going forward.

Is there was a way for this feature break to happen strictly at a module level?  That will probably help with the discussion here.

UebelAndre:
How does a large chunk of the planning code change?

UebelAndre:
@illicitonion I'll need you to address CLA 🙏

illicitonion:
@googlebot I consent.

UebelAndre:
@dfreese re https://github.com/bazelbuild/rules_rust/issues/415#issuecomment-843382449

> However, there wasn't any further discussion specifically on the module question.

https://github.com/google/cargo-raze/pull/407#issuecomment-814985080 is me trying to understand what you're saying so I can appropriately assess your module quetstioon

dfreese:
@UebelAndre that wasn't intended to be the main point of that comment, the main thing is that planning.rs and metadata.rs now have two paths #[cfg(feature = "binary_deps")] and #[cfg(not(feature = "binary_deps"))] for quite a number of things.  This splits tests and makes it harder to keep track of what any specific code does.

If there could be one spot where we could do:
#[cfg(feature = "binary_deps")]
mod some_name_for_raze_binary_related_functionality
#[cfg(not(feature = "binary_deps"))]
mod fake_some_name_for_raze_binary_related_functionality as some_name_for_raze_binary_related_functionality;

and the rest of the code could remain free of feature gates, that would convince me that this is maintainable within cargo-raze, or that there's a clear set of functionality that can be pulled into a -core or -base crate.

UebelAndre:
@dfreese Can you provide a concrete example of this?

UebelAndre:
Also, I feel so strongly that this feature is just an anti-pattern and should be deleted (https://github.com/google/cargo-raze/issues/305). It's undefined behavior and something that could probably be solved without requiring this tool to have a built-in solution. It should just be removed.

PR #408, opened 2021-04-05
Update deps

titanous:
In the process of chasing down some build errors related to openssl and Rust 1.51, I ended up updating everything.


titanous:
It looks like a test failed while building `socket2` on Rust 1.45, should I find a version constraint that is compatible, or is the minimum Rust version flexible?

titanous:
I bumped the MSRV across all CI jobs to 1.46, given that it is already 1.46 for the tests.

titanous:
The clippy issues are fixed in #406.

acmcarther:
👍

PR #411, opened 2021-04-09
Report error if lockfile generation fails

illicitonion:
Currently we ignore the exit code from this command, swallowing the
error, and typically fail in the subsequent call to `cargo` with a
message along the lines of "No Cargo.lock file was found".

Instead, fatally report the error from `cargo generate-lockfile`.

### DIFF
--- a/impl/src/metadata.rs
+++ b/impl/src/metadata.rs
@@ -91,10 +91,19 @@ impl LockfileGenerator for CargoLockfileGenerator {
     let lockfile_path = crate_root_dir.join("Cargo.lock");
 
     // Generate lockfile
-    std::process::Command::new(&self.cargo_bin_path)
+    let output = std::process::Command::new(&self.cargo_bin_path)
       .arg("generate-lockfile")
       .current_dir(&crate_root_dir)
-      .output()?;
+      .output()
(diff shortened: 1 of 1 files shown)


impl/src/metadata.rs
```diff
@@ -91,11 +91,18 @@ impl LockfileGenerator for CargoLockfileGenerator {
     let lockfile_path = crate_root_dir.join("Cargo.lock");
 
     // Generate lockfile
-    std::process::Command::new(&self.cargo_bin_path)
+    let output = std::process::Command::new(&self.cargo_bin_path)
       .arg("generate-lockfile")
       .current_dir(&crate_root_dir)
       .output()?;
```
dfreese, line 97: LGTM, though, while you're here and working on error handling, would you mind adding a context to the potential error on this line?

dfreese (approved):

impl/src/metadata.rs
illicitonion: Done - thanks!

dfreese (approved):
Thanks!

illicitonion:
@dfreese Could I nudge you for a merge? Thanks!

dfreese:
Sorry!  Approved before CI was ready and forgot about it.

PR #412, opened 2021-04-13
Support additional registries

### LINKED ISSUE #24: Support non-standard registries in remote mode
Remote GenMode is currently masquerading as if it supports arbitrary registries, but in reality it only supports the standard registry due to how it is pulling down the crates:

https://github.com/google/cargo-raze/blob/726148a42640d257577bb3859bae629529d53f3b/src/templates/remote_crates.bzl.template#L11

Normally, this should be using something derived from the registry url, but there is a weird issue described in ancient times here: https://github.com/acmcarther/cargo-raze-vendor/issues/3

illicitonion:
This relies on the registry information being in .cargo/config.toml (and
will proactively error if other keys are found in there).

The reason for this is that we need to run `cargo metadata` before we
even parse out config, and we can't run `cargo metadata` without having
this data active for that `cargo` invocation. It also simplifies
implementation a lot, by just copying around one file rather than
needing to generate config for `cargo` to use.

Fixes #24.

### DIFF
--- a/impl/src/planning/subplanners.rs
+++ b/impl/src/planning/subplanners.rs
@@ -19,7 +19,7 @@ use std::{
   str::FromStr,
 };
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use cargo_lock::SourceId;
 use cargo_metadata::{DepKindInfo, DependencyKind, Node, Package};
 use cargo_platform::Platform;
@@ -34,7 +34,7 @@ use crate::{
   error::{RazeError, PLEASE_FILE_A_BUG},
   metadata::RazeMetadata,
   planning::license,
-  settings::{format_registry_url, CrateSettings, GenMode, RazeSettings},
+  settings::{CrateSettings, GenMode, RazeSettings},
   util,
 };
 
@@ -42,6 +42,7 @@ use super::{
   crate_catalog::{CrateCatalog, CrateCatalogEntry},
   PlannedBuild,
 };
+use url::Url;
 
 /// Named type to reduce declaration noise for deducing the crate contexts
 type CrateContextProduction = (Vec<CrateContext>, Vec<DependencyAlias>);
@@ -411,14 +412,13 @@ impl<'planner> CrateSubplanner<'planner> {
       links: package.links.clone(),
       raze_settings,
       canonical_additional_build_file,
-      source_details: self.produce_source_details(package, &package_root),
+      source_details: self.produce_source_details(
+        &self.settings.registry,
+        package,
+        &package_root,
+      )?,
       expected_build_path: self.crate_catalog_entry.local_build_path(self.settings)?,
       sha256: self.sha256.clone(),
-      registry_url: format_registry_url(
-        &self.settings.registry,
-        &package.name,
-        &package.version.to_string(),
-      ),
       lib_target_name,
       targets,
     };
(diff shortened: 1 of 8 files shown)


impl/src/cargo_config.rs
```diff
@@ -0,0 +1,18 @@
+use serde::Deserialize;
+use std::collections::BTreeMap;
+
+/// A representation of the parts of .cargo/config.toml files understood by cargo-raze.
+/// Unknown fields are set to deny, as cargo-raze does not know how to interpret them,
```
acmcarther, line 5: Am I understanding correctly that this rejects all fields except for `registries`?

We don't control the specification of `.cargo/config.toml` and shouldn't be panicing or returning errors if new fields are added here right? What's the rationale for erroring here?

acmcarther (changes requested):
Generally lg, but i feel somewhat strongly that this should _not_ fail on unknown fields.

impl/src/cargo_config.rs
illicitonion: Yeah, this was me being overly defensive in case there are things in the .cargo/config.toml that would be problematic for raze (say, a config setting which referenced a relative path, or which changed some semantics in some way), but it's probably not actually appropriate for raze - easier for folks to file bugs if raze is broken by a setting they use, than to need to manually support every non-breaking feature :)

I've removed the check

illicitonion:
@acmcarther I addressed the comments - could I nudge you towards a merge? :) Thanks!

illicitonion:
@acmcarther Friendly ping? :)

impl/src/planning/subplanners.rs
```diff
@@ -636,23 +636,77 @@ impl<'planner> CrateSubplanner<'planner> {
...
   /// Generates source details for internal crate.
-  fn produce_source_details(&self, package: &Package, package_root: &Path) -> SourceDetails {
-    SourceDetails {
-      git_data: self.source_id.as_ref().filter(|id| id.is_git()).map(|id| {
+  fn produce_source_details(
+    &self,
+    crates_io_template: &str,
+    package: &Package,
+    package_root: &Path,
+  ) -> Result<SourceDetails> {
```
UebelAndre, line 644: If this now returns a `Result`, should we cleanup the use of `unwrap` here?

impl/src/planning/subplanners.rs
```diff
@@ -636,23 +636,77 @@ impl<'planner> CrateSubplanner<'planner> {
...
+            .parse()?,
+        );
+      }
+      if let Some(index_url) = source.repr.strip_prefix("registry+") {
+        let index = crates_index::BareIndex::from_url(index_url)?;
+        let index = index.open_or_clone()?;
+        let config = index.index_config()?;
+        let url = config.download_url(&package.name, &package.version.to_string());
+        if let Some(url) = url {
+          return Ok(url.parse()?);
```
UebelAndre, line 696: Isn't this already returning a `Result`?
```suggestion
          return url.parse();
```

UebelAndre (approved):
This looks good to me 😄. A couple of small comments and can you run `bazel run //tools:bootstrap && bazel run //tools:examples_raze` locally just to make sure nothing changed about the generated files?

impl/src/planning/subplanners.rs
illicitonion: Good call - done

impl/src/planning/subplanners.rs
illicitonion: Ish - the `?` is doing an `Into` conversion on the error type. But I've changed this to actually do some wrapping of error messages :)

illicitonion:
> This looks good to me 😄. A couple of small comments and can you run `bazel run //tools:bootstrap && bazel run //tools:examples_raze && bazel test //...` locally just to make sure nothing changed about the generated files?

I have done - the changes are included in this PR :) (Specifically adding the `serde` feature to `url` :))

UebelAndre (approved):
Thanks for doing the cleanup! Looks good to me 😄

PR #415, opened 2021-04-19
Made some structs sortable and deserializable

UebelAndre:
This is incredibly useful when using `cargo-raze` as a library.


UebelAndre:
@dfreese would you be able to take a look at this?

UebelAndre:
Actually, this can wait until https://github.com/google/cargo-raze/pull/282 goes in. I'd rather not risk a conflict there.

dfreese:
I have reservations with how we'd need to support this sort of use case.  Since the cargo-raze library is only intended to be used by the cargo-raze-bin target, these are only public for the purposes of main.

Specifically with this change, the change from HashMap to BTreeMap is a pessimization, since we don't need the ordering.  That would also be signal to me, later on, that order is important.

The use-case for Deserialize is fairly obvious to me, but can you explain why you'd need `CrateSettings` to be ordered?

UebelAndre:
> The use-case for Deserialize is fairly obvious to me, but can you explain why you'd need `CrateSettings` to be ordered?

It's so `crate_universe` can render lockfiles with minimal diffs when things update. This is accomplished by sorting the output which includes `CrateSettings`. https://github.com/bazelbuild/rules_rust/pull/692

dfreese:
Sorry for the long delay.  #425 (an update of #282) has been merged, so I'm good with this being merged if you'd want to update it to latest main.

PR #425, opened 2021-07-05
Rework planning to use resolve tree and not packages (updated)

sayrer:
This addresses @dfreese's comments in #282 and preserves @GregBowyer's commit in the history.

### DIFF
--- a/impl/src/context.rs
+++ b/impl/src/context.rs
@@ -12,7 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::path::PathBuf;
+use std::{
+  collections::BTreeSet,
+  hash::{Hash, Hasher},
+  path::PathBuf,
+};
 
 use crate::settings::CrateSettings;
 use semver::Version;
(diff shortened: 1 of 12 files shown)


sayrer:
I ran this patch on a large repo, and it seems to produce identical output to #282.

impl/src/context.rs
```diff
@@ -92,7 +95,7 @@ pub struct CrateDependencyContext {
   // build_data_dependencies can only be set when using cargo-raze as a library at the moment.
   pub build_data_dependencies: Vec<BuildableDependency>,
   pub dev_dependencies: Vec<BuildableDependency>,
-  pub aliased_dependencies: Vec<DependencyAlias>,
+  pub aliased_dependencies: BTreeMap<String, DependencyAlias>,
```
dfreese, line 98: Can you comment on the meaning of the string here?

Not  necessary, but a struct, holding a string may provide more meaning long term if that's fairly straight-forward.

dfreese (approved):
LGTM, one comment and associated question.  Thanks for following up on this.

The google CLA bot will require both you and @GregBowyer to sign off, since it requires consent from all commit authors.  Squashing it down may be a relevant option here.

impl/src/context.rs
sayrer: I'm happy to work on this a bit more, but I might suggest that it would be easier to commit this change and iterate, or just let you change my patch to your satisfaction in a followup.

The pattern here is something I would do if I reviewed a coworker's code on a daily basis, and I wanted to make that process smoother going forward. But I think I'll only be sending cargo-raze patches here and there.

As for the CLAs, mine is signed. There was also a warning about this in #282 regarding @GregBowyer's sign-off.

GregBowyer:
@googlebot I consent.

PR #427, opened 2021-07-08
Bump dep on semver to 1.0.0

illicitonion:


dfreese (approved):
LGTM.  If you don't mind, I might wait on #425.  I had a minor comment there, and then I think it's good to go.

dfreese:
@illicitonion sorry for the delay on that, but #425 has been merged.

impl/src/metadata.rs
```diff
@@ -516,9 +515,7 @@ pub mod tests {
 
   impl DummyCargoMetadataFetcher {
     fn render_metadata(&self, mock_workspace_path: &Path) -> Option<Metadata> {
-      if self.metadata_template.is_none() {
-        return None;
-      }
+      self.metadata_template.as_ref()?;
```
dfreese, line 518: Can you store this to a value and use it later on line 532?  That'll remove both the unwrap in those lines, and make the intent of this more clear.

dfreese (approved):
Looks like it's breaking the bazel build of this by not having serde-derive as a proc macro dependency?  I'm not super familiar with the bazel build for this, but does something need to be regenerated?  @UebelAndre, not sure if you'd see an obvious thing right away here.

Overall, looks good to me, though.

UebelAndre:
Yes, if the dependencies are updated, then `cargo-raze` will need to be re-bootstrapped
```
bazel run //tools:bootstrap
```
Then commit the results and things should be good.

PS: @dfreese Can you check slack?

illicitonion:
Yay, rebasing past https://github.com/google/cargo-raze/pull/441 fixed CI :)

PR #437, opened 2021-08-18
Don't allow default deps in planning of platform deps.

### LINKED ISSUE #389: BUG: flate2 (1.0.20) deps are incorrectly generated
Deps are generated as
```
    deps = [
        "@raze__cfg_if__1_0_0//:cfg_if",
        "@raze__crc32fast__1_2_1//:crc32fast",
        "@raze__libc__0_2_86//:libc",
        "@raze__miniz_oxide__0_4_4//:miniz_oxide",
    ] + selects.with_or({
        # cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))
        (
            "@rules_rust//rust/platform:wasm32-unknown-unknown",
            "@rules_rust//rust/platform:wasm32-wasi",
        ): [
            "@raze__miniz_oxide__0_4_4//:miniz_oxide",
        ],
        "//conditions:default": [],
    }),
```

And you get
`
ERROR: /home/palfrey/.cache/bazel/_bazel_palfrey/1e69055091f05c565038b6edab4c56d2/external/raze__flate2__1_0_20/BUILD.bazel:77:13: Label '@raze__miniz_oxide__0_4_4//:miniz_oxide' is duplicated in the 'deps' attribute of rule 'flate2'
`

This is because in [the Cargo.toml](https://github.com/rust-lang/flate2-rs/blob/master/Cargo.toml) `miniz_oxide` is optional by default, except for in the wasm32 target when it's non-optional.

sayrer:
I'll add some tests if folks think this is the right approach. It doesn't fix every edge case I'm aware of, but adding coverage for this very common case would help move things forward. It's very easy to hit this bug if you enable a lot of platforms in raze settings. For example, this patch fixes #389 for me.

### DIFF
--- a/impl/src/context.rs
+++ b/impl/src/context.rs
@@ -108,17 +108,28 @@ impl CrateDependencyContext {
 
   pub fn subtract(&mut self, other: &CrateDependencyContext) {
     self.dependencies = subtract_deps(&self.dependencies, &other.dependencies);
-    self.proc_macro_dependencies = subtract_deps(&self.proc_macro_dependencies, &other.proc_macro_dependencies);
+    self.proc_macro_dependencies = subtract_deps(
+      &self.proc_macro_dependencies,
+      &other.proc_macro_dependencies,
+    );
     self.build_dependencies = subtract_deps(&self.build_dependencies, &other.build_dependencies);
-    self.build_proc_macro_dependencies = subtract_deps(&self.build_proc_macro_dependencies, &other.build_proc_macro_dependencies);
+    self.build_proc_macro_dependencies = subtract_deps(
+      &self.build_proc_macro_dependencies,
+      &other.build_proc_macro_dependencies,
+    );
     self.dev_dependencies = subtract_deps(&self.dev_dependencies, &other.dev_dependencies);
   }
 }
 
-fn subtract_deps(own: &Vec<BuildableDependency>, other: &Vec<BuildableDependency>) -> Vec<BuildableDependency> {
-  own.iter().filter(|dep| {
-    !other.contains(dep)
-  }).cloned().collect()
+fn subtract_deps(
+  own: &[BuildableDependency],
+  other: &[BuildableDependency],
+) -> Vec<BuildableDependency> {
+  own
+    .iter()
+    .filter(|dep| !other.contains(dep))
+    .cloned()
+    .collect()
 }
 
 #[derive(Debug, Clone, Serialize, PartialEq, Eq, PartialOrd, Ord)]
--- a/impl/src/context.rs
+++ b/impl/src/context.rs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{collections::BTreeMap, path::PathBuf};
+use std::{
+  collections::{BTreeMap, BTreeSet},
+  path::PathBuf,
+};
 
 use crate::settings::CrateSettings;
 use semver::Version;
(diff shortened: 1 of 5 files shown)


sayrer:
Looks like there are a bunch of Clippy errors in code I didn't change. I'll follow up on that if this patch should move forward.

sayrer:
It looks like most of the other problems of this kind I'm encountering are due to https://github.com/bazelbuild/bazel/issues/13785

illicitonion:
This looks reasonable to me.

FWIW I _think_ a way to work around https://github.com/bazelbuild/bazel/issues/13785 would be to stop concatenating selects - to make `deps` always be exactly one `select`, and to flatten/duplicate all of the default deps into all branches. So instead of:

```python
    deps = [
        "@raze__cfg_if__1_0_0//:cfg_if",
        "@raze__crc32fast__1_2_1//:crc32fast",
        "@raze__libc__0_2_86//:libc",
        "@raze__miniz_oxide__0_4_4//:miniz_oxide",
    ] + selects.with_or({
        # cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))
        (
            "@rules_rust//rust/platform:wasm32-unknown-unknown",
            "@rules_rust//rust/platform:wasm32-wasi",
        ): [
            "@raze__miniz_oxide__0_4_4//:miniz_oxide",
        ],
        "//conditions:default": [],
    }),
```

we would generate

```python
    deps = selects.with_or({
        (
            "@rules_rust//rust/platform:whatever",
        ): [
            "@raze__cfg_if__1_0_0//:cfg_if",
            "@raze__crc32fast__1_2_1//:crc32fast",
            "@raze__libc__0_2_86//:libc",
        ],
        (
            "@rules_rust//rust/platform:whatever_else",
        ): [
            "@raze__cfg_if__1_0_0//:cfg_if",
            "@raze__crc32fast__1_2_1//:crc32fast",
            "@raze__libc__0_2_86//:libc",
        ],
        (
            "@rules_rust//rust/platform:wasm32-unknown-unknown",
            "@rules_rust//rust/platform:wasm32-wasi",
        ): [
            "@raze__cfg_if__1_0_0//:cfg_if",
            "@raze__crc32fast__1_2_1//:crc32fast",
            "@raze__libc__0_2_86//:libc",
            "@raze__miniz_oxide__0_4_4//:miniz_oxide",
        ],
        "//conditions:default": [],
    }),
```

Bazel today reasons about single selects fine, it's just the concatenation that it runs into trouble with.

It's less documenting in what it generates, but it's generated code anyway, so... We could generate comments if we really wanted to make it easier to read...

sayrer:
Hmm. While that solution would function, it wouldn't be great for my use case. I've added local crate and test rendering in my fork, patch for that coming later. My goal is to produce readable output, so that a project can convert to Bazel and drop Cargo--mostly for private monorepo-style things, where publishing crates is not a goal. Also, the comments in the code bazelbuild/bazel#13785 seem to admit it is unsound, which seems bad.

In any case, it looks like we can discuss bazelbuild/bazel#13785 elsewhere without blocking this PR.

dfreese:
I'm generally good with the approach.  Along the lines of what's being done in #415, it does seem imply
dependencies, proc_macro_dependencies, build_dependencies, build_proc_macro_dependencies, and dev_dependencies in CrateDependenciesContext should be sets of some kind, probably BTreeSet, rather than hand implementing set subtraction.

I fixed the clippy and rustfmt errors upstream.

sayrer:
This seems to address the comments so far. If you comment out the line that calls `subtract`, the new test will fail.

dfreese (approved):
Looks great.  Thanks for running with that comment.

PR #441, opened 2021-08-24
Fixed regression for proc-macro crates

UebelAndre:
This fixes `crates.bzl` never having any crates defined in the `_PROC_MACRO_DEPENDENCIES` map.

This fixes a regression from https://github.com/google/cargo-raze/pull/425 (or https://github.com/google/cargo-raze/pull/282) and regenerates cargo-raze outputs (which should have been done on the merged commit). 

This also fixes a regression in how BUILD files were rendered, resulting in buildifier defects. I cherry picked a change I had an another branch which is ahead of `main` and contains the fix. This change includes https://github.com/google/cargo-raze/pull/393

Finally, this fixes another regression where `extra_aliased_targets` was being ignored.


illicitonion (approved):
LGTM - thanks!

(Wow, the amount of regen makes these PRs kind of hard to review...)

PR #444, opened 2021-08-29
Enable rust_analyzer and VSCode integration.

sayrer:
This worked for me, but I think this should wait until https://github.com/bazelbuild/rules_rust/pull/907 is done.


sayrer:
It looks like the one test error is due to

`Error: rust_repositories() got unexpected keyword argument: include_rustc_srcs`

running on Bazel 4.0.0. While we wait for that `rules_rust` patch to land, I think we could probably bump the versions here.

dfreese:
Would this be good to go now that https://github.com/bazelbuild/rules_rust/pull/1010 is merged?

PR #452, opened 2021-11-15
Update rules_rust and rules_foreign_cc to latest

djmarcin:
These two repos have breaking changes in latest. This means that importing this repo in a `WORKSPACE` fails if newer versions of rules_rust and/or rules_foreign_cc are already imported.

### DIFF
--- a/third_party/openssl/BUILD.openssl.bazel
+++ b/third_party/openssl/BUILD.openssl.bazel
@@ -20,10 +20,6 @@ CONFIGURE_OPTIONS = [
 configure_make(
     name = "openssl",
     configure_command = "config",
-    env = select({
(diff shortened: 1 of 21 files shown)


abrisco (approved):
Nice! Thank you!

UebelAndre (approved):

third_party/openssl/BUILD.openssl.bazel
```diff
@@ -38,8 +38,8 @@ configure_make(
     }),
     lib_source = ":all_srcs",
     out_static_libs = [
```
djmarcin, line 40: The ordering here is apparently significant. Without reordering, rust linking fails. I'm not sure if that's an issue in some recent version of rules_rust or what.

PR #455, opened 2021-12-03
Preserve original crate name as a Bazel target tag

vmax:
As `cargo raze` sanitizes target names (replacing `-` with `_` - could you shed some light on why that needs to be done?), it's impossible to retrieve the name of the original crate.

As a solution, we're including the original name of the crate into Bazel target's `tags`.

### DIFF
--- a/impl/src/rendering/templates/crate.BUILD.template
+++ b/impl/src/rendering/templates/crate.BUILD.template
@@ -35,6 +35,7 @@ licenses([
 {%      include "templates/partials/build_script.template" %}
 {%- endif -%}
 {%- for target in crate.targets -%}
+{%-     set target_name_original = target.name %}
 {%-     set target_name_sanitized = target.name | replace(from="-", to="_") %}
 {%-     if target.kind == "bin" %}
 
--- a/impl/src/rendering/templates/partials/common_attrs.template
+++ b/impl/src/rendering/templates/partials/common_attrs.template
@@ -53,5 +53,6 @@
     tags = [
         "cargo-raze",
         "manual",
+        "crate-name={{ target_name_original }}"
     ],
     version = "{{ crate.pkg_version }}",


dfreese:
`-` isn't a valid character for bazel targets.

Can you give a bit more detail about your use case for the information?

Just wanting to make sure we head towards the right thing.  Specifying the crate_name attribute with the original name might also be a possibility.

vmax:
@dfreese 

>- isn't a valid character for bazel targets.

I don't think it's true _in the general case_. You can't use it for Python targets (because Python disallows `-` in package/source file names), but majority of rules work just fine with targets being named `-` (e.g. Java, Rust rules and others)

>Can you give a bit more detail about your use case for the information?

Currently we're working on a set Bazel rules that allows users to assemble & deploy `.crate` packages; you can see some ongoing work [here](https://github.com/vaticle/bazel-distribution/blob/1a501f07391bf199a9db83abd5ebfc8905c4e889/crates/rules.bzl).

And as part of that we also want to automatically generate `Cargo.toml` and collect `[dependencies]` that go into it (see
vaticle/bazel-distribution#323 and vaticle/bazel-distribution#325 for more detail)

lolski:
> @dfreese
> 
> > * isn't a valid character for bazel targets.
> 
> I don't think it's true _in the general case_. You can't use it for Python targets (because Python disallows `-` in package/source file names), but majority of rules work just fine with targets being named `-` (e.g. Java, Rust rules and others)
> 

Yeah, I believe this is correct.

If you want to generate Bazel target for Python, then you *must* replace any `-` with `_`. However given that you're generating it for Rust, which has no such name restriction, we don't need to. Right @vmax?

Just to give you an example, we've got a target called `release-validate-deps` (ie., uses `-`) that we're calling in our CI: https://github.com/vaticle/typedb/blob/master/.grabl/automation.yml#L170.

> > Can you give a bit more detail about your use case for the information?
> 
> Currently we're working on a set Bazel rules that allows users to assemble & deploy `.crate` packages; you can see some ongoing work [here](https://github.com/vaticle/bazel-distribution/blob/1a501f07391bf199a9db83abd5ebfc8905c4e889/crates/rules.bzl).
> 
> And as part of that we also want to automatically generate `Cargo.toml` and collect `[dependencies]` that go into it (see [vaticle/bazel-distribution#323](https://github.com/vaticle/bazel-distribution/issues/323) and [vaticle/bazel-distribution#325](https://github.com/vaticle/bazel-distribution/pull/325) for more detail)

dfreese:
Thanks for the clarification (and for waiting).  I wanted to see if there was a better way to surface this information.  This seems to be something that would be good to expose as a feature flag for folks who want to have the option to have the actual name of the targets match the crate.  Unfortunately, doing that seems less trivial that I would like due to the various places we perform this `-`->`_` substitution.

I'm generally good with this being included for now, and then we can re-evaluate if doing the clean-up to include that flag makes sense.

vmax:
@dfreese rebased

dfreese:
In v0.14.0.

vmax:
@dfreese thank you!

PR #462, opened 2022-01-02
Sort "tags". 

PiotrSikora:
While there, regenerate examples and third_party.

Missed in #455.


illicitonion (approved):

illicitonion:
It looks like this breaks buildifier? I guess maybe tags are meant to be sorted?

PiotrSikora:
@illicitonion great find, thanks for spotting it! I fixed the template and regenerated the artifacts. PTAL.

Perhaps `v0.14.1` is in order, once this is merged in?

dfreese (approved):
Thanks!

dfreese:
In v0.14.1.

PR #475, opened 2022-03-21
Add aliases support to build_script.template

scasagrande:
This PR adds support for generated BUILD files containing `cargo_build_script` to make use of the `aliases` keyword.

The need for this was discovered when using `cargo raze` on a project that pulled in [value-bag](https://github.com/sval-rs/value-bag) as a sub-dependency. This project requires an alias of `version-check` as `rustc` in order to run `build.rs`.

Copy-pasting the code block from [common-attrs](https://github.com/google/cargo-raze/blob/main/impl/src/rendering/templates/partials/common_attrs.template#L3) where this has already been solved for `rust_library` adds the same solution to `cargo_build_script`.

I am a Rust noob, so if this isn't the best-practices solution to this problem, feel free to close.


scasagrande:
Thanks @dfreese ! I have updated my branch, and it looks like we just need CI to be approved to run again

dfreese:
Still have this on my plate.  #477 made me want to double check first that this wouldn't cause breakages by unintentionally adding aliases on build dependencies.  I don't know off hand if they should be shared between build and normal deps.

scasagrande:
No problem @dfreese . I'm jumping between many different problems in my bazel adventure, so as I learn more I might be able to cycle back and more rigorously implement this change

PR #476, opened 2022-03-22
Add rust analyzer

sayrer:


transitive_deps.bzl
```diff
@@ -6,4 +6,4 @@ load("@rules_rust//rust:repositories.bzl", "rust_repositories")
 def cargo_raze_transitive_deps():
     """Loads all dependnecies from repositories required for cargo-raze"""
     rules_foreign_cc_dependencies()
-    rust_repositories()
+    rust_repositories(include_rustc_srcs = True)
```
dfreese, line 9: Is both this and the `RULES_RUST_TOOLCHAIN_INCLUDE_RUSTC_SRCS` change in the bazelrc needed?

dfreese (approved):
Thanks for fixing all of the clippy issues on the new stable.  One quick question.

transitive_deps.bzl
sayrer: You're right, it's necessary to do both.

PR #478, opened 2022-03-28
Improve per-platform feature selection.

### LINKED ISSUE #451: Duplicated deps when multiple targets share dependencies
`errno` crate lists multiple targets that depend on `libc` crate ([Cargo.toml](https://github.com/lambda-fairy/rust-errno/blob/main/Cargo.toml)):
```
[target.'cfg(unix)'.dependencies]
libc = "0.2"

[target.'cfg(windows)'.dependencies]
winapi = { version = "0.3", features = ["errhandlingapi", "minwindef"] }

[target.'cfg(target_os="dragonfly")'.dependencies]
errno-dragonfly = "0.1.1"

[target.'cfg(target_os="wasi")'.dependencies]
libc = "0.2"

[target.'cfg(target_os="hermit")'.dependencies]
libc = "0.2"
```
which gets translated by `cargo raze` into this ([BUILD.errno-0.2.8.bazel](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/57ec3cd42e382710adda5322624e4ae0069f6065/bazel/cargo/remote/BUILD.errno-0.2.8.bazel)):
```
    deps = [
    ] + selects.with_or({
        # cfg(target_os = "wasi")
        (
            "@rules_rust//rust/platform:wasm32-wasi",
        ): [
            "@proxy_wasm_cpp_host__libc__0_2_106//:libc",
        ],
        "//conditions:default": [],
    }) + selects.with_or({
        # cfg(unix)
        (
            "@rules_rust//rust/platform:aarch64-apple-darwin",
            "@rules_rust//rust/platform:aarch64-apple-ios",
            "@rules_rust//rust/platform:aarch64-linux-android",
            "@rules_rust//rust/platform:aarch64-unknown-linux-gnu",
            "@rules_rust//rust/platform:arm-unknown-linux-gnueabi",
            "@rules_rust//rust/platform:i686-apple-darwin",
            "@rules_rust//rust/platform:i686-linux-android",
            "@rules_rust//rust/platform:i686-unknown-freebsd",
            "@rules_rust//rust/platform:i686-unknown-linux-gnu",
            "@rules_rust//rust/platform:powerpc-unknown-linux-gnu",
            "@rules_rust//rust/platform:s390x-unknown-linux-gnu",
            "@rules_rust//rust/platform:x86_64-apple-darwin",
            "@rules_rust//rust/platform:x86_64-apple-ios",
            "@rules_rust//rust/platform:x86_64-linux-android",
            "@rules_rust//rust/platform:x86_64-unknown-freebsd",
            "@rules_rust//rust/platform:x86_64-unknown-linux-gnu",
        ): [
            "@proxy_wasm_cpp_host__libc__0_2_106//:libc",
        ],
        "//conditions:default": [],
    }) + selects.with_or({
        # cfg(windows)
        (
            "@rules_rust//rust/platform:i686-pc-windows-msvc",
            "@rules_rust//rust/platform:x86_64-pc-windows-msvc",
        ): [
            "@proxy_wasm_cpp_host__winapi__0_3_9//:winapi",
        ],
        "//conditions:default": [],
    }),
```
However, such duplicates are disallowed in Bazel and `bazel query` treats it as an error (see: https://github.com/bazelbuild/bazel/issues/13785):
```
$ bazel query 'deps(set(@proxy_wasm_cpp_host__errno__0_2_8//:errno))'
```
```
ERROR: .../external/proxy_wasm_cpp_host__errno__0_2_8/BUILD.bazel:34:13: Label '@proxy_wasm_cpp_host__libc__0_2_106//:libc' is duplicated in the 'deps' attribute of rule 'errno'
ERROR: .../external/proxy_wasm_cpp_host__errno__0_2_8/BUILD.bazel:34:13: Target '@proxy_wasm_cpp_host__errno__0_2_8//:src/lib.rs' contains an error and its package is in error and referenced by '@proxy_wasm_cpp_host__errno__0_2_8//:errno'
ERROR: .../external/proxy_wasm_cpp_host__errno__0_2_8/BUILD.bazel:34:13: Target '@proxy_wasm_cpp_host__errno__0_2_8//:src/hermit.rs' contains an error and its package is in error and referenced by '@proxy_wasm_cpp_host__errno__0_2_8//:errno'
ERROR: .../external/proxy_wasm_cpp_host__errno__0_2_8/BUILD.bazel:34:13: Target '@proxy_wasm_cpp_host__errno__0_2_8//:src/unix.rs' contains an error and its package is in error and referenced by '@proxy_wasm_cpp_host__errno__0_2_8//:errno'
ERROR: .../external/proxy_wasm_cpp_host__errno__0_2_8/BUILD.bazel:34:13: Target '@proxy_wasm_cpp_host__errno__0_2_8//:src/wasi.rs' contains an error and its package is in error and referenced by '@proxy_wasm_cpp_host__errno__0_2_8//:errno'
ERROR: .../external/proxy_wasm_cpp_host__errno__0_2_8/BUILD.bazel:34:13: Target '@proxy_wasm_cpp_host__errno__0_2_8//:src/windows.rs' contains an error and its package is in error and referenced by '@proxy_wasm_cpp_host__errno__0_2_8//:errno'
ERROR: Evaluation of query "deps(set(@proxy_wasm_cpp_host__errno__0_2_8//:errno))" failed: errors were encountered while computing transitive closure
```

This could be solved by either generating a single `select` (as suggested here: https://github.com/google/cargo-raze/pull/437#issuecomment-901776797) or by combining all `select` conditions that evaluate into the same value into one (although, that would work only when dependencies are exactly the same and not simply overlapping). cc @illicitonion

sayrer:
This patch should fix #451.

The patch runs `cargo-tree` to determine which features are enabled on each platform, and then groups them by using vectors of platforms as map keys. It's a lot of code, and there are no tests yet. But I wanted to post this initial effort and get some feedback from the code owners. @illicitonion has already seen this code, but I believe it will be new to @dfreese and @PiotrSikora (who filed the linked issue).

The new file is copyright AgileBits, as this was work for hire for them. They are fine with contributing, and my repo has been public for months.


sayrer:
I've looked around the examples and the impl/ directory, and I'm not sure where to add this crate: https://github.com/lambda-fairy/rust-errno/blob/main/Cargo.toml

This patch should consolidate the `unix` and `wasi` requirements for the `libc` crate.

dfreese (commented):
Overall, I'm not thrilled by the idea of having to invoke and parse cargo-tree for every platform, but cargo isn't giving good alternatives.

I'm generally supportive of the approach.  Are there any other tactics for this in the wild?

illicitonion:
I'm also generally supportive of the approach given the state of cargo, but sad that there don't seem to be viable alternatives - thanks so much for putting this together and upstreaming it 👍

impl/src/features.rs
```diff
@@ -0,0 +1,276 @@
...
+    .arg("--prefix=none")
+    .arg("--frozen")
+    .arg(format!("--target={}", triple))
+    .arg("--format={p}|{f}|"); // The format to print output with
+
+  let tree_output = cargo_tree.output()?;
+  assert!(tree_output.status.success());
+
+  let text = String::from_utf8(tree_output.stdout)?;
+  let mut crates: BTreeSet<String> = BTreeSet::new();
```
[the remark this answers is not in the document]
sayrer: I think what I've done should work. I tried a few different output formats before what's there now, and left that extra code in.

sayrer:
It looks like this patch will not fix `errno`, but it will fix a bunch of related issues with features. I must have another, shorter, patch in my code for that problem.

As for approaches to this problem, I only found two choices: Import cargo as a crate directly, and use its internal resolver implementation (which is `pub`), or do what I have here.

I chose to use `cargo-tree`, because taking cargo internals as a dependency seemed even less appealing than what I have here. I haven't seen anyone else solve this problem in the wild, and I definitely looked before writing this code.

impl/src/features.rs
```diff
@@ -0,0 +1,276 @@
...
+fn consolidate_features(
+  pkg: (PackageId, BTreeMap<String, BTreeSet<String>>),
+) -> (PackageId, Features) {
+  let (id, features) = pkg;
+
+  // Find the features common to all targets
+  let sets: Vec<&BTreeSet<String>> = features.values().collect();
+  let common_features = sets.iter().skip(1).fold(sets[0].clone(), |acc, hs| {
+    acc.intersection(hs).cloned().collect()
+  });
```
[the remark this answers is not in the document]
sayrer: I tried making this change, but it made everything below messier, at least with my hand at the keyboard. Can you show the whole function written this way? I'm afraid I'm missing the target.

impl/src/features.rs
```diff
@@ -0,0 +1,276 @@
...
+    .arg("tree")
+    .arg("--prefix=none")
+    .arg("--frozen")
+    .arg(format!("--target={}", triple))
+    .arg("--format={p}|{f}|"); // The format to print output with
+
+  let tree_output = cargo_tree.output()?;
+  assert!(tree_output.status.success());
+
+  let text = String::from_utf8(tree_output.stdout)?;
```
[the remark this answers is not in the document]
sayrer: I will do this once the two more difficult problems below are resolved.

impl/src/features.rs
```diff
@@ -0,0 +1,276 @@
...
+}
+
+fn make_package_map(
+  crates: Vec<String>,
+  packages: &[Package],
+) -> Result<BTreeMap<PackageId, BTreeSet<String>>> {
+  let mut package_map: BTreeMap<PackageId, BTreeSet<String>> = BTreeMap::new();
+  for c in &crates {
+    let (name, version, features) = process_line(c)?;
+    let id = find_package_id(name, version, packages)?;
```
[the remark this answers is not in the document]
sayrer: I definitely didn't give much though to the order of growth here, figuring it would only be a couple thousand crates at most, but maybe I haven't seen the largest use cases.

Having a lookup by package id is definitely helpful in `subplanners.rs` around line 170 here:

`      features: &self.metadata.features.get(&node.id),`

I think the fix here might be to change the data structure for `packages`, ahead of this loop.

leebenson:
Very much looking forward to this landing!

dfreese:
Sorry, missed that you had also responded to review comments as well.  I'm good with this as long as the new code can be covered by unit tests where reasonable.

The copyright/license notice probably needs to be updated, but I'm not entirely sure what needs to happen given the CLA.  @hlopko do you have any familiarity with that?

third_party/cargo/remote/BUILD.tokio-1.4.0.bazel
```diff
@@ -60,8 +57,31 @@ cargo_build_script(
...
+        "//conditions:default": [],
+    }) + selects.with_or({
+        (
+            "@rules_rust//rust/platform:i686-pc-windows-msvc",
+            "@rules_rust//rust/platform:x86_64-pc-windows-msvc",
+        ): [
+            "winapi",
+        ],
+        "//conditions:default": [],
+    }),
```
sayrer, line 84: This change is what we're looking for here. Building `winapi` only on Windows.

impl/src/features.rs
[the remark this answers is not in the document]
sayrer: I'm going to resolve this one. Feel free to re-open if you feel strongly, but I didn't think modifying these lines improved much.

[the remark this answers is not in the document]
dfreese: That's fine by me if it was making things less clear.

dfreese (approved):

.github/workflows/ci.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
       matrix:
         rust:
           - stable
-          - 1.46.0
```
dfreese, line 13: Was there a specific feature/crate that was necessitating 1.51?

sayrer: 1.51 is required to use `resolver = "2"` in the project's `Cargo.toml`, but it can be reverted if we don't want to build `raze` itself with the new resolver. I think we should, but it can be reverted easily enough.

sayrer:
I have some better tests coming, and I still need to go through the `?` early exits and give them some messages.

dfreese:
> I have some better tests coming, and I still need to go through the `?` early exits and give them some messages.

Sounds good. Thanks for the update!

sayrer:
@dfreese I think this PR now addresses your comments, and tests all the way through the planning phase. lmk what should be done about the copyright. I also have 3-4 smaller improvements that depend on this branch.

impl/src/features.rs
```diff
@@ -0,0 +1,482 @@
...
+  Ok(
+    transpose_keys(triple_map)
+      .into_iter()
+      .map(consolidate_features)
+      .collect(),
+  )
+}
+
+fn clean_cargo_tree_output(cargo_tree_output: &str) -> Vec<String> {
+  let mut crates: Vec<String> = vec![];
```
dfreese, line 123: Can this be changed to:
```suggestion
  let mut crates = Vec::<String>::new();
```
or even just
```suggestion
  let mut crates = Vec::new();
```
?

sayrer:
This needs to be rebased after #506, but it looks like there are some chores to do in that context. I'll wait a bit for that stuff, and check in again.

impl/src/rendering/bazel.rs
```diff
@@ -533,7 +538,10 @@ mod tests {
       pkg_name: "test-binary".to_owned(),
       pkg_version: Version::parse("1.1.1").unwrap(),
       edition: "2015".to_owned(),
-      features: vec!["feature1".to_owned(), "feature2".to_owned()],
+      features: Features {
+        features: vec!["feature1".to_owned(), "feature2".to_owned()].to_owned(),
```
dfreese, line 542: is the `.to_owned()` on the `vec!` necessary?

dfreese: same goes for lines 592 and 639.

impl/src/rendering/templates/partials/features.template
```diff
@@ -0,0 +1,18 @@
+
+    crate_features = [
```
dfreese, line 2: Know you're preserving behavior here, but since these lines will be changing in quite a few files, if there are no features, can this not be serialized?

dfreese (commented):
Handful of comments.  Generally looks good to me.  Thanks for putting all of this together.

Can you rebase to HEAD as well just to make sure no issues have cropped up?

sayrer:
Will address these, but I think the Dependabot stuff and #482 should go in first. Not in a rush over the weekend.

sayrer:
OK, I think this is all set.

impl/src/rendering/templates/partials/features.template
```diff
@@ -0,0 +1,21 @@
...
+        ): [
+    {%- for feature in targeted_feature.features %}
+            "{{ feature }}",
+    {%- endfor %}
+        ],
+        "//conditions:default": [],
+    })
+    {%- endfor -%},
+    {%- endif %}
+
```
dfreese, line 21: Is this extra line causing that new line to be added to all of the build files?

dfreese (approved):
Lgtm, one question.

impl/src/rendering/templates/partials/features.template
sayrer: With my last patch, the generated .bazel files match the buildifier versions for me. I was missing a subtlety of Tera.

dfreese:
Great. Thanks. Can you merge with main and rerun bootstrap?  Then I think this is good to go. Thanks for all the work on this.

sayrer:
Merged main, running bootstrap had no effect vs the patch with the Tera fix.

dfreese (approved):

PiotrSikora:
I'm not sure if this is intentional or not, but rules for crates with `default = ["std"]` no longer include:
```
crate_features = [
    "default",
    "std",
],
```
(e.g. [third_party/cargo/remote/BUILD.blake2b_simd-0.5.11.bazel](https://github.com/google/cargo-raze/pull/478/files#diff-03be73f5b6ff9883bdea6b4a4525a7d00d491c9abf5a94b82f73602290cfe180) in this PR), which brakes crates that verify that they were built with `"std"` feature enabled (e.g. [BurntSushi/aho-corasick](https://github.com/BurntSushi/aho-corasick/blob/9af8bb9339f365976c95240cb3350743aefe86de/src/lib.rs#L184-L187)).

bsilver8192:
Regenerating the examples with this PR makes many of them fail to build because no crates have any features enabled. I'm looking at fixing it, FYI for anybody else encountering random build failures.

Issue #423, opened 2021-07-02
"Using cargo-raze through Bazel" documentation example doesn't compile anymore


sayrer:
If you update close to the main branch, something like this works:

```
#
# Use "cargo raze" through Bazel to download Cargo dependencies.
#
git_repository(
    name = "cargo_raze",
    commit = "ee2effee5b6c95cfcfd5e496677ce61dcd35c472",
    remote = "https://github.com/google/cargo-raze"
)

load("@cargo_raze//:repositories.bzl", "cargo_raze_repositories")

cargo_raze_repositories()

load("@cargo_raze//:transitive_deps.bzl", "cargo_raze_transitive_deps")

cargo_raze_transitive_deps()
```

depthwise:
This works, BTW:
```starlark
http_archive(
    name = "cargo_raze",
    sha256 = "0a7986b1a8ec965ee7aa317ac61e82ea08568cfdf36b7ccc4dd3d1aff3b36e0b",
    strip_prefix = "cargo-raze-0.12.0",
    url = "https://github.com/google/cargo-raze/archive/refs/tags/v0.12.0.tar.gz",
)
```

rbtcollins:
This http_archive snippet is failing for me, a lack of PCRE on the bazel build mirror:
```
$ bazel run @cargo_raze//:raze -- --manifest-path=$(realpath Cargo.toml)
WARNING: Download from https://mirror.bazel.build/ftp.pcre.org/pub/pcre/pcre-8.44.tar.gz failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
```

photex:
This is also failing for me on Linux and afaik MacOS. SSL failing in multiple ways depending the crates I've tried, but libgit, libssh, and openssl are all apparently quite a hassle. No problems with cargo and the libs living in the system however.

photex:
Also seems to fail on Windows. 

On MacOS, cloning this repo and running `bazel build //impl/...` does work however.

cfrantz:
I'm experiencing this issue wrt the pcre dependency.  ftp.pcre.org seems to be down and the the copy on mirror.bazel.build no longer seems to be present.

Is there a more reliable or stable URL for pcre-8.44?

UebelAndre:
> I'm experiencing this issue wrt the pcre dependency. ftp.pcre.org seems to be down and the the copy on mirror.bazel.build no longer seems to be present.
> 
[...]

PR #481, opened 2022-04-27
Fix build issues around libssh2

cfrantz:
I've been unable to build cargo-raze on my system.  It seems that libgit2's cmake configuration has trouble identifying which version of libssh2 to use.  It seems that the libssh2-sys crate has a similar problem.

These patches are IMHO a bit hacky, but I'm able to build `//impl:cargo_raze_bin`.


dfreese:
Can you provide more details about your system?  This is a pretty significant patch that I wouldn't have the bandwidth to maintain.

cfrantz:
> Can you provide more details about your system? This is a pretty significant patch that I wouldn't have the bandwidth to maintain.

It's a glinux desktop in glapstation configuration (ie: WFH, not on corp).  I don't have anything particularly exotic installed.

Without the changes to `//third_party/libgit2/BUILD.libgit2.bazel`, I see this error:
https://github.com/google/cargo-raze/issues/423#issuecomment-966432879

That error seems to be a variant of this issue people have reported when trying to build libgit for Windows: https://stackoverflow.com/questions/34421413/libgit2-with-libssh2-and-libopenssl-on-windows/34430810#34430810


Having resolved that issue, I then found that the crate libssh2-sys can't find the in-project libssh2 and wants to build its own copy of libssh2, thus causing a link error (duplicate symbols).  The patch to libssh2 allows an environment variable to specify that I've brought my own libssh2 and that one should be used.

I'd be happy to know if I'm just doing something wrong and that's why I can't build.

cfrantz:
Would you consider this patch easier to maintain if the patch for `ssh2-rs` were upstreamed to its repo?

dfreese:
> Would you consider this patch easier to maintain if the patch for `ssh2-rs` were upstreamed to its repo?

Yeah, I think it would be preferred to see what the `libssh2-sys` crate would accept in terms of patches first before trying to integrate them here.  The rest of the changes seems to be reasonable configuration.

cfrantz:
I found that by manipulating `PKG_CONFIG_PATH` and `LIBSSH2_SYS_USE_PKG_CONFIG` I could convince libssh2-sys to build with a pkgconfig file of my choice (no patch needed).  The cmake build rule seems to emit a broken `libssh2.pc` file and I don't think I have a hope in hell of figuring out what's going wrong in the cmake build, so I created a genrule to fix the file.

(The emitted pkgconfig file contains the string `${EXT_BUILD_ROOT}`, but that variable isn't defined, resulting in an empty expansion this yielding a pkgconfig that appears to refere to its libs and includes at the FS root).

cfrantz:
After examining the failure in CI, I've determined that the container image `gcr.io/bazel-public/ubuntu1804-java11:latest` does not have pkg-config installed, and therefore the rust crate `libssh2_sys`'s build.rs script cannot identify that I'd like it to build with a particular library.

I can build from the `main` branch within the ubuntu1804 container without any issue, but I cannot build from `main` on my glinux workstation.  I still believe that cmake is getting confused by my local packages, as I have pkg-config installed on my workstation.  I'd like to integrate `cargo-raze` into my project (as suggested by the README), but I cannot if I'm unable to build it.

PR #482, opened 2022-04-27
Replace incomplete repository_os logic with get_host_triple

dtolnay:
https://github.com/bazelbuild/rules_rust/pull/1289 exposes a better way to figure out the right triple to use with `load_arbitrary_tool`. In particular, we avoid assuming that `x86_64` is always the correct architecture.

### DIFF
--- a/tools/examples_repository.bzl
+++ b/tools/examples_repository.bzl
@@ -2,6 +2,7 @@
 actual vendored sources"""
 
 load("@rules_rust//rust:repositories.bzl", "load_arbitrary_tool")
+load("@rules_rust//rust/platform:triple.bzl", "get_host_triple")
 
 def _examples_dir(repository_ctx):
     """Returns the path to the cargo-raze workspace root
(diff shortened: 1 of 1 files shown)


transitive_deps.bzl
```diff
@@ -6,4 +6,7 @@ load("@rules_rust//rust:repositories.bzl", "rust_repositories")
 def cargo_raze_transitive_deps():
     """Loads all dependnecies from repositories required for cargo-raze"""
     rules_foreign_cc_dependencies()
-    rust_repositories(include_rustc_srcs = True)
+    rust_repositories(
+        edition = "2018",
```
dtolnay, line 10: This change is needed due to https://github.com/bazelbuild/rules_rust/pull/1256.

dfreese:
Sorry for the delay on this, but I think the issues with upstream rules_rust have been resolved with #506, so the main thrust of this change should not have a problem now.

dtolnay:
Rebased.

dfreese (approved):

dfreese:
Thanks!

PR #484, opened 2022-04-29
Add aarch64-apple-ios-sim triple

keith:
This triple is used for the arm64 iOS simulator on M1 macs. This
required a cfg-expr update to get this triple added upstream.


dfreese:
What's the Min Supported Rust Version (MSRV) on the new version of cfg-if?

keith:
It looks like `include_str` requires 1.54, I've updated CI as such, is that ok?

dfreese (approved):
I'm good with that.

We're 9+ months out from the [July 29, 2021 release of 1.54](https://blog.rust-lang.org/2021/07/29/Rust-1.54.0.html).  cargo-raze doesn't have a clear MSRV policy, but that's significantly less aggressive than when we bumped to 1.45.

keith:
Thanks!

PR #485, opened 2022-05-06
Testing bazel-based build breakage with upstream rules_rust

dfreese:


dfreese:
@UebelAndre, in attempting to track down some of the errors in #482, it looks like cargo no longer gets pulled in automatically after bumping the version of rules_rust.  I'm assuming that was an intentional change upstream.  Before I splunk too much further, is the appropriate fix on that obvious to you?

dfreese:
Looks like the breaking change was https://github.com/bazelbuild/rules_rust/pull/1079, which flipped incompatible_disable_custom_test_launcher.

UebelAndre:
I'm so sorry, I don't know how I missed this issue! I think all that's needed is for the tests to explicitly handle making the `CARGO` and `RUSTC` environment variables absolute paths. Then I think all should be fine.

dfreese:
fixed with #506, Thanks @sayrer

Issue #486, opened 2022-05-20
Feature Request: Add "tools" to CratesSettings


alealv:
It would be handy to have `tools` argument in order not to generate a custom `cargo_build_script`

As an example, I need the following to build bitar just because I cannot fill the `tools` section.

```bazel
# To be included as additional build script in `bitar`
load(
    "@rules_rust//cargo:cargo_build_script.bzl",
    "cargo_build_script",
)

cargo_build_script(
    name = "bitar_build_script",
    srcs = glob(["**/*.rs"]),
    build_script_env = {"PROTOC": "$(execpath @prx//:protoc)"},
    crate_features = [
        "brotli",
        "compress",
    ],
    crate_root = "build.rs",
    data = glob(["proto/**/*.proto"]),
    edition = "2018",
    rustc_flags = ["--cap-lints=allow"],
    tools = ["@com_google_protobuf//:protoc"],
    version = "0.9.0",
    visibility = ["//visibility:private"],
    deps = ["@raze__prost_build__0_8_0//:prost_build"],
)
```

This is needed because of https://github.com/bazelbuild/rules_rust/issues/884#issuecomment-1132782277

redshirtrob:
It seems like this should be closed since it was solved with #487.

PR #487, opened 2022-05-20
Add tools attribute for CrateSettings

alealv:
This solves https://github.com/google/cargo-raze/issues/486

### DIFF
--- a/impl/src/rendering/templates/partials/build_script.template
+++ b/impl/src/rendering/templates/partials/build_script.template
@@ -54,6 +54,13 @@ cargo_build_script(
         "{{flag}}",
         {%- endfor %}
     ],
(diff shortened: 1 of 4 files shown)


impl/src/rendering/templates/partials/build_script.template
```diff
@@ -54,6 +54,11 @@ cargo_build_script(
         "{{flag}}",
         {%- endfor %}
     ],
+    tools = [
```
dfreese, line 57: Can you make sure this doesn't print if tools is empty?  See line 44 for an example.

alealv: done

dfreese (approved):

PR #492, opened 2022-06-01
Allow skipping submodules

cfrantz:
Some crates may not need their submodules in order to build.  Add a
`CrateSettings` parameter to allow skipping submodule initialization
on such crates.

Signed-off-by: Chris Frantz <cfrantz@google.com>


impl/src/rendering/templates/remote_crates.bzl.template
```diff
@@ -18,6 +18,11 @@ def {{workspace.gen_workspace_prefix}}_fetch_remote_crates():
         commit = "{{crate.source_details.git_data.commit}}",
         build_file = Label("{{workspace.workspace_path}}/remote:BUILD.{{crate.pkg_name}}-{{crate.pkg_version}}.bazel"),
         init_submodules = True,
```
dfreese, line 20: Looks like this line needs to be removed.

dfreese (approved):
LGTM, can you rebase?

impl/src/rendering/templates/remote_crates.bzl.template
cfrantz: Fixed.

PR #506, opened 2022-06-17
Update to rules_rust 0.6.0

sayrer:
This fixes the path problems in #485.

The `rust_analyzer` rule is no longer needed, but rust_analyzer still works as we have it here. See https://github.com/bazelbuild/rules_rust/pull/1405.

### DIFF
--- a/impl/src/testing.rs
+++ b/impl/src/testing.rs
@@ -21,8 +21,6 @@ use tempfile::TempDir;
 
 use std::{
   collections::HashMap,
-  env,
-  env::{current_dir, set_var},
   fs::{create_dir_all, write, File},
   io::Write,
   path::Path,
@@ -40,11 +38,11 @@ use crate::{
 #[cfg(test)]
 #[ctor::ctor]
 fn init() {
-  let current_dir = current_dir().unwrap();
+  let current_dir = std::env::current_dir().unwrap();
   let env_vars_to_fix = ["CARGO", "CARGO_HOME", "RUSTC"];
   for var_name in env_vars_to_fix {
-    if let Ok(var_value) = env::var(var_name) {
-      set_var(var_name, current_dir.join(var_value));
+    if let Ok(var_value) = std::env::var(var_name) {
+      std::env::set_var(var_name, current_dir.join(var_value));
     }
   }
 }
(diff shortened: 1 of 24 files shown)


transitive_deps.bzl
```diff
@@ -6,4 +6,4 @@ load("@rules_rust//rust:repositories.bzl", "rust_repositories")
 def cargo_raze_transitive_deps():
     """Loads all dependnecies from repositories required for cargo-raze"""
     rules_foreign_cc_dependencies()
-    rust_repositories(include_rustc_srcs = True)
+    rust_repositories(edition = "2018", include_rustc_srcs = True)
```
dfreese, line 9: Can `include_rustc_srcs` be removed since rust_analyzer is being removed?

impl/src/testing.rs
```diff
@@ -21,6 +21,8 @@ use tempfile::TempDir;
 
 use std::{
   collections::HashMap,
+  env,
+  env::{current_dir, set_var},
```
dfreese, line 25: can you drop these use statements?  It ends up being only marginally longer, but more readable at the call site.

dfreese (approved):

transitive_deps.bzl
sayrer: This is still needed in one way or another, aiui. rust_analyzer still works, it's just an aspect now, and doesn't need an explicit rule. The workspace dependency on it is still there. We've gone back and forth on the `.bazelrc` approach vs this one, I just left it.

See: https://bazelbuild.github.io/rules_rust/rust_analyzer.html

impl/src/testing.rs
sayrer: Sure, do you want `std::env::` or `env::` inline? I had `std::env::` at first, but thought it was a bit ugly. No strong preference, though.

dfreese: I'd generally prefer `std::env::`.  Sorry, I'm a bit of a curmudgeon when it comes to this.

transitive_deps.bzl
dfreese: Sounds good, thanks for the clarification.

impl/src/testing.rs
sayrer: No worries, will fix. I think your way makes sense verbally, but is a bit visually ugly. 50/50 from my perspective.

Issue #510, opened 2022-06-20
Switch path code to Camino


sayrer:
I have some new features that use Camino, and I ended up replacing all of `cargo-raze`'s path code with Camino while hunting down bugs. Cargo requires UTF-8 paths anyway, as the Camino docs state, so I think it's a reasonable change. Ok to PR?

https://github.com/camino-rs/camino

illicitonion:
Works for me - see also https://github.com/google/cargo-raze/issues/421 :)

PR #511, opened 2022-06-21
Convert all paths to Camino

### LINKED ISSUE #421: Consider how to handle camino::Utf8PathBuf
https://github.com/oli-obk/cargo_metadata/pull/152 updated `cargo_metadata` to use [`camino::Utf8PathBuf`](https://docs.rs/camino/1.0.4/camino/struct.Utf8PathBuf.html) instead of `std::path::PathBuf`. The idea is that, as `cargo` only supports UTF-8 paths, this simplifies some handling (e.g. `Utf8PathBuf` can `impl Display`, rather than needing `.to_string_lossy()` and `.display()` calls)

We should work out whether we want to treat this as a localised change in that API (converting back to `PathBuf`/`Path` on the boundary with `cargo_metadata`), or whether we want to adopt `Utf8PathBuf` ourselves.

sayrer:
Still needs to be DRYed up, especially in the tests. Will close #421 and #510.

### DIFF
--- a/impl/Cargo.lock
+++ b/impl/Cargo.lock
@@ -351,9 +351,9 @@ checksum = "631ae5198c9be5e753e5cc215e1bd73c2b466a3565173db433f52bb9d3e66dba"
 
 [[package]]
 name = "camino"
-version = "1.0.4"
+version = "1.0.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


sayrer:
@dfreese @illicitonion I think this is ready.

illicitonion (approved):
Fantastic, thanks so much!

Issue #479, opened 2022-03-28
Regression in 0.13.0 adds aliases for ALL crate dependencies in a workspace, instead of for just the current package


joell:
In cargo raze 0.12.0, when generating Bazel rules for a workspace of multiple packages, the `BUILD.bazel` file for each package has `alias` rules for just the dependencies of that package.

In cargo raze 0.13.0, this behavior changed to add aliases for **all** the dependencies of **every** package in the workspace.

This causes Bazel alias conflicts when two packages within the workspace depend on different major versions of a crate (e.g., clap 2.x vs 3.x).  In such cases, the Bazel project can no longer build.

It is possible that issue #474 might be reflecting this deeper problem.

Issue #474, opened 2022-03-16
Duplicated alias rule in generated BUILD file


jgao54:
Hello, I have an existing project that uses cargo workspace and I am converting it to build with Bazel; and I am using cargo-raze (v0.15.0) in Remote mode. 

There are two packages: A and B. A has a devDependency on clap 3.x.x; B depends on A and another crate with a transitive dependency on clap 2.x.x. 

After running cargo raze, both versions are populated in the generated BUILD.bazel file with the same name:
```
alias(
    name = "clap",
    actual = "@raze__clap__2_33_3//:clap",
    tags = [
        "cargo-raze",
        "manual",
    ],
)

alias(
    name = "clap",
    actual = "@raze__clap__3_0_14//:clap",
    tags = [
        "cargo-raze",
        "manual",
    ],
)
```

Because of this, the project failed to build with:
```
Error in alias: alias rule 'clap' in package A conflicts with existing alias rule, defined at /path/to/A/cargo/BUILD.bazel:141:6
```

Temporary workaround is to manually edit the BUILD.bazel file and remove one of the aliases.

I also tried to rename clap 3.x.x in A's Cargo.toml file, however it did not affect the generated BUILD.bazel file.

(I saw https://github.com/google/cargo-raze/pull/282 and https://github.com/google/cargo-raze/pull/425, and thought this was fixed, but not familiar enough with cargo-raze to deduce why it didn't propagate the renaming)

jgao54:
I took a look at the source code; it looks like the generated BUILD file by default iterates thru the package names from the flat list of `packages` in cargo metadata, and use "resolve" graph to determine crate renaming where applicable via the `produce_workspace_aliases` function. 

However, since package A here is a workspace member and not a binary, it get skipped [here](https://github.com/google/cargo-raze/blob/8f151adf9a71e509e89f45ba8ac5f3e2da0734d5/impl/src/planning/subplanners.rs#L143-L146), and the renaming context from resolve graph is lost -- so the final generated BUILD file contains the original package name.

jgao54:
Here's a re-pro of the scenario described where the renaming is not honored in the generated aliases: https://github.com/jgao54/cargo-raze-example

PR #514, opened 2022-07-06
Begin work on resolving duplicate aliases.

sayrer:
This branch seems to resolve the initial problems encountered in https://github.com/liftedinit/many-rs/issues/143.

The code in `subplanner.rs` doesn't seem right. I'm just posting this as a WIP, since the crate here still doesn't build: https://github.com/fmorency/bazelalias

However, it seems like the `rand` crate there doesn't build, even without aliases.


sayrer:
See also:

https://github.com/google/cargo-raze/issues/479
https://github.com/google/cargo-raze/issues/474

PR #523, opened 2022-09-12
Print the error message when cargo tree fails

bsilver8192:
I made a simple typo (spelled a platform wrong), and figure out what was wrong was next to impossible without this output. With this change, it's obvious what the problem is.


illicitonion (approved):
Thanks!

bsilver8192:
@illicitonion I made it build with older Rust, could you take another look?

PR #527, opened 2022-09-17
Remove UebelAndre from CODEOWNERS

UebelAndre:
It's been quite a while since I've had bandwidth to review changes here or even context on it's current state. To save myself from notification noise and avoid giving folks false hope for reviews I'd like to be removed from CODEOWNERS.


dfreese:
I'm probably in a similar boat with changes in job/family situations.  :/

PR #537, opened 2022-11-25
Remove dfreese from CODEOWNERS

dfreese:
With a job switch and some other life changes, my meager bandwidth I've provided to this project is non-existant.  That leaves not a lot remaining, so it's probably worth having a conversation about what should happen going forward.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,3 +1,3 @@
 # Code owners of cargo-raze
 
-* @dfreese @illicitonion @acmcarther
+* @illicitonion @acmcarther


illicitonion (approved):
I hope the job and life changes are going well!

I appreciate reflecting the reality of the situation in terms of review bandwidth - I'm in a similar boat (though have a little more bandwidth for "easy obvious fixes"), and realistically I think we're probably at a point where `cargo-raze` effectively is without active maintainers... I'm not sure where we may find suitable replacements.

I do think that https://github.com/bazelbuild/rules_rust/tree/main/crate_universe currently provides a reasonable alternative that folks seem happy enough to be using, and which @UebelAndre and I have sufficient work-bandwidth and self-interest to keep maintaining... I think https://github.com/bazelbuild/rules_rust/issues/1662 is the last major piece of functionality crate_universe is missing that cargo-raze currently supports. After that's implemented, maybe we should consider a "maintainers needed / what blocks folks using the maintained thing" conversation (though I'm not sure exactly who the participants of that conversation would need to be)...

PR #543, opened 2023-02-21
Fix upcoming macOS toolchain incompatibility

keith:
Once https://github.com/bazelbuild/bazel/pull/16619 lands this target would fail to build because of ar vs libtool incompatibilities. This replaces the logic around setting AR to an empty string but I left that piece to continue supporting older versions of bazel as well.


keith:
going to fix upstream instead I think

